PR #2, opened 2016-10-21
Set up CI and coverage

jacehensley-wf:
We need CI and coverage turned on. Also we should add an `.analysis_options` file. 

Travis and codecov will still need to be turned on once we make this repo public. But all the code will be in place once we do that.


jacehensley-wf:
Closing because old fork is gone. See https://github.com/Workiva/over_react/pull/3

PR #3, opened 2016-10-24
Set up CI and coverage

jacehensley-wf:
We need CI and coverage turned on. Also we should add an `.analysis_options` file. 

Travis and codecov will still need to be turned on once we make this repo public. But all the code will be in place once we do that.

See https://github.com/Workiva/over_react/pull/2 for original PR.

@aaronlademann-wf @greglittlefield-wf @clairesarsam-wf @joelleibow-wf


aviary-wf:
## Raven
Number of Findings: 0

aaronlademann-wf:
+1

colefeisthamel-wf:
RM +1 manually reviewed dependencies

Issue #14, opened 2016-10-27
Make OverReact strong-mode compliant


jacehensley-wf:
> - Issue Type: FEATURE REQUEST

We should make OverReact strong-mode compliant. With all the new DDC stuff coming out including the Dart Console, it would be really nice to let our consumers use those new tools.

- [x] Written code is strong-mode compliant -  https://github.com/Workiva/over_react/pull/15
- [ ] Transformed code is strong-mode compliant

---

> **FYI:** @greglittlefield-wf @aaronlademann-wf @clairesarsam-wf @joelleibow-wf

jacehensley-wf:
Related issue, WRT transformed code, of what seems like a strong-mode bug: https://github.com/dart-lang/sdk/issues/28369

PR #15, opened 2016-10-28
UIP-1637 First pass at strong mode compliance

aaronlademann-wf:
**Ultimate problem:**
#14 - OverReact should be strong-mode compliant!

**How it was fixed:**
1. [Some changes were made to the react-dart wrapper](https://github.com/cleandart/react-dart/pull/106).
2. Lots of explicit type casting.
3. Resolving misc. Dart analyzer warnings, errors and hints.
4. For `_FluxComponentMixin`... I made it implement `BatchedRedraws`, but I'm not 100% confident in that solution.  Would be great to get @maxwellpeterson-wf, @evanweible-wf or @trentgrover-wf to weight in on the validity of that.
5. Turned on strong mode in `.analysis_options`.

**Testing suggestions:**
1. Verify that all tests pass
2. Verify that demo components behave as expected
3. Try it out on some of our more complex internal Workiva components to verify that regressions are not being introduced.

**Potential areas of regression:**
Pretty much everything 😁 
---

> **FYA:** @greglittlefield-wf @aaronlademann-wf @jacehensley-wf @clairesarsam-wf @joelleibow-wf @maxwellpeterson-wf, @evanweible-wf or @trentgrover-wf


aviary-wf:
## Raven
Number of Findings: 0

test/test_util/react_util.dart
```diff
@@ -156,8 +156,8 @@ Element renderAndGetDom(dynamic component, {bool addToRenderedQueue: true}) {
   return findDomNode(render(component, addToRenderedQueue: addToRenderedQueue));
 }
 
-/// Renders a React component or builder into a detached node and returns the associtated Dart component.
-react.Component renderAndGetComponent(dynamic component) => getDartComponent(render(component));
+/// Renders a React component or builder into a detached node and returns the associated Dart component.
+UiComponent renderAndGetComponent(dynamic component) => getDartComponent(render(component)) as UiComponent;
```
greglittlefield-wf: Why is this changed to `UiComponent`? It's possible to use this to render a non-over_react component, so we should leave it as `react.Component`.

web/src/demo_components/progress.dart
```diff
@@ -9,13 +9,6 @@ UiFactory<ProgressProps> Progress;
 
 @Props()
 class ProgressProps extends UiProps {
-  /// The id for the [Progress] component.
-  ///
-  /// If left unspecified one will be auto-generated for you to ensure
-  /// that the [caption] element is properly linked for accessibility purposes.
-  @override
-  String get id;
```
greglittlefield-wf: Could the `@virtual` annotation be used on the super field to allow this? (See [meta package](https://pub.dartlang.org/packages/meta) 1.0.4 changelog)

test/test_util/react_util.dart
```diff
@@ -287,7 +287,7 @@ bool _hasTestId(Map props, String key, String value) {
 
   bool first = false;
 
-  var results = react_test_utils.findAllInRenderedTree(root, allowInterop((descendant) {
+  var results = react_test_utils.findAllInRenderedTree(root, allowInterop((Function descendant) {
```
greglittlefield-wf: Pretty sure `descendant` is always either `ReactComponent` or `Element`, not `Function`...

test/test_util/react_util.dart
aaronlademann-wf: @greglittlefield-wf I changed it because on line 335 of `common_component_tests.dart`, `consumedProps` is not available on `Component`.

Should I just do a cast to `UiComponent` in the test?

aaronlademann-wf: Done here https://github.com/Workiva/over_react/pull/15/commits/2804edffc9261c4a7d918bf94f527f0944dccb17

web/src/demo_components/progress.dart
aaronlademann-wf: @greglittlefield-wf same as what I found in the `react-dart` PR... [`@virtual` doesn't seem to make the error go away](https://github.com/cleandart/react-dart/pull/106#discussion_r85552421).

jacehensley-wf: I think that only works when you override a field with a field 😢

aaronlademann-wf:
@jacehensley-wf @greglittlefield-wf @clairesarsam-wf this is ready for another pass.

integrate/pubspec.lock
```diff
@@ -0,0 +1,436 @@
+# Generated by pub
```
jacehensley-wf: I don't think this should be committed

evanweible-wf:
@aaronlademann-wf over_react's SDK range includes the latest (>=1.17.1), but the travis CI setup is pinned at 1.17.1. If you upgrade to the latest version of Dart, the analyzer step fails.

aaronlademann-wf:
nm @evanweible-wf @greglittlefield-wf ... @jacehensley-wf [got me straightened out](https://github.com/Workiva/over_react/pull/15/commits/01519355671728f56803b7cb113954f084438ae8).

greglittlefield-wf:
@aaronlademann-wf Actually, you only need that parameter in one place. The rest of the errors are due to either:
* the types of the `actions`/`store` getters not being statically declared
* the generic parameter not being supplied properly in the component class

From your penultimate commit, these changes fix all the warnings:
```diff
diff --git a/lib/src/component_declaration/flux_component.dart b/lib/src/component_declaration/flux_component.dart
index eb98d1a..08702d7 100644
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -18,7 +18,7 @@ abstract class FluxUiProps<ActionsT, StoresT> extends UiProps {
   /// There is no strict rule on the [ActionsT] type. Depending on application
   /// structure, there may be [Action]s available directly on this object, or
   /// this object may represent a hierarchy of actions.
-  get actions => props[_actionsPropKey] as ActionsT;
+  ActionsT get actions => props[_actionsPropKey] as ActionsT;
   set actions(ActionsT value) => props[_actionsPropKey] = value;
 
   /// The prop defined by [StoresT].
@@ -37,7 +37,7 @@ abstract class FluxUiProps<ActionsT, StoresT> extends UiProps {
   /// [StoresT] should be a class that provides access to these multiple stores.
   /// Then, you can explicitly select the [Store] instances that should be
   /// listened to by overriding [_FluxComponentMixin.redrawOn].
-  get store => props[_storePropKey] as StoresT;
+  StoresT get store => props[_storePropKey] as StoresT;
   set store(StoresT value) => props[_storePropKey] = value;
 }
 
@@ -121,7 +121,7 @@ abstract class _FluxComponentMixin<TProps extends FluxUiProps> implements Batche
   ///     redrawOn() => [store.tasks, store.users];
   List<Store> redrawOn() {
     if (props.store is Store) {
-      return [props.store];
+      return <Store>[props.store];
     } else {
       return [];
     }
diff --git a/test/over_react/component_declaration/flux_component_test/handler_precedence.dart b/test/over_react/component_declaration/flux_component_test/handler_precedence.dart
index 803f63c..1e977af 100644
--- a/test/over_react/component_declaration/flux_component_test/handler_precedence.dart
+++ b/test/over_react/component_declaration/flux_component_test/handler_precedence.dart
@@ -7,7 +7,7 @@ UiFactory<TestHandlerPrecedenceProps> TestHandlerPrecedence;
 class TestHandlerPrecedenceProps extends FluxUiProps<TestActions, TestStores> {}
 
 @Component()
-class TestHandlerPrecedenceComponent extends FluxUiComponent {
+class TestHandlerPrecedenceComponent extends FluxUiComponent<TestHandlerPrecedenceProps> {
   int numberOfRedraws = 0;
   int numberOfHandlerCalls = 0;
 
diff --git a/test/over_react/component_declaration/flux_component_test/redraw_on.dart b/test/over_react/component_declaration/flux_component_test/redraw_on.dart
index fa2b3d7..9fa3dfc 100644
--- a/test/over_react/component_declaration/flux_component_test/redraw_on.dart
+++ b/test/over_react/component_declaration/flux_component_test/redraw_on.dart
@@ -7,7 +7,7 @@ UiFactory<TestRedrawOnProps> TestRedrawOn;
 class TestRedrawOnProps extends FluxUiProps<TestActions, TestStores> {}
 
 @Component()
-class TestRedrawOnComponent extends FluxUiComponent {
+class TestRedrawOnComponent extends FluxUiComponent<TestHandlerPrecedenceProps> {
   int numberOfRedraws = 0;
 
   @override
diff --git a/test/over_react/component_declaration/flux_component_test/store_handlers.dart b/test/over_react/component_declaration/flux_component_test/store_handlers.dart
index dd81431..54ae1c0 100644
--- a/test/over_react/component_declaration/flux_component_test/store_handlers.dart
+++ b/test/over_react/component_declaration/flux_component_test/store_handlers.dart
@@ -7,7 +7,7 @@ UiFactory<TestStoreHandlersProps> TestStoreHandlers;
 class TestStoreHandlersProps extends FluxUiProps<TestActions, TestStore> {}
 
 @Component()
-class TestStoreHandlersComponent extends FluxUiComponent {
+class TestStoreHandlersComponent extends FluxUiComponent<TestStoreHandlersProps> {
   int numberOfHandlerCalls = 0;
 
   @override
```

aaronlademann-wf:
@greglittlefield-wf nice!

I've updated my latest commit to supply the generic params in the component class.

evanweible-wf:
@aaronlademann-wf @greglittlefield-wf there's a [linter rule (`avoid_as`)](http://dart-lang.github.io/linter/lints/avoid_as.html) with an explanation of why `as` should be avoided when possible - most usages in this lib are going to be required since over_react is providing typing where typing did not exist before, but there are a few places where it's unnecessary.

One example:

_lib/src/component_declaration/flux_component.dart L78-93_
```dart
  void componentWillMount() {
    Map<Store, Function> handlers = new Map.fromIterable(redrawOn(),
        value: (_) => (_) => redraw())..addAll(getStoreHandlers());

    handlers.forEach((store, handler) {
      StreamSubscription subscription = store.listen(handler as StoreHandler);
      _subscriptions.add(subscription);
    });
  }
```

Each handler here is being cast to `StoreHandler`, which means we'll get an exception in checked mode if someone returned a function from `getStoreHandlers()` that wasn't of type `StoreHandler`. IMO it would be better to put that type requirement up front by changing the type of `getStoreHandlers()`:

```diff
- Map<Store, Function> getStoreHandlers() {
+ Map<Store, StoreHandler> getStoreHandlers() {
    return {};
  }
```

Then we have a guarantee that every handler will be of type `StoreHandler` and we can get rid of the explicit type cast in componentWillMount:

```diff
  void componentWillMount() {
-   Map<Store, Function> handlers = new Map.fromIterable(redrawOn(),
+   Map<Store, StoreHandler> handlers = new Map.fromIterable(redrawOn(),
        value: (_) => (_) => redraw())..addAll(getStoreHandlers());

    handlers.forEach((store, handler) {
-     StreamSubscription subscription = store.listen(handler as StoreHandler);
+     StreamSubscription subscription = store.listen(handler);
      _subscriptions.add(subscription);
    });
  }
```

Thoughts?

Also @greglittlefield-wf nice use of penultimate.

greglittlefield-wf:
@evanweible-wf My junior year HS English teacher loved that word 😄.

integrate/lib/test_react_element_typing.dart
```diff
@@ -20,7 +20,14 @@ import './test_component_declaration.dart';
 
 /// Verify that treating the invocation of the builder as an unparameterized [ReactElement]
 /// does not result in the following analyzer warning:
-/// > Unsound implicit cast from ReactElement<dynamic> to ReactElement<Component>
-void _reactElementTypingTest() {
+///
+///     Unsound implicit cast from ReactElement<dynamic> to ReactElement<Component>
```
greglittlefield-wf: `ReactElement` no longer has a generic parameter as of a breaking release or two ago, so we can probably just remove this whole file.

aaronlademann-wf:
ok... @evanweible-wf @greglittlefield-wf I think we're at a good place here.  Final review maybe?

test/over_react/util/handler_chain_util_test.dart
```diff
@@ -51,7 +51,9 @@ main() {
         }
 
         List<TestGenericType> generateBadTypeArgs() {
-          return new List.generate(arity, (_) => new Object());
+          // ignore: Type check failed
```
greglittlefield-wf: Hm, I think this should just be `List generateBadTypeArgs() {`

test/over_react/util/handler_chain_util_test.dart
```diff
@@ -158,7 +160,7 @@ main() {
             test('calls all functions in order', () {
               var calls = [];
 
-              var functions = new List.generate(5, (index) {
+              List<Function> functions = new List.generate(5, (index) {
```
greglittlefield-wf: #supernit for something like this, I think it might be preferred to do
```dart
var functions = new List<Function>.generate(5, (index) {
```
@evanweible-wf ?

test/over_react/util/rem_util_test.dart
```diff
@@ -205,7 +205,7 @@ main() {
           'correctly dispatches an event in resopnse to the first change', () async {
         expect(querySelector('#rem_change_sensor'), isNull);
 
-        var calls = [];
+        List<double> calls = [];
```
greglittlefield-wf: #supernit Same here: should this be 
```dart
var calls = <double>[]?
```

Asking about these out mostly to improve my own knowledge.

.travis.yml
```diff
@@ -1,6 +1,6 @@
 language: dart
 dart:
-  - "1.17.1"
+  - "1.19.1"
```
evanweible-wf: There are two more spots where the analyzer is failing on 1.20.1, but they're easy fixes:

_test/over_react/component_declaration/flux_component_test/handler_precedence.dart L21-25_
```diff
  @override
  getStoreHandlers() => {props.store.store1: increment};

- increment(_) {
+ increment(Store store) {
    numberOfHandlerCalls += 1;
  }
```

Same thing for _test/over_react/component_declaration/flux_component_test/store_handlers.dart L16-21_.

With those two changes, I think you'd be fine to bump this up to 1.20.1

lib/src/component_declaration/component_base.dart
```diff
@@ -175,7 +166,10 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
   @override
   TProps get props {
     var unwrappedProps = this.unwrappedProps;
-    TProps typedProps = _typedPropsCache[unwrappedProps];
+    /// Have to cast as [TProps] until we can parameterize [_typedPropsCache].
```
greglittlefield-wf: Hm, I wonder if we could reintroduce the type parameter on this Expando to avoid this cast... it seems the linked issue might only apply to mixins, so the omitted parameterization may just be carried over from when this code was part of a mixin. 

These getters are used a lot, so it would be nice to avoid the perf hit of this cast if possible.

lib/src/transformer/declaration_parsing.dart
```diff
@@ -64,8 +64,8 @@ class ParsedDeclarations {
 
     // Validate the types of the annotated declarations.
 
-    List topLevelVarsOnly(String annotationName, Iterable<CompilationUnitMember> declarations) {
-      var topLevelVarDeclarations = [];
+    List<CompilationUnitMember> topLevelVarsOnly(String annotationName, Iterable<CompilationUnitMember> declarations) {
+      var topLevelVarDeclarations = <CompilationUnitMember>[];
```
evanweible-wf: I think these should be `<TopLevelVariableDeclaration>`

lib/src/transformer/declaration_parsing.dart
```diff
@@ -81,8 +81,8 @@ class ParsedDeclarations {
       return topLevelVarDeclarations;
     };
 
-    List classesOnly(String annotationName, Iterable<CompilationUnitMember> declarations) {
-      var classDeclarations = [];
+    List<CompilationUnitMember> classesOnly(String annotationName, Iterable<CompilationUnitMember> declarations) {
+      var classDeclarations = <CompilationUnitMember>[];
```
evanweible-wf: ..and these should be `<ClassDeclaration>` - that should fix the failing tests

.travis.yml
jacehensley-wf: Even if you're not using the variable is has to be typed?

evanweible-wf:
@aaronlademann-wf @greglittlefield-wf I tried running tests on 1.20.1, but there were several failures that seem like the must be a bug. Same tests pass on 1.19.1:

```
      'http://localhost:58105/over_react/component/resize_sensor_test.dart': error: line 79 pos 51: expression is not a valid compile-time constant
              {ResizeSensorHandler onResize, int width: defaultContainerWidth, int height: defaultContainerHeight,
                                                        ^

      dart:async                                           Future.Future.microtask
      over_react/component/resize_sensor_test.dart         main.<fn>.expectResizeAfter
      over_react/component/resize_sensor_test.dart 239:32  main.<fn>.<fn>.<fn>.<async>
      ===== asynchronous gap ===========================
      dart:async                                           Future.Future.microtask
      over_react/component/resize_sensor_test.dart         main.<fn>.<fn>.<fn>
```

greglittlefield-wf:
@evanweible-wf Looks like a bug indeed; I've got a reduced test case and am filing an SDK bug

lib/src/component_declaration/component_base.dart
jacehensley-wf: Just tried this locally and it works

test/over_react/util/handler_chain_util_test.dart
aaronlademann-wf: @greglittlefield-wf @evanweible-wf when I do that, I get the following test failures:

```
00:11 +887 -1: [Dartium Content Shell] test/over_react_test.dart: HandlerChainUtil generic chaining: CallbackUtil0Arg chainFromList() returns a function of arity 0 that calls all functions in order
      type 'List<Function>' is not a subtype of type 'List<Callback0Arg>' of 'callbacks' where
        List is from dart:core
        Function is from dart:core
        List is from dart:core

      package:over_react/src/util/handler_chain_util.dart 212:41  CallbackUtil.chainFromList
      over_react/util/handler_chain_util_test.dart 167:42         main.<fn>.<fn>.sharedTests.<fn>.<fn>.<fn>
```

lib/src/component_declaration/component_base.dart
aaronlademann-wf: @greglittlefield-wf @jacehensley-wf `Expando`s parameterized, casts removed in https://github.com/Workiva/over_react/pull/15/commits/c8b41b7bffb83b930102c2f10586314704dbeb3c

aaronlademann-wf:
@greglittlefield-wf @jacehensley-wf @evanweible-wf both dependent PRs have merged, `react 3.0.1` has been published.

This should be ready for a final pass.

lib/src/transformer/impl_generation.dart
```diff
@@ -364,19 +364,20 @@ class ImplGenerator {
 
     typedMap.node.members
         .where((member) => member is FieldDeclaration)
-        .where((FieldDeclaration member) => !member.isStatic)
-        .forEach((FieldDeclaration field) {
+        .where((member) => !(member as FieldDeclaration).isStatic) // ignore: avoid_as
```
jacehensley-wf: Instead of this and the previous line you could just do:

```dart
typedMap.node.members
    .where((memeber) => member is FieldDeclaration && !member.isStatic)
    .forEach(...
```

lib/src/transformer/impl_generation.dart
```diff
@@ -364,19 +364,20 @@ class ImplGenerator {
...
+        .forEach((field) {
```
jacehensley-wf: #nit Just so there is not a bunch of changes could the param be prefixed with an underscore and the new var not?

```dart
.forEach((_field) {
  final field = _field as FieldDeclaration;
})
```

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -294,7 +294,7 @@ main() {
 
       group('modifyProps()', () {
         test('passes the provided modifier itself', () {
-          modifier(UiProps props) {
+          modifier(Map<dynamic, dynamic> props) {
```
jacehensley-wf: #nit Unnecessary parameters

aaronlademann-wf:
@jacehensley-wf feedback addressed.

test/test_util/custom_matchers.dart
```diff
@@ -55,7 +55,9 @@ class ClassNameMatcher extends Matcher {
   }
 
   @override
-  bool matches(String className, Map matchState) {
+  bool matches(className, Map matchState) {
+    className = className as String; // ignore: avoid_as
```
jacehensley-wf: We wouldn't have to do this if this were to be accepted https://github.com/dart-lang/matcher/issues/37

test/over_react/util/handler_chain_util_test.dart
greglittlefield-wf: Ah, sorry, should it be this, then?
```dart
var functions = new List<Callback0Arg>.generate(5, (index) {
```

I'm fine leaving it as-is if it works, though

greglittlefield-wf:
#shipit

__+1__

evanweible-wf:
+1

aaronlademann-wf:
@leviwith-wf this is ready for QA / merge.

test/test_util/custom_matchers.dart
jacehensley-wf: Update from that issue. On Dart 1.20.1 you can use `@checked` to override param types.

leviwith-wf:
QA Resource Approval: +10
- tests pass (observed noted failures in current dart sdk)
- provided examples behave themselves.

-----
 
- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] Unit test created/updated
- [X] All unit tests pass

Merging.

PR #16, opened 2016-10-29
changed react_dom.render(FooComponent()() to react_dom.render(Foo()()

denniskaselow:
The example code always uses:

``` dart
@Factory()
UiFactory<FooProps> Foo;
```

So shouldn't Foo also be used in the initial render call? FooComponent is a bit misleading.

### DIFF
--- a/README.md
+++ b/README.md
@@ -94,7 +94,7 @@ mount / render it into the HTML element you created in step 3.
       react_client.setClientConfiguration();
       
       // Mount / render your component. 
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Raven
Number of Findings: 0

johncblandii:
Good catch.

johncblandii:
+1

aaronlademann-wf:
+1

colefeisthamel-wf:
RM +1 manually reviewed dependencies.

PR #29, opened 2016-11-11
Add map_util method newStyleFromProps

clairesarsam-wf:
## Ultimate problem:

We frequently run into an `over_react` use case where we have a computed inline style we'd like to merge into an existing props map that may or may not be defined.

## How it was fixed:

Add a new method to `map_util`, `newStyleFromProps`.

## Testing suggestions:

* All tests should pass

None.

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @jacehensley-wf @joelleibow-wf

### DIFF
--- a/lib/src/util/map_util.dart
+++ b/lib/src/util/map_util.dart
@@ -64,7 +64,7 @@ Map getPropsToForward(Map props, {bool omitReactProps: true, bool onlyCopyDomPro
 /// Returns a copy of the style map found in [props].
 ///
 /// Returns an empty map if [props] or its style map are null.
-Map newStyleFromProps(Map props) {
+Map<String, dynamic> newStyleFromProps(Map props) {
   if (props == null) return {};
 
   var existingStyle = domProps(props).style;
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/util/map_util.dart
```diff
@@ -60,6 +61,16 @@ Map getPropsToForward(Map props, {bool omitReactProps: true, bool onlyCopyDomPro
   return propsToForward;
 }
 
+/// Returns a copy of the style map found in [props].
+///
+/// Returns an empty map if [props] or its style map are null.
+Map newStyleFromProps(Map props) {
```
aaronlademann-wf: IMO, this return signature should match the [type signature of the `style` prop](https://github.com/Workiva/over_react/blob/master/lib/src/component/prop_mixins.dart#L65)...

```dart
Map<String, dynamic> newStyleFromProps(Map props) {
  // ...
}
```

greglittlefield-wf:
+1

greglittlefield-wf:
Closing, since these changes will be pulled in from the web_skin_dart commit history.

PR #51, opened 2017-02-24
UIP-1590: Deprecate the Required annotation (Includes UIP-1835)

jacehensley-wf:
## Ultimate problem:
The `Required` annotation used by the `over_react` transformer collides with the annotation from `package:meta`.

## How it was fixed:
- Deprecate `Required`, and add more options to the `Accessor` annotation.
- Update logic around the transformer to support the new options on an `Accessor` annotation.
- Added shorthand consts for `Accessor(isRequired: true)` and `Accessor(isRequired: true, isNullable: true)`.

## Testing suggestions:
- Verify tests pass
- On one of the example components, make a prop required via the new options on an `Accessor` annotation.
  - Verify that when you don't set it, the component throws an error.

- Required annotation support.

---

### DIFF
--- a/lib/src/component_declaration/annotations.dart
+++ b/lib/src/component_declaration/annotations.dart
@@ -186,6 +186,9 @@ class StateMixin implements TypedMap {
 
 /// Marks a `prop` as required to be set.
 ///
+/// Validation occurs in `UiComponent.validateRequiredProps` which requires super calls into `componentWillMount` and
+/// `componentWillReceiveProps`.
+///
 ///     @Props()
 ///     abstract class FooProps {
 ///       @requiredProp
@@ -194,6 +197,9 @@ class StateMixin implements TypedMap {
 const Accessor requiredProp = const Accessor(isRequired: true);
 
 /// Marks a `prop` as required to be set, but allowed to be set explicitly to `null`.
+///
+/// Validation occurs in `UiComponent.validateRequiredProps` which requires super calls into `componentWillMount` and
+/// `componentWillReceiveProps`.
 ///
 ///     @Props()
 ///     abstract class FooProps {
--- a/lib/src/component_declaration/annotations.dart
+++ b/lib/src/component_declaration/annotations.dart
@@ -184,13 +184,42 @@ class StateMixin implements TypedMap {
   const StateMixin({this.keyNamespace: null});
 }
 
+/// Marks a `prop` as required to be set.
+///
+///     @Props()
+///     abstract class FooProps {
+///       @requiredProp
+///       String requiredProp;
+///     }
+const Accessor requiredProp = const Accessor(isRequired: true);
+
+/// Marks a `prop` as required to be set, but allowed to be set explicitly to `null`.
+///
+///     @Props()
+///     abstract class FooProps {
+///       @nullableRequiredProp
+///       String nullableRequiredProp;
+///     }
+const Accessor nullableRequiredProp = const Accessor(isRequired: true, isNullable: true);
+
 /// Annotation used with the `over_react` transformer to customize individual accessors (props/state fields).
+///
+/// Validation occurs in `UiComponent.validateRequiredProps` which requires super calls into `componentWillMount` and
+/// `componentWillReceiveProps`.
 ///
 ///     @Props()
 ///     abstract class FooProps {
 ///       @Accessor(keyNamespace: '', key: 'custom_key')
 ///       String bar;
+///
+///       @Accessor(isRequired: true)
+///       String requiredProp;
+///
+///       @Accessor(isRequired: true, isNullable: true)
+///       String nullableRequiredProp;
 ///     }
+///
+/// Related: [requiredProp], [nullableRequiredProp].
 class Accessor {
   /// A key for the annotated accessor, overriding the default of the accessor's name.
   final String key;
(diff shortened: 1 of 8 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/component_declaration/annotations.dart
```diff
@@ -199,22 +228,26 @@ class Accessor {
   /// overriding the default of `'${enclosingClassName}.'`.
   final String keyNamespace;
 
+  /// Whether the accessor is required to be set.
+  final bool isRequired;
+
+  /// Whether setting a prop to null is allowed.
+  final bool isNullable;
+
+  /// The erro message displayed when the accessor is not set.
```
aaronlademann-wf: #nit s/erro/error

lib/src/component_declaration/annotations.dart
```diff
@@ -199,22 +228,26 @@ class Accessor {
...
-///
-/// This is validated in `UiComponent.validateRequiredProps` which requires super calls into `componentWillMount` and
-/// `componentWillReceiveProps`.
-///
-///     @Props()
-///     abstract class FooProps {
-///       @Required()
-///       String bar;
-///     }
+/// Deprecated use [Accessor], [requiredProp], or [nullableRequiredProp] instead.
```
aaronlademann-wf: #nit 

```dart
/// Deprecated.
///
/// Use [Accessor], [requiredProp] or [nullableRequiredProp] instead.
```

jacehensley-wf:
@aaronlademann-wf Feedback has been addressed

aaronlademann-wf:
+1

test/over_react/component_declaration/transformer_integration_tests/required_accessor_integration_test.dart
```diff
@@ -0,0 +1,161 @@
...
+
+  @Accessor(isRequired: true, isNullable: true, requiredErrorMessage: 'This prop can be set to null!')
+  var nullableProp;
+}
+
+@Component()
+class ComponentTestComponent extends UiComponent<ComponentTestProps> {
+  @override
+  render() => Dom.div()();
+}
```
greglittlefield-wf: We should have integration tests for the alias/shorthand consts as well.

I actually don't think that they're getting picked up right now...

lib/src/transformer/impl_generation.dart
```diff
@@ -402,10 +402,6 @@ class ImplGenerator {
             annotations.Accessor accessorMeta = instantiateAnnotation(field, annotations.Accessor);
```
greglittlefield-wf: Yeah, I don't think this will pick up the constants.

You'll probably have to just match them by name, like so:
```dart

T getConstantAnnotation<T>(AnnotatedNode member, String name, T value) =>
    member.metadata.any((annotation) => annotation.name?.name == name)) ? value : null;


annotations.Accessor requiredProp = instantiateAnnotation(field, 'requiredProp', annotations.requiredProp);
annotations.Accessor nullableRequiredProp = instantiateAnnotation(field, 'nullableRequiredProp', annotations.nullableRequiredProp);

if (requiredProp != null) {
  // ...
}
if (nullableRequiredProp != null) {
  // ...
}
```

greglittlefield-wf: We should also probably emit a warning if they try to use `@requiredProp`/`nullableRequiredProp`/`@Accessor` together.

jacehensley-wf: Oh I thought it would pick up on that sort of thing too. What happens if a consumers has there own const shorthand for some reason, should we be supporting that?

greglittlefield-wf: Unfortunately we can't support that without performing a full analysis 😢

lib/src/component_declaration/annotations.dart
```diff
@@ -184,13 +184,42 @@ class StateMixin implements TypedMap {
...
+///     abstract class FooProps {
+///       @nullableRequiredProp
+///       String nullableRequiredProp;
+///     }
+const Accessor nullableRequiredProp = const Accessor(isRequired: true, isNullable: true);
+
 /// Annotation used with the `over_react` transformer to customize individual accessors (props/state fields).
 ///
+/// Validation occurs in `UiComponent.validateRequiredProps` which requires super calls into `componentWillMount` and
+/// `componentWillReceiveProps`.
```
jacehensley-wf: Addressed

test/over_react/component_declaration/transformer_integration_tests/required_accessor_integration_test.dart
jacehensley-wf: Addressed

lib/src/transformer/impl_generation.dart
jacehensley-wf: Addressed

jacehensley-wf:
@greglittlefield-wf Feedback has been addressed. The `requiredProp` and `nullableRequiredProp` are now actually being picked up by the transformer.

lib/src/transformer/impl_generation.dart
```diff
@@ -415,11 +417,44 @@ class ImplGenerator {
...
+              constantValue += ', isRequired: true';
+            }
+
+            if (nullableRequiredProp != null) {
+              annotationCount++;
+              constantValue += ', isRequired: true, isNullable: true';
+            }
+
+            if (annotationCount > 1) {
+              logger.error('At most only a single annotation can be applied to an accessor.', span: getSpan(sourceFile, field));
```
greglittlefield-wf: This message might be a little misleading, since other annotations like `@deprecated` are also allowed.

What about something like:
>``` 
>@requiredProp/@nullableProp/@Accessor cannot be used together.
>You can use `@Accessor(required: true)` or `isNullable: true` instead of the shorthand versions.
>```

greglittlefield-wf:
- Code looks good
- Tests pass in CI (Dart)
- Tests pass locally in dart2js
- Transformer fails gracefully w/ error message when multiple conflicting annotations are used

+10

leviwith-wf:
QA +1

- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #55, opened 2017-03-09
UIP-1802 CSS Transition Timeout/Warning

kaanaksoy-wk:
## Ultimate problem:

When creating custom `Modals`, it's possible to set transition to a value that `ModalTrigger` doesn't know about, which causes the cleanup of the modal to hang.

We add a timeout to ensure that the modal is cleaned up even if something goes wrong with the transition, and warn the consumer.

## How it was fixed:

- Add a `transitionTimeout` getter to `AbstractTransitionComponent` that can be overridden to adjust the desired timeout duration.
- Use a `Timer` and display a warning after a timeout has occurred.
- Write a test to verify that the timeout actually occurs.

## Testing suggestions:
- Verify that the tests pass.

- N/A

---

### DIFF
--- a/lib/src/component/abstract_transition.dart
+++ b/lib/src/component/abstract_transition.dart
@@ -123,6 +123,9 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
   /// Whether the Element returned by [getTransitionDomNode] will have a transition event.
   bool get hasTransition => true;
 
+  /// The duration that can elapse before a transition timeout occurs.
+  Duration get transitionTimeout => const Duration(seconds: 15);
+
   // --------------------------------------------------------------------------
   // Private Utility Methods
   // --------------------------------------------------------------------------
@@ -184,7 +187,12 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
       skipCount = 0;
     }
 
+    var timer = new Timer(transitionTimeout, () {
+      assert(ValidationUtil.warn("A transition timeout has occurred."));
+    });
+
     _endTransitionSubscription = getTransitionDomNode()?.onTransitionEnd?.skip(skipCount)?.take(1)?.listen((_) {
+      timer.cancel();
       complete();
     });
   }
--- a/lib/src/component/abstract_transition.dart
+++ b/lib/src/component/abstract_transition.dart
@@ -124,7 +124,7 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
   bool get hasTransition => true;
 
   /// The duration that can elapse before a transition timeout occurs.
-  Duration get transitionTimeout => const Duration(seconds: 15);
+  Duration get transitionTimeout => const Duration(seconds: 1);
 
   // --------------------------------------------------------------------------
   // Private Utility Methods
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
Number of Findings: 1
* Watched file ```aviary.yaml``` added

As a result of the finding(s) listed above, a security review is required for this pull request.

lib/src/component/abstract_transition.dart
```diff
@@ -184,7 +187,12 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S
       skipCount = 0;
     }
 
+    var timer = new Timer(transitionTimeout, () {
+      assert(ValidationUtil.warn("A transition timeout has occurred."));
```
jacehensley-wf: This should be a more informative message. Something like: 'The amount of transitions expected to complete have not completed. Something is most likely wrong.' The wording isn't great but something that get across the point that we are expecting some amount of transitions and they have not fired.

test/over_react/component/abstract_transition_test.dart
```diff
@@ -413,6 +413,25 @@ main() {
...
+      startRecordingValidationWarnings();
+
+      var renderedInstance = render(Transitioner()
+        ..transitionCount = 1
+        ..transitionTimeout = const Duration(seconds: 0)
+      );
+
+      TransitionerComponent transitioner = getDartComponent(renderedInstance);
+
+      transitioner.hide();
```
jacehensley-wf: You should add a `expect` before and after this that verifies that the `state.transtionPhase` is `TransitionPhase.SHOWN`.

lib/src/component/abstract_transition.dart
```diff
@@ -123,6 +123,9 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S
   /// Whether the Element returned by [getTransitionDomNode] will have a transition event.
   bool get hasTransition => true;
 
+  /// The duration that can elapse before a transition timeout occurs.
+  Duration get transitionTimeout => const Duration(seconds: 15);
```
greglittlefield-wf: I'd say make this a prop so that it's configurable, and default it to something much smaller, like 1s, so that consumers see the warning right away.

jacehensley-wf: I feel like a getter is better for this, similar to `hasTransition`. That way consumers have the ability to either proxy it with a prop or have a custom implementation.

1s sounds good to me tho.

greglittlefield-wf: Yeah, the only reason I said prop is that it would be easier to override. Good point, I'm fine leaving it as a getter.

lib/src/component/abstract_transition.dart
```diff
@@ -184,7 +187,14 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S
       skipCount = 0;
     }
 
+    var timer = new Timer(transitionTimeout, () {
+      assert(ValidationUtil.warn(
+        'The number of transitions expected to complete have not completed. Something is most likely wrong.'
+      ));
```
jacehensley-wf: Something else that I was thinking about. Should we still call `complete` if the timer completes? I could go either way, was just wondering what the desired behavior is.

test/over_react/component/abstract_transition_test.dart
```diff
@@ -444,6 +471,7 @@ class TransitionerComponent extends AbstractTransitionComponent<TransitionerProp
     ..addProps(super.getDefaultProps())
     ..hasTransition = true
     ..initiallyShown = true
+    ..transitionTimeout = const Duration(seconds: 15)
```
jacehensley-wf: This should be updated to match the default in `AbstractTransitionComponent`

lib/src/component/abstract_transition.dart
greglittlefield-wf: Mm, yes, we should, in order to account for when the warning isn't caught by the time this is deployed

jacehensley-wf:
+1

aaronlademann-wf:
+10

teresarevious-wf:
RM +1 verified dependencies

teresarevious-wf:
RM +1 verified dependencies

PR #58, opened 2017-03-24
UIP-2153: Transition warning sometimes fires when it shouldn't

kaanaksoy-wk:
## Ultimate problem:
The newly-added transition warning ([UIP-1802](https://github.com/Workiva/over_react/pull/55)),
`VALIDATION WARNING: The number of transitions expected to complete have not completed. Something is most likely wrong.` fires sometimes when it shouldn't.

For instance, sometimes `transitionend` events never fire on purpose, such as when the user cancels the transition (think hiding a popover while it's still showing). This results in a warning when it should not.

I also noticed in the code that the timer does not unsubscribe the listener function, which it probably should.

## How it was fixed:
- Create an instance field, `_transitionEndTimer`, that keeps track of whether a transition timeout has occurred.
- When a transition is canceled (ie, after hiding a popover while it's still showing), `_transitionEndTimer` is canceled as well.
- Make the timer unsubscribe the listener function.

## Testing suggestions:
- Verify that the tests pass.
- While in Dart checked mode, verify that rapidly clicking the "Pick a date" button twice (on [WSD Docs](https://docs.workiva.org/web_skin_dart/latest/components/#datepicker-custom-usage)) does not result in a validation warning.

---

### DIFF
--- a/lib/src/component/abstract_transition.dart
+++ b/lib/src/component/abstract_transition.dart
@@ -188,10 +188,10 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
       assert(ValidationUtil.warn(warningMessage));
 
       skipCount = 0;
-
-      _cancelTransitionEventListener();
-      _cancelTransitionEndTimer();
     }
+
+    _cancelTransitionEventListener();
+    _cancelTransitionEndTimer();
 
     _transitionEndTimer = new Timer(transitionTimeout, () {
       assert(ValidationUtil.warn(
@@ -199,7 +199,7 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
       ));
 
       _cancelTransitionEventListener();
-      
+
       complete();
     });
 
@@ -312,8 +312,6 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
 
   /// Method that will be called when [AbstractTransitionComponent]  first enters the `preShowing` state.
   void handlePreShowing() {
-    _cancelTransitionEndTimer();
-
     onNextTransitionEnd(() {
       if (state.transitionPhase == TransitionPhase.SHOWING) {
         setState(newState()
@@ -348,6 +346,7 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
         );
       });
 
+      _cancelTransitionEventListener();
       _cancelTransitionEndTimer();
     } else {
       onNextTransitionEnd(() {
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/component/abstract_transition.dart
```diff
@@ -187,16 +190,22 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S
...
       ));
 
+      _cancelTransitionEventListener();
       complete();
     });
 
     _endTransitionSubscription = getTransitionDomNode()?.onTransitionEnd?.skip(skipCount)?.take(1)?.listen((_) {
-      timer.cancel();
+
+      if (_transitionEndTimer != null) {
```
jacehensley-wf: There should be a private helper function to do this. Also you can just use a null-aware operator.

```dart
void _cancelTransitionEndTimer() {
  _transitionEndTimer?.cancel();
  _transitionEndTimer = null;
}
```

jacehensley-wf:
+1

greglittlefield-wf:
I'm seeing warnings when I shouldn't when double-clicking the datepicker button rapidly when it's already open. 😢

lib/src/component/abstract_transition.dart
```diff
@@ -187,16 +190,19 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S
       skipCount = 0;
     }
```
greglittlefield-wf: Hm, I wonder if we should just always do the following here:
```dart
_cancelTransitionEventListener();
_cancelTransitionEndTimer();
```
to clean up any previous calls

kaanaksoy-wk: @greglittlefield-wf Do you think the extraneous warnings still appearing could be resolved by placing `_cancelTransitionEndTimer()` at the top of the `handlePreShowing` method?

 I tried it and it seems to work, but I'm not sure if doing this defeats the purpose of the timer in the first place.

greglittlefield-wf: Potentially, but wouldn't that be accomplishing the same as what I linked above, albeit in a subset of cases?

greglittlefield-wf:
- [x] Changes look good
- [x] Rapidly toggling visibility of a transitioning element does not result in the warning
- [x] Warning is still logged as expected when CSS transitions are manually disabled, simulating the use-case of the warning

+10

jacehensley-wf:
+1

leviwith-wf:
QA +1

- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #59, opened 2017-03-30
UIP-2186 Don't call setState when AbstractTransitionComponent is unmounting

joshbeam-wf:
This resolves a warning of the form "Warning: setState(...): Can only update a mounted or mounting component. This usually means you called setState() on an unmounted component. This is a no-op."

## Ultimate problem:
An AbstractTransitionComponent's `setState` call to set the transition phase to `TransitionPhase.HIDDEN` can occur after the component has been unmounted. This may happen if the component contains a button that causes the AbstractTransitionComponent to be unmounted when clicked.

## How it was fixed:
Added `_isMounted` boolean to the component to keep track of whether or not the component is mounted (it's set to true in `componentDidMount` and set to false in `componentWillUnmount`). The value of this boolean is checked before calling `setState` to set the transition phase to `TransitionPhase.HIDDEN`.

## Testing suggestions:
Verify that tests pass.

Components based on AbstractTransitionComponent.

---

### DIFF
--- a/lib/src/component/abstract_transition.dart
+++ b/lib/src/component/abstract_transition.dart
@@ -281,8 +281,16 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
     }
   }
 
+  var _isMounted = false;
+
+  @override
+  void componentDidMount() {
+    _isMounted = true;
+  }
+
   @override
   void componentWillUnmount() {
+    _isMounted = false;
     _cancelTransitionEventListener();
   }
 
@@ -333,7 +341,7 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S 
       });
     } else {
       onNextTransitionEnd(() {
-        if (state.transitionPhase == TransitionPhase.HIDING) {
+        if (_isMounted && state.transitionPhase == TransitionPhase.HIDING) {
           setState(newState()
             ..transitionPhase = TransitionPhase.HIDDEN
           );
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Raven
Number of Findings: 0

greglittlefield-wf:
- Change and tests look great.
- Tests pass.
- No regressions found in transition-related behavior.

+10

Thanks for the contribution!

jacehensley-wf:
+1

lib/src/component/abstract_transition.dart
```diff
@@ -281,8 +281,16 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps, S
     }
   }
 
+  var _isMounted = false;
+
+  @override
+  void componentDidMount() {
+    _isMounted = true;
```
johnbland-wf: Doesn't this fall back to the antipattern of tracking mounts? 

https://facebook.github.io/react/blog/2015/12/16/ismounted-antipattern.html

jacehensley-wf: No that is the `isMounted` method on a React component. They actually recommend this pattern:

"An easy migration strategy for anyone upgrading their code to avoid isMounted() is to track the mounted status yourself. Just set a _isMounted property to true in componentDidMount and set it to false in componentWillUnmount, and use this variable to check your component's status."

johnbland-wf: Fair enough. Just saw it and noticed.

leviwith-wf:
QA +1

- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #60, opened 2017-04-03
UIP-1164 (OR): Provide easy access to DOM node in ValidationUtil.warn messages

kaanaksoy-wk:
This pull request is connected with [UIP-1164 (WSD)](https://github.com/Workiva/web_skin_dart/pull/716)

## Ultimate problem:
`ValidationUtil.warn` should take an optional `Element` parameter which will be `console.log/.warn'd `to the browser console, allowing consumers to navigate to the offending DOM node in the inspector.

All usages of `ValidationUtil.warn` should be updated to pass in the corresponding DOM node, where applicable.

## How it was fixed:
- Because `ValidationUtil.warn` is often used in components' `render` methods, it cannot use the `findDomNode` method, since it causes the following error message:

> render() should be a pure function of props and state. It should never access something that requires stale data from the previous render, such as refs. Move this logic to componentDidMount and componentDidUpdate instead.

- To get around this issue, `findDomNode` will no longer be called within the `ValidationUtil.warn` method. As a result, the component will be logged to the console rather than the DOM node itself.

- Make the `ValidationUtil.warn` method accept an optional second parameter for a component.
- Following the standard warning message, log the component itself as well as its props in a collapsed group.

## Testing suggestions:
- Verify that all current tests continue to pass.

- N/A

---

### DIFF
--- a/lib/src/util/validation_util.dart
+++ b/lib/src/util/validation_util.dart
@@ -16,6 +16,10 @@ library over_react.validation_util;
 
 import 'dart:html';
 
+import 'package:over_react/src/util/pretty_print.dart';
+import 'package:over_react/src/util/react_wrappers.dart';
+import 'package:react/react.dart' as react;
+
 typedef void ValidationUtilWarningCallback(String message);
 
 /// Utility for logging validation errors or warning.
@@ -31,13 +35,13 @@ class ValidationUtil {
   ///
   ///     assert(ValidationUtil.warn('Some warning message'));
   ///
-  /// Optionally, a component can be passed as the second parameter
+  /// Optionally, a component or element can be passed as the second parameter
   /// to provide additional information in the console.
   ///
   ///     assert(ValidationUtil.warn('Some warning message', component));
   ///
   /// The message will be printed out to the console.
-  static bool warn(String message, [dynamic component]) {
+  static bool warn(String message, [dynamic data]) {
     WARNING_COUNT += 1;
 
     if (onWarning != null) {
@@ -51,11 +55,15 @@ class ValidationUtil {
 
       window.console.warn('VALIDATION WARNING: $message');
 
-      if (component != null) {
+      if (data != null) {
           window.console.groupCollapsed('(Warning info)');
+          window.console.log(data);
 
-          window.console.log(component);
-          window.console.log('props: ${component.props}');
+          if (isValidElement(data)) {
+            window.console.log('props: ${prettyPrintMap(getProps(data))}');
+          } else if (data is react.Component){
+            window.console.log('props: ${data.props}');
+          }
 
           window.console.groupEnd();
       }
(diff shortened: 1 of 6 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/util/validation_util.dart
```diff
@@ -44,6 +50,15 @@ class ValidationUtil {
       }
 
       window.console.warn('VALIDATION WARNING: $message');
+
+      if (component != null) {
+          window.console.groupCollapsed('(Warning info)');
+
+          window.console.log(component);
+          window.console.log('props: ${component.props}');
```
aaronlademann-wf: Can we "pretty print" these props?

greglittlefield-wf: `UiProps.toString` has pretty-printing built in, so this will pretty-print them.

lib/src/util/validation_util.dart
```diff
@@ -28,10 +28,16 @@ class ValidationUtil {
...
   ///
-  /// The message will get print out to the console.
-  static bool warn(String message) {
+  /// Optionally, a component can be passed as the second parameter
+  /// to provide additional information in the console.
+  ///
+  ///     assert(ValidationUtil.warn('Some warning message', component));
+  ///
+  /// The message will be printed out to the console.
+  static bool warn(String message, [dynamic component]) {
```
greglittlefield-wf: If we're expecting a component, we can type this to be `react.Component`.

greglittlefield-wf: Hm, or I wonder if we should accept `ReactElement`s as well (would we ever want to print just the child?)...

Could do something like:
```dart
static bool warn(String message, [dynamic data]) {
  ...

  if (data != null) {
    window.console.groupCollapsed('(Warning info)');
    window.console.log(data);

    if (isValidElement(data)) {
      window.console.log('props: ${prettyPrintMap(getProps(data))}');
    } else if (data is react.Component) {
      window.console.log('props: ${prettyPrintMap(data.props)}');
    }
  }

 ...
}

greglittlefield-wf: Not trying to move the goalposts of this PR, just putting it out there as an idea 😃.

kaanaksoy-wk: @greglittlefield-wf Do I need to use `prettyPrintMap` in both cases? I thought that the `toString` method on `UiProps` already handled that for us.

greglittlefield-wf: Nope, you're right. The only case it'd be needed is if they're using plain react-dart component, but we don't really have to worry about that.

greglittlefield-wf:
+1

aaronlademann-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10
- [x] All unit tests pass
- [x] Dependency Scan Clean

Merging.

PR #63, opened 2017-04-20
UIP-2208 Add a docs.yml

maxwellpeterson-wf:
## Ultimate problem:
We need a `docs.yml` file so that this lib's README can be displayed in the Workiva dev portal.

## How it was fixed:
Added a `docs.yml`

---

### DIFF
--- a/docs.yml
+++ b/docs.yml
@@ -1,0 +1,3 @@
+title: over_react
+base: github:Workiva/over_react/
+src: README.md
--- a/docs.yml
+++ b/docs.yml
@@ -1,3 +1,7 @@
 title: over_react
 base: github:Workiva/over_react/
 src: README.md
+topics:
+  title: Transformer
+  base: github:Workiva/over_react/
+  src: lib/src/transformer/README.md


aviary-wf:
## Raven
Number of Findings: 0

jacehensley-wf:
+1

greglittlefield-wf:
We also have a nested README that's linked to by the main README, and should be inlcuded: https://github.com/Workiva/over_react/blob/master/lib/src/transformer/README.md

Does that need to go into docs.yml?

maxwellpeterson-wf:
@greglittlefield-wf we'd want to include that in the docs.yml only if you want that to show up as a child leaf under `over_react` in the nav tree.

greglittlefield-wf:
Does it not get included if we don't add it? If so, we should add it as a leaf.

Otherwise, I'm fine leaving it out—doesn't matter.

maxwellpeterson-wf:
@greglittlefield-wf added

greglittlefield-wf:
+1

johnbland-wf:
+1 refresh

maxwellpeterson-wf:
Ready for merge :)

jayudey-wf:
QA +1, doc change only

Merging into master

PR #69, opened 2017-05-17
UIP-2303 getSelectionStart Util for CTE Cursor Caching

kaanaksoy-wk:
## Ultimate Problem
In order to implement cursor index caching in CTEs, we need a streamlined function that gives us the `selectionStart` value for any input type. 

The `getSelectionStart` function in this PR handles inconsistent behavior across browsers—Chrome does not support `selectionStart` for number and email inputs—while also properly handling text areas.
## Solution
- Create a `getSelectionStart` function similar to the already existing `setSelectionRange` utility.
- Write tests to verify the functionality of the `getSelectionStart` utility.
## Testing Suggestions
- Verify that tests pass.
## Possible Areas of Regression
- N/A
---
FYA: @aaronlademann-wf @greglittlefield-wf @clairesarsam-wf @joelleibow-wf @jacehensley-wf

### DIFF
--- a/lib/src/util/dom_util.dart
+++ b/lib/src/util/dom_util.dart
@@ -154,6 +154,12 @@ void setSelectionRange(/* TextInputElement | TextAreaElement */Element input, in
   }
 }
 
+/// Custom implementation to avoid issues with `selectionStart` when accessing
+/// an [EmailInputElement] or [NumberInputElement] on Chrome.
+///
+/// If `selectionStart` is not supported on the input, `null` will be returned.
+///
+/// See: <https://bugs.chromium.org/p/chromium/issues/detail?id=324360>
 int getSelectionStart(Element input) {
   if (input is TextAreaElement) {
     return input.selectionStart;
--- a/lib/src/util/dom_util.dart
+++ b/lib/src/util/dom_util.dart
@@ -84,6 +84,10 @@ bool supportsSelectionRange(InputElement element) {
   // Uncomment once https://github.com/dart-lang/sdk/issues/22967 is fixed.
   // if (element is TextInputElementBase) return true;
 
+  if (element is TextAreaElement) {
+    return true;
+  }
+
   final type = element.getAttribute('type');
   return inputTypesWithSelectionRangeSupport.contains(type);
 }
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Raven
Number of Findings: 0

lib/src/util/dom_util.dart
```diff
@@ -149,3 +149,27 @@ void setSelectionRange(/* TextInputElement | TextAreaElement */Element input, in
     throw new ArgumentError.value(input, 'input', 'must be an instance of `TextInputElementBase`, `NumberInputElement` or `TextAreaElement`');
   }
 }
+
+/// Custom implementation to avoid issues with `selectionStart` when accessing
```
clairesarsam-wf: See [the Dartlang style guide here](https://www.dartlang.org/guides/language/effective-dart/documentation#prefer-starting-function-or-method-comments-with-third-person-verbs). This comment would be more compliant as something like:


> Returns the index of the first selected character in [input].
>
>  To workaround [this chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=324360), returns `null` for [EmailInputElement] or [NumberInputElement] on Chrome. `null` will also be returned for any [input] that does not support `selectionStart`.

lib/src/util/dom_util.dart
```diff
@@ -149,3 +149,27 @@ void setSelectionRange(/* TextInputElement | TextAreaElement */Element input, in
...
+/// an [EmailInputElement] or [NumberInputElement] on Chrome.
+///
+/// If `selectionStart` is not supported on the input, `null` will be returned.
+///
+/// See: <https://bugs.chromium.org/p/chromium/issues/detail?id=324360>
+int getSelectionStart(Element input) {
```
clairesarsam-wf: a type of `InputElement` may be more appropriate here instead of `Element`

test/over_react/util/dom_util_test.dart
```diff
@@ -154,31 +154,41 @@ main() {
     });
   });
 
-  group('setSelectionRange', () {
-    test('throws an ArgumentError if called on an unsupported Element type', () {
+  group('setSelectionRange and getSelectionStart: ', () {
```
clairesarsam-wf: I think these tests are better suited to separate groups - although they're related methods, the test descriptions aren't as clear as they could be and separated tests would be more atomic/unit-ish. So something that looks like:

```dart
group('setSelectionRange', () {
  test('throws an ArgumentError if called on an unsupported Element type', () {
    ...
  });
  ...
});

group('getSelectionStart', () {
  test('returns normally if called on an unsupported Element type', () {
    ...
  });
  ...
});
```

lib/src/util/dom_util.dart
kaanaksoy-wk: The reason I used `Element` here is because, from what I understand, `TextAreaElement` does not inherit from `InputElement`, but we still want to support it in this function. It is the same reason why `setSelectionRange` also accepts an `Element` as its parameter.

clairesarsam-wf: Oh, right, that's an annoying quirk. Looks good then!

clairesarsam-wf:
@kaanaksoy-wk Will +1 once test failures are resolved

clairesarsam-wf:
+1

greglittlefield-wf:
I'm going to make this a separate ticket as a subtask of the main ticket, so that the main one can be used for the Web Skin Dart PR.

greglittlefield-wf:
- Code looks good
- Test pass in when run locally in both content-shell and Chrome

+10

leviwith-wf:
QA +10

- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #74, opened 2017-06-12
UIP-2373 Make AbstractTransitionProps available in mixin form

aaronlademann-wf:
## Ultimate problem:
The props in `AbstractTransitionProps` cannot be consumed as a mixin, which makes component extension difficult if the consumer cannot extend directly from `AbstractTransitionProps`.

## How it was fixed:
A props mixin / map view were added, and the existing abstract props class now mixes in the new mixin.

## Testing suggestions:
Verify that tests pass

None expected

---

> __FYA:__ @Workiva/ui-platform-pp

### DIFF
--- a/lib/src/component/abstract_transition_props.dart
+++ b/lib/src/component/abstract_transition_props.dart
@@ -1,0 +1,61 @@
+// Copyright 2017 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+library abstract_transition_props;
+
+import 'dart:collection';
+
+import 'package:over_react/over_react.dart';
+
+/// Props that mirror the implementation of [AbstractTransitionProps], made available as a mixin for components
+/// that cannot extend directly from [AbstractTransitionComponent].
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Raven
Number of Findings: 0

jacehensley-wf:
@aaronlademann-wf Should the UIComponent prop generic be updated to be the mixin instead of the abstract class?

```dart
abstract class AbstractTransitionComponent<T extends AbstractTransitionPropsMixin, S extends AbstractTransitionState>
```

Also IMO having the keywords abstract and mixin in the class name is too much, what about just `TranstionPropsMixin`?

aaronlademann-wf:
@jacehensley-wf anyone extending `AbstractTransitionComponent` should still have their component props extending from `AbstractTransitionProps`, IMO. 

Providing a mixin simply makes it more feasible for consumers to expose / forward the correct props on their custom components that render an underlying composite component that extends from `AbstractTransitionComponent`, as opposed to directly extending from it.

aaronlademann-wf:
@jacehensley-wf addressed the name of the mixin / map view

jacehensley-wf:
Ahh okay, maybe add a doc comment expressing the correct use case of the mixin?

lib/src/component/abstract_transition_props.dart
```diff
@@ -0,0 +1,42 @@
+part of 'package:over_react/src/component/abstract_transition.dart';
```
jacehensley-wf: You can't use the new URI pattern yet 😢 that's why travis keeps failing

aaronlademann-wf: :flips table:

aaronlademann-wf:
@jacehensley-wf addressed feedback

jacehensley-wf:
+1

leviwith-wf:
QA +10

- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #77, opened 2017-06-13
UIP-2114: Update FluxUiComponent subscriptions when props are updated

greglittlefield-wf:
## Ultimate problem:
If a `FluxUiComponent` is rerendered with different `props.store` or different props that affect the return value of `redrawOn`, it does not currently update its subscriptions to look at the latest store. This can happen indirectly as a result of rerendering larger views.

## How it was fixed:
- Re-bind store handlers/subscriptions whenever a component is rerendered with new props, ensuring that the latest are always referenced.
- Add `mustCallSuper` annotations to help consumers detect when they're interfering with existing mount/unmount hooks, as well as the new lifecycle hooks added to support this new behavior
    - __Note:__ If existing consumers don't call super on `componentWillReceiveProps`/`componentDidUpdate`, then they just won't get this improved behavior. However, they may hit state issues if they call super on one and not the other.
- Add test

## Testing suggestions:
* Verify that tests pass

* `FluxUiComponent` rerendering

---

### DIFF
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -166,6 +166,8 @@ abstract class _FluxComponentMixin<TProps extends FluxUiProps> implements Batche
     for (var subscription in _subscriptions) {
       subscription?.cancel();
     }
+
+    _subscriptions = null;
   }
 
   @override
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -139,7 +139,7 @@ abstract class _FluxComponentMixin<TProps extends FluxUiProps> implements Batche
 
   /// Subscribe to all applicable stores.
   ///
-  /// [Store]s returned by [redrawOn] will have their triggers mapped directly to this components
+  /// [Store]s returned by [redrawOn] will have their triggers mapped directly to this component's
   /// redraw function.
   ///
   /// [Store]s included in the [getStoreHandlers] result will be listened to and wired up to their
@@ -178,7 +178,7 @@ abstract class _FluxComponentMixin<TProps extends FluxUiProps> implements Batche
   @override
   void componentWillReceiveProps(Map prevProps) {
     // Unbind store handlers so they can be re-bound in componentDidUpdate
-    // once the new props are available, ensuring the values returned [redrawOn]
+    // once the new props are available, ensuring the values returned by [redrawOn]
     // are not outdated.
     _unbindStoreHandlers();
   }
@@ -192,7 +192,7 @@ abstract class _FluxComponentMixin<TProps extends FluxUiProps> implements Batche
 
   @override
   void componentWillUnmount() {
-    // Ensure that unmounted components don't batch render
+    // Ensure that unmounted components don't batch render.
     shouldBatchRedraw = false;
 
     _unbindStoreHandlers();
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -81,41 +81,71 @@ abstract class FluxUiStatefulComponent<TProps extends FluxUiProps, TState extend
 /// Helper mixin to keep [FluxUiComponent] and [FluxUiStatefulComponent] clean/DRY.
 ///
 /// Private so it will only get used in this file, since having lifecycle methods in a mixin is risky.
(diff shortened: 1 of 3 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/src/component_declaration/flux_component.dart
```diff
@@ -76,46 +101,101 @@ abstract class FluxUiComponent<TProps extends FluxUiProps> extends UiComponent<T
...
+  /// Subscribe to all applicable stores.
+  ///
+  /// [Store]s returned by [redrawOn] will have their triggers mapped directly to this components
+  /// redraw function.
+  ///
+  /// [Store]s included in the [getStoreHandlers] result will be listened to and wired up to their
+  /// respective handlers.
+  void _bindStoreHandlers() {
+    if (_areStoreHandlersBound) {
+      throw new StateError('Store handlers are already bound');
```
aaronlademann-wf: @greglittlefield-wf does this need a unit test?

greglittlefield-wf: The only case this would be hit is if someone doesn't add a super-call.

I wasn't sure if it was worth adding a test around, but I can add one if you'd like.

lib/src/component_declaration/flux_component.dart
```diff
@@ -76,46 +101,101 @@ abstract class FluxUiComponent<TProps extends FluxUiProps> extends UiComponent<T
...
   ///
   /// These subscriptions are canceled when the component is unmounted.
-  List<StreamSubscription> _subscriptions = [];
+  List<StreamSubscription> _subscriptions;
+
+  bool get _areStoreHandlersBound => _subscriptions != null;
+
+  /// Subscribe to all applicable stores.
+  ///
+  /// [Store]s returned by [redrawOn] will have their triggers mapped directly to this components
```
trentgrover-wf: `component's `

test/over_react/component_declaration/flux_component_test.dart
```diff
@@ -164,6 +163,30 @@ void main() {
       await nextTick();
       expect(component.numberOfRedraws, equals(0));
     });
+
+    test('updates the subscriptions registered from redrawOn when rerendered with new props', () async {
+      var stores = new TestStores();
+      var newStores = new TestStores();
+
+      var testJacket = mount<TestRedrawOnComponent>((TestRedrawOn()..store = stores)());
```
greglittlefield-wf: ❤️

lib/src/component_declaration/flux_component.dart
```diff
@@ -76,46 +101,101 @@ abstract class FluxUiComponent<TProps extends FluxUiProps> extends UiComponent<T
...
+
+  @override
+  void componentWillMount() {
+    _bindStoreHandlers();
+  }
+
+  @override
+  void componentWillReceiveProps(Map prevProps) {
+    // Unbind store handlers so they can be re-bound in componentDidUpdate
+    // once the new props are available, ensuring the values returned [redrawOn]
```
evanweible-wf: returned _by_?

lib/src/component_declaration/flux_component.dart
```diff
@@ -76,46 +101,101 @@ abstract class FluxUiComponent<TProps extends FluxUiProps> extends UiComponent<T
...
+  void componentWillMount() {
+    _bindStoreHandlers();
+  }
+
+  @override
+  void componentWillReceiveProps(Map prevProps) {
+    // Unbind store handlers so they can be re-bound in componentDidUpdate
+    // once the new props are available, ensuring the values returned by [redrawOn]
+    // are not outdated.
+    _unbindStoreHandlers();
```
maxwellpeterson-wf: Would it be worth only doing this if `prevProps.store != props.store`?

greglittlefield-wf: We'd have to compare the equality of the return value of `redrawOn` and `getStoreHandlers`, and we don't have access to those updated values until `componentDidUpdate` (since those functions often rely on values within `props`). So, we'd have to store the old values in here and then check them and null them out in `componentDidUpdate`.

We can do that, but I figured that unbinding/rebinding was cheap enough that it was worth doing so to avoid that kind of comparison.

aaronlademann-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [ ] Dev/QA +10
- [x] All unit tests pass
- [x] Dependency Scan Clean

Merging.

PR #81, opened 2017-06-19
UIP-1638: Make transformer output strong mode compliant

jacehensley-wf:
> Solution for https://github.com/Workiva/over_react/issues/14

## Ultimate problem:
Code outputted by generated code was not strong-mode compliant under certain circumstances (concrete classes with generic parameter for props/state type).

## How it was fixed:
- Add untyped generation of `typedPropsFactory` and `typedStateFactory` to get around that.
- Add warning if consumers are implementing their own `typedPropsFactory` or `typedStateFactory`

## Testing suggestions:
- Verify tests pass

- Code generation

---
> __FYA:__ @greglittlefield-wf @aaronlademann-wf @clairesarsam-wf @joelleibow-wf


aviary-wf:
## Raven
Number of Findings: 0

lib/src/transformer/impl_generation.dart
```diff
@@ -273,6 +277,22 @@ class ImplGenerator {
             ' extends Object with $componentClassImplMixinName'
         );
       }
+
+      if (new RegExp(r'typed(Props|State)Factory\(.+\)(\s+)?({|=>)').hasMatch(sourceFile.getText(0))) {
```
greglittlefield-wf: I think it'd be much more robust to use the AST to detect this.

test/integration/strong_mode_test.dart
```diff
@@ -0,0 +1,89 @@
...
+  var files = <String>[]
+    ..addAll(new Directory(projectPath).listSync()
+        .where((e) => FileSystemEntity.isFileSync(e.path) && e.path.endsWith('.dart'))
+        .map((e) => e.path)
+    );
+
+  var args = ['--strong']
+    ..addAll(files);
+
+  return await Process.run('dartanalyzer', args);
```
greglittlefield-wf: You can just give `dartanalyzer` a path: 
```dart
return await Process.run('dartanalyzer', ['--strong', projectPath]);
```

test/integration/strong_mode_test.dart
```diff
@@ -0,0 +1,89 @@
...
+
+  return lines.join('\n');
+}
+
+main() {
+  group('OverReact strong mode compliance:', () {
+    test('generates strong mode compliant code', () async {
+      var testProject = copyProject(strongModeProject);
+
+      // Run `pub get` and the bootstrap up-front so that back-to-back runs on the same directory complete faster.
```
greglittlefield-wf: I think this comment was relevant to this code being run in `setUpAll`, and is no longer applicable here.

test_fixtures/strong_mode/web/generic_component.dart
```diff
@@ -0,0 +1,13 @@
+import 'package:over_react/over_react.dart';
+
+@Factory()
+UiFactory<GenericStatelessProps> GenericStateless;
+
+@Props()
+class GenericStatelessProps extends UiProps {}
+
+@Component()
+class GenericStatelessComponent<T extends GenericStatelessProps> extends UiComponent<T> {
```
greglittlefield-wf: We should add a test fixture for a component with generic state as well, to fully test the generation of `typedStateFactory` in that scenario.

lib/src/transformer/impl_generation.dart
greglittlefield-wf: Could do something like this:
```dart
typedMap.node.members.any((member) => member is MethodDeclaration && !member.isStatic && (
    member.name.name == 'typedPropsFactory' ||
    member.name.name == 'typedStateFactory'
));
```

jacehensley-wf: Yeah I like that a lot better! Thanks

test/integration/strong_mode_test.dart
jacehensley-wf: It keeps timing out when I do that

jacehensley-wf:
@greglittlefield-wf Feedback has been addressed/replied to

test/integration/strong_mode_test.dart
greglittlefield-wf: Hm, interesting.

I'd try using the project path as the working directory for the command. If that doesn't work, that's fine, but we should have a comment here as to why we're listing out all of the files.

greglittlefield-wf:
- Code looks good
- Tests pass

+10

test/integration/strong_mode_test.dart
jacehensley-wf: Addressed!

aaronlademann-wf:
QA +1

- [x] Testing instruction
    - All tests pass.
    - Tests include new test to verify that transformer output is strong-mode clean.    
- [x] Dev +1 or +10
- [x] All unit tests pass
- [x] Dependency Scan Clean

Merging.

PR #83, opened 2017-06-21
UIP-2430 Use over_react_test, rip out duplicated test utils

greglittlefield-wf:
## Ultimate problem:
We had duplicated code for test utils that has since made its way into over_react_test.

## How it was fixed:
Pull in over_react_test, rip out duplicated test utils

## Testing suggestions:
- Verify all tests pass in content-shell and chrome (should work on either Dart 1.23 or 1.24)
    - `ddev test -p content-shell`
    - `ddev test -p chrome`

## Potential areas of regression:
Tests

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @jacehensley-wf @clairesarsam-wf @joelleibow-wf

PR #86, opened 2017-06-22
UIP-2413: Get tests passing with the DDC

jacehensley-wf:
> Dependent on <s>https://github.com/Workiva/over_react/pull/83</s> and <s>https://github.com/cleandart/react-dart/pull/122</s>

## Ultimate problem:
Tests were not passing with the DDC

## How it was fixed:
- Update errors that came up from running the tests with the DDC

## Testing suggestions:
- Verify tests by:
  - Run `pub serve test --web-compiler=dartdevc` in one terminal
  - And `pub run test -p vm -p chrome -x no-ddc --pub-serve=8080 test/vm_tests/ test/over_react_component_declaration_test.dart test/over_react_component_test.dart test/over_react_util_test.dart` in another terminal

N/A

---


aviary2-wf:
## Raven
Number of Findings: 0

test/over_react/component/dom_components_test.dart
```diff
@@ -25,15 +25,20 @@ import 'package:over_react/over_react.dart';
...
-    List<MethodMirror> methods = domClassMirror.staticMembers.values.toList();
+    List<MethodMirror> methods = [];
+    ClassMirror domClassMirror;
 
     setUpAll(() {
-      expect(methods, isNotEmpty, reason: 'should have properly reflected the static members.');
+      domClassMirror = reflectClass(Dom);
+
+      // staticMembers is not implemented for the DDC and will throw is this test is loaded even if it's not run.
+      try {
```
greglittlefield-wf: We shouldn't need this try-catch, then, if it's not run in the DDC, right?

test/over_react/component/dom_components_test.dart
```diff
@@ -57,5 +62,5 @@ main() {
         expect(component.type, equalsIgnoringCase(expectedTagName));
       });
     }
-  });
+  }, tags: 'no-ddc');
```
greglittlefield-wf: Why is this disabled in the DDC? Should have a comment

test/over_react/component/resize_sensor_test.dart
```diff
@@ -437,6 +437,6 @@ void main() {
 
     group('common component functionality:', () {
       commonComponentTests(ResizeSensor);
-    });
+    }, tags: 'no-ddc');
```
greglittlefield-wf: Why is this disabled in the DDC? Should have a comment

test/over_react/shared/map_proxy_tests.dart
```diff
@@ -126,7 +126,7 @@ void mapProxyTests(Map mapProxyFactory(Map proxiedMap)) {
       expect(proxy.values, equals(['value']));
       verify(mock.values);
     });
-  });
+  }, tags: 'no-ddc');
```
greglittlefield-wf: Why is this disabled in the DDC? Should have a comment

test/over_react/util/event_helpers_test.dart
```diff
@@ -73,7 +73,7 @@ main() {
     expect(syntheticKeyboardEvent.metaKey, isFalse);
     expect(syntheticKeyboardEvent.repeat, isFalse);
     expect(syntheticKeyboardEvent.shiftKey, isFalse);
-  });
+  }, tags: 'no-ddc');
```
greglittlefield-wf: Why is this disabled in the DDC? Should have a comment

test/over_react/util/event_helpers_test.dart
```diff
@@ -126,7 +126,7 @@ main() {
     expect(syntheticMouseEvent.screenX, isNull);
     expect(syntheticMouseEvent.screenY, isNull);
     expect(syntheticMouseEvent.shiftKey, isFalse);
-  });
+  }, tags: 'no-ddc');
```
greglittlefield-wf: Why is this disabled in the DDC? Should have a comment

test/over_react/component/dom_components_test.dart
jacehensley-wf: It throws even if the test isn't run 😢

test/over_react/component/dom_components_test.dart
jacehensley-wf: Can't use `staticMethod`. I'll add a comment

test/over_react/component/resize_sensor_test.dart
jacehensley-wf: Can't use `instanceMembers` in the DDC. I'll add a comment

test/over_react/shared/map_proxy_tests.dart
jacehensley-wf: Mocks don't work in the DDC. I'll add a comment

test/over_react/component/dom_components_test.dart
greglittlefield-wf: Wait, the contents of `setUpAll` are evaluated even when the group is disabled? Is that a test package bug?

jacehensley-wf: I don't know if it's evaluated, I think the DDC might just be checking for instances of that being used, because it throws if it's in a test too

jacehensley-wf: But if I use `skip:` it's fine. So maybe it is a test package bug.

test/over_react/component/dom_components_test.dart
greglittlefield-wf: Hm, we should really still test these in the DDC if possible. 

What about if we enumerated these, and just checked them against the reflected values, like so?
```dart
const expectedStaticMethods = const <Symbol> [#div, #span, ...];
var actualMethods;
try {
  actualMethods = domClassMirror.staticMembers.values.map((m) => m.simpleName).toList();
} catch(e) {}
if (actualMethods != null) {
  expect(actualMethods, unorderedEquals(expectedStaticMethods),
      reason: '`expectedStaticMethods` needs to be updated');
}

for (var method in expectedStaticMethods) {
  String name = MirrorSystem.getName(method);
...
```

test/over_react/component_declaration/transformer_integration_tests/component_integration_test.dart
```diff
@@ -22,12 +22,6 @@ import '../../../test_util/test_util.dart';
 
 main() {
   group('transformed component integration:', () {
-    test('props class cannot be instantiated directly', () {
```
clairesarsam-wf: Why was this test removed?

jacehensley-wf: Prop classes are made abstract by the transformer and the DDC was failing on instantiating an abstract object.

greglittlefield-wf: And this test was leftover from before that was the case, so it's not really necessary anymore

clairesarsam-wf: 👍

aaronlademann-wf:
+1

greglittlefield-wf:
+1

leviwith-wf:
QA +10

- [X] Testing instruction
- [X] Dev +1's
- [X] Dev/QA +10
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #91, opened 2017-06-29
UIP-2161 Implement DisposableManagerV3 for UiComponent

bencampbell-wf:
## Ultimate problem:

Consumers are required to manually deal with lifecycle management within the component lifecycle. This either requires mixing lifecycle management functionality into all of their components or independently managing the lifetime of `StreamControllers`, `StreamSubscriptions, etc in each component. 

## How it was fixed:

The DisposableManagerV3 interface has been mixed into the base implementation of `UiComponent`. This interface proxies all calls to an internal instance of `w_common`'s Disposable. The call to dispose is tied to the `componentWillUnmount` call ensuring that managed objects are correctly cancelled or concluded when the component is removed from the DOM.

## Testing suggestions:

- The disposable functionality has been exercised in the unit tests. The CI run should pass to confirm this functionality is correct.

- An override of `react.Component.componentWillUnmount`  was added to the `UiComponent` implementation with an annotation that super must be called. There is concern how this may affect current consumers extending `UiComponent` and are already overriding `componentWillUnmount`.

---

### DIFF
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -316,7 +316,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   TState newState() => typedStateFactory({});
 
   //
-  //   END Typed props helpers
+  //   END Typed state helpers
   // ----------------------------------------------------------------------
   // ----------------------------------------------------------------------
 }
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -264,7 +264,7 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component imple
   }
 
   //
-  //   END Typed props helpers
+  //   END DisposableManagerV3 interface implementation
   // ----------------------------------------------------------------------
   // ----------------------------------------------------------------------
 }
@@ -316,7 +316,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   TState newState() => typedStateFactory({});
 
   //
-  //   END DisposableManagerV3 interface implementation
+  //   END Typed props helpers
   // ----------------------------------------------------------------------
   // ----------------------------------------------------------------------
 }
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -316,7 +316,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   TState newState() => typedStateFactory({});
 
   //
-  //   END Typed state helpers
+  //   END DisposableManagerV3 interface implementation
   // ----------------------------------------------------------------------
   // ----------------------------------------------------------------------
 }
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -14,6 +14,8 @@
 
 library over_react.component_declaration.component_base;
 
+import 'dart:async';
+
 import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart' show
     ClassNameBuilder,
@@ -31,6 +33,7 @@ import 'package:over_react/src/component_declaration/component_type_checking.dar
 import 'package:over_react/src/util/ddc_emulated_function_name_bug.dart' as ddc_emulated_function_name_bug;
 import 'package:react/react.dart' as react;
 import 'package:react/react_client.dart';
+import 'package:w_common/disposable.dart';
 
 export 'package:over_react/src/component_declaration/component_type_checking.dart' show isComponentOfType, isValidElementOfType;
 
@@ -94,8 +97,12 @@ typedef TProps BuilderOnlyUiFactory<TProps extends UiProps>();
 ///
 /// Extends [react.Component].
 ///
+/// Implements [DisposableManagerV3]
+///
 /// Related: [UiStatefulComponent]
-abstract class UiComponent<TProps extends UiProps> extends react.Component {
+abstract class UiComponent<TProps extends UiProps> extends react.Component implements DisposableManagerV3 {
+  Disposable _disposableProxy;
+
   /// The props for the non-forwarding props defined in this component.
   Iterable<ConsumedProps> get consumedProps => null;
 
@@ -156,6 +163,12 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
     validateRequiredProps(props);
   }
 
+  @override
+  @mustCallSuper
+  void componentWillUnmount() {
+    _disposableProxy?.dispose();
+  }
+
 
   // ----------------------------------------------------------------------
   // ----------------------------------------------------------------------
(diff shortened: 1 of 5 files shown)


aviary-wf:
## Raven
Number of Findings: 0

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
   // ----------------------------------------------------------------------
   // ----------------------------------------------------------------------
+
+  // ----------------------------------------------------------------------
+  // ----------------------------------------------------------------------
+  //   BEGIN DisposableManagerV3 interface implementation
+  //
+
+  @override
+  Future<T> awaitBeforeDispose<T>(Future<T> future) =>
```
aaronlademann-wf: Should we make the parameter covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+    _getDisposableProxy().awaitBeforeDispose<T>(future);
+
+  @override
+  Future<T> getManagedDelayedFuture<T>(Duration duration, T callback()) =>
```
aaronlademann-wf: Should we make the callback param covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+  @override
+  Timer getManagedPeriodicTimer(Duration duration, void callback(Timer timer)) =>
+    _getDisposableProxy().getManagedPeriodicTimer(duration, callback);
+
+  @override
+  Timer getManagedTimer(Duration duration, void callback()) =>
+    _getDisposableProxy().getManagedTimer(duration, callback);
+
+  @override
+  Completer<T> manageCompleter<T>(Completer<T> completer) =>
```
aaronlademann-wf: Should we make the parameter covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+    _getDisposableProxy().manageCompleter<T>(completer);
+
+  @override
+  void manageDisposable(Disposable disposable) =>
```
aaronlademann-wf: Should we make the parameter covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+    _getDisposableProxy().manageDisposable(disposable);
+
+  @override
+  void manageDisposer(Disposer disposer) =>
```
aaronlademann-wf: Should we make the parameter covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+    _getDisposableProxy().manageDisposer(disposer);
+
+  @override
+  void manageStreamController(StreamController controller) =>
```
aaronlademann-wf: Should we make the parameter covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+    _getDisposableProxy().manageStreamController(controller);
+
+  @override
+  void manageStreamSubscription(StreamSubscription subscription) =>
```
aaronlademann-wf: Should we make the parameter covariant?

lib/src/component_declaration/component_base.dart
```diff
@@ -199,6 +212,61 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component {
...
+  /// [DisposableManagerV3] method.
+  Disposable _getDisposableProxy() {
+    if (_disposableProxy == null) {
+      _disposableProxy = new Disposable();
+    }
+    return _disposableProxy;
+  }
+
+  //
+  //   END Typed props helpers
```
aaronlademann-wf: #nit 

```dart
//   END DisposableManagerV3 interface implementation
```

lib/src/component_declaration/component_base.dart
```diff
@@ -248,7 +316,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   TState newState() => typedStateFactory({});
 
   //
-  //   END Typed state helpers
+  //   END DisposableManagerV3 interface implementation
```
aaronlademann-wf: I believe this is accidental - and should be on line 267 instead?

bencampbell-wf:
@aaronlademann-wf I don't see how covariant would hurt anything given the consumption of the `DisposableManager` interface is internal. At the same time I unable to see a use case where a consumer would need covariant. I'd be glad to make all the parameters covariant if that is a common pattern through the library.

lib/src/component_declaration/component_base.dart
bencampbell-wf: I am unsure if/how inheritance would work in the case of typedefs. Does covariant make sense in that case?

lib/src/component_declaration/component_base.dart
jacehensley-wf: I don't think it should be covariant as this should most likely always accept as `Future`.

lib/src/component_declaration/component_base.dart
aaronlademann-wf: oh... didn't realize `Disposer` was a typedef - not a class.

greglittlefield-wf:
@aaronlademann-wf I don't think there's a need to use use covariant, and adding it makes it easier for consumers to accidentally/unintentionally violate the liskov substitution principle.

If consumers need covariant, they can add it on their overrides.

lib/src/component_declaration/component_base.dart
```diff
@@ -248,7 +316,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   TState newState() => typedStateFactory({});
 
   //
-  //   END Typed state helpers
+  //   END Typed props helpers
```
greglittlefield-wf: Looks like this got changed accidentally

bencampbell-wf: 😰

greglittlefield-wf:
+1

lib/src/component_declaration/component_base.dart
```diff
@@ -31,6 +33,7 @@ import 'package:over_react/src/component_declaration/component_type_checking.dar
 import 'package:over_react/src/util/ddc_emulated_function_name_bug.dart' as ddc_emulated_function_name_bug;
 import 'package:react/react.dart' as react;
 import 'package:react/react_client.dart';
+import 'package:w_common/disposable.dart';
```
robbielamb-wf: should this pull in w_common/disposable_browser.dart instead of just disposable.dart?

johnbland-wf: +1

bencampbell-wf: It is dependent on updating the the DisposableManager interface to include the new browser specific utility functions.

bencampbell-wf: I have created [RAP-2130](https://jira.atl.workiva.net/browse/RAP-2130) to address this.

jacehensley-wf:
+1

aaronlademann-wf:
QA +10

- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10
- [x] All unit tests pass
- [x] Dependency Scan Clean

Merging.

PR #112, opened 2017-09-21
Address Dart 2 infos

jacehensley-wf:
## Ultimate Problem
Dart 2 dev just got released, and when pulling it into OverReact there are some new infos.

## Solution
- Address all infos

## Testing Suggestions
- Run
```bash
brew upgrade dart --devel --with-content-shell --with-dartium
```
- Verify you have Dart version `2.0.0-dev.1.0`
- Verify that there are no infos

## Possible Areas of Regression
- N/A

---
FYA: @Workiva/ui-platform-pp

### DIFF
--- a/lib/src/util/map_util.dart
+++ b/lib/src/util/map_util.dart
@@ -32,7 +32,7 @@ Map getPropsToForward(Map props, {
     Iterable keysToOmit,
     Iterable<Iterable> keySetsToOmit
 }) {
-  Map propsToForward = new Map.from(props);
+  var propsToForward = new Map<String, dynamic>.from(props);
 
   if (omitReactProps) {
     propsToForward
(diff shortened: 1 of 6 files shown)


aviary-wf:
## Raven
<details>

</details>

lib/src/util/map_util.dart
```diff
@@ -56,13 +56,13 @@ Map getPropsToForward(Map props, {
   }
 
   if (onlyCopyDomProps) {
-    new List.from(propsToForward.keys).forEach((String key) {
-      if (key.startsWith('aria-')) return;
-      if (key.startsWith('data-')) return;
-      if (_validDomProps.contains(key)) return;
+    for (var key in new List.from(propsToForward.keys)) {
```
aaronlademann-wf: The last time we changed a `forEach` to a `for` loop... we broke a bunch of stuff... can we verify that we didn't do that again here? _(I don't recall specifically what went wrong last time)_

johnbland-wf: I believe it was this same sort of loop too where you had the list manipulation within the declaration.

@clairesarsam-wf found/fixed it so she might recall the specifics.

greglittlefield-wf: It was due to `return` not being swapped out with `continue`

johnbland-wf: That's right.

PR #118, opened 2017-10-03
Add BuiltReduxUiComponent

jacehensley-wf:
## Ultimate problem:
We should have support for the [built_redux](https://github.com/davidmarne/built_redux) library.

## How it was fixed:
- Add `BuiltReduxUiProps` and `BuiltReduxUiComponent` that wire up listeners to the store's stream.

## Testing suggestions:
- Verify tests pass

- N/A All new stuff

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @clairesarsam-wf @joelleibow-wf
> __FYI:__ @davidmarne-wf


aviary2-wf:
## Raven
<details>

</details>

pubspec.yaml
```diff
@@ -9,6 +9,8 @@ environment:
 dependencies:
   analyzer: ">=0.30.0 <0.31.0"
   barback: "^0.15.0"
+  built_redux: ^6.1.0
+  built_value: ^4.2.0
```
greglittlefield-wf: Is this dependency needed?

pubspec.yaml
```diff
@@ -9,6 +9,8 @@ environment:
 dependencies:
   analyzer: ">=0.30.0 <0.31.0"
   barback: "^0.15.0"
+  built_redux: ^6.1.0
```
greglittlefield-wf: This dependency worries me since it's undergoing major version bumps quite frequently. If we include this, that means we have to stay on top of bumping the upper bound to avoid being a source of version lock.

pubspec.yaml
jacehensley-wf: Hmm I guess it's only needed for tests, will move

pubspec.yaml
jacehensley-wf: Yeah, that's a good point

greglittlefield-wf:
Is this set of base classes a pattern that's also being used in comments? Trying to get a feel for how mature this approach is, and whether more exploration is needed before committing to this API.

jacehensley-wf:
I believe w_comments it just listening to the streams on mount. Let me double check

greglittlefield-wf:
And to clarify, I'm not against this approach, I just want to make sure it's tried and true before adding it to this repo.

jacehensley-wf:
@greglittlefield-wf You can see their usage [here](https://github.com/Workiva/w_comments/search?utf8=%E2%9C%93&q=%22stream.listen%22&type=)

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,163 @@
...
+import 'package:meta/meta.dart';
+import 'package:over_react/over_react.dart';
+
+/// Builds on top of [UiProps], adding typed props for [Store] in order to integrate with `built_redux`.
+///
+/// Use with the over_react transformer via the `@Props()` ([annotations.Props]) annotation.
+abstract class ReduxUiProps<StoreT> extends UiProps {
+  String get _storePropKey => '${propKeyNamespace}store';
+
+  /// The prop defined by [StoresT].
```
aaronlademann-wf: s/StoresT/StoreT

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,163 @@
...
+  ///
+  /// This object should either be an instance of [Store].
```
aaronlademann-wf: either an instance of [Store] ... or ... what else? ;)

test/over_react/component_declaration/redux_component_test/test_reducer.g.dart
```diff
@@ -0,0 +1,222 @@
+// GENERATED CODE - DO NOT MODIFY BY HAND
```
aaronlademann-wf: @jacehensley-wf is it necessary to commit generated files?

tool/build.dart
```diff
@@ -0,0 +1,19 @@
+import 'dart:async';
+
+import 'package:build_runner/build_runner.dart';
+import 'package:built_value_generator/built_value_generator.dart';
+import 'package:source_gen/source_gen.dart';
+import 'package:built_redux/generator.dart';
+
+/// Build the generated files in the built_value chat example.
```
aaronlademann-wf: What chat example?

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,163 @@
+// Copyright 2016 Workiva Inc.
```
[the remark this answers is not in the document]
aaronlademann-wf: Same goes for all files added in this PR

pubspec.yaml
johnbland-wf: It only looks that way because, per @davidmarne-wf, because he went 1.0 too early and broke some things so he had to rev for semver purposes.

test/over_react/component_declaration/redux_component_test/test_reducer.g.dart
johnbland-wf: If you don't commit, build it into your build process to build before tests run.

test/over_react/component_declaration/redux_component_test/default.dart
```diff
@@ -0,0 +1,22 @@
...
+@Component()
+class TestDefaultComponent extends ReduxUiComponent {
+  int numberOfRedraws = 0;
+
+  @override
+  render() => Dom.div()();
+
+  @override
+  void setState(_, [callback()]) {
+    print('here');
```
johnbland-wf: log

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,163 @@
...
+  }
+
+  @mustCallSuper
+  @override
+  void componentWillReceiveProps(Map nextProps) {
+    super.componentWillReceiveProps(nextProps);
+    // Unbind store subs so they can be re-bound in componentDidUpdate
+    // once the new props are available, ensuring the values returned by [redrawOn]
+    // are not outdated.
+    _tearDownSubs();
```
johnbland-wf: If the store prop is the same, we don't need new subs, right?

johnbland-wf:
@greglittlefield-wf, the preferred use in Redux.js is to use a HOC called a provider and connect the data primitive components. In Dart, we don't have that until react-dart supports context.

We're starting to use this in grc and we're looking at a wrapper component vs a base component. The typical flow, per David, is to have a wrapper component handle the subscription and wire up the child components from there. This PR would be used as base class(es) we extend for our wrapper component.

johnbland-wf:
Oh and you really should bring over the pure component in Comments. If we're using built-redux, a `PureReduxComponent` should exist for everyone.

test/over_react/component_declaration/redux_component_test/test_reducer.g.dart
jacehensley-wf: Yeah, I can just build it during CI, but while deving you'll have analysis warnings till you generate

lib/src/component_declaration/redux_component.dart
jacehensley-wf: It's not worth checking that, it doesn't take much to teardown and re-bind the subs

jacehensley-wf:
@johnbland-wf We can't do the HOC until we add support for React's context to react-dart.

For the `PureReduxComponent` I could see that being useful but I'd rather just get this in for now, and adding something like that later.

pubspec.yaml
davidmarne-wf: Yeah, I understand your hesitation for sure, but I really do believe it finally settled down. I'm very happy with where it stands right now.

Plus at this point it is a dependency for CDS, w_comments, review_bar, and soon to be graph_app (and many other repos via transitivity from w_comments) so any major bumps are already going to have to be dealt with anyways. 

However, if we want I could easy make another repo that just has this component.

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,163 @@
...
+
+  @mustCallSuper
+  @override
+  void componentWillUnmount() {
+    super.componentWillUnmount();
+    _tearDownSubs();
+  }
+}
+
+abstract class _ReduxComponentMixin<T extends ReduxUiProps> implements UiComponent<T> {
```
davidmarne-wf: I think it would make a lot of sense to have a connect function here. The default implementation could be something like

```dart
Substate connect(State state) => state;
```

But then that could be overridden to only redraw when specific pieces of your state change

```dart
@override
Substate connect(State state) => state.myComponentWantsToRedrawOn;
```

then redrawOn would be something like
```dart
List<Stream> redrawOn() {
     if (props.store is Store) {
       return <Stream>[props.store.nextSubstate(connect)];
     } else {
       return <Stream>[];
     }
  }
```

jacehensley-wf: With that how would someone listen to multiple stores?

johnbland-wf: There is no concept of multiple stores. It is one store in redux.

johnbland-wf:
@jacehensley-wf yes (re: react context). I mentioned that in hopes @greglittlefield-wf would find it in his heart to push that forward. ;)

Agreed on `PureReduxComponent` being separate. Just getting the list out there. ;)

greglittlefield-wf:
My main concern was marching forward with this without thoroughly planning this feature, but it seems like there's a consensus that this is what we want 👍.

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,163 @@
...
+  StoreT get store => props[_storePropKey];
+  set store(StoreT value) => props[_storePropKey] = value;
+}
+
+/// Builds on top of [UiComponent], adding `built_redux` integration.
+///
+/// * Redux components are responsible for rendering application views and turning
+///   user interactions and events into [Action]s.
+/// * Redux components can use data from one or many [Store] instances to define
+///   the resulting component.
```
evanweible-wf: Just one store, right?

johnbland-wf: Yup

lib/src/component_declaration/redux_component.dart
johnbland-wf: I'm curious why it isn't worth it. It seems like you'd want to not force something that is not necessary.

greglittlefield-wf: Tearing down and rebinding subscriptions is cheap, I don't think it's necessary to expend effort to avoid it.

Also, we we'd have to compare the return value of `redrawOn` and not just `props.store`, and we don't have access to that updated return value until `componentDidUpdate` (since that function can rely on values within `props`). So, we'd have to store the old values in here and then check them and null them out in componentDidUpdate.

In this case, unbinding/rebinding is more straightforward, and again, should cheap enough to be worth it.

lib/src/component_declaration/redux_component.dart
jacehensley-wf: True, what about multiple values on the single store?

johnbland-wf: Agreed. We'll want a substate on multiple values.

davidmarne-wf: So one thing you can do is map all the values into a new built_value. For example say by store's state object is:
```dart
abstract class AppState implements Built<AppState, AppStateBuilder> {
  int get foo;
  int get bar;
  int get baz;
}
```

But my component only wants foo and bar. I can still only using a single nextSubstate by defining a new model that contains the subset of state my component cares about and having my connect function return that. So i would define

```dart
abstract class DataComponentWants implements Built<DataComponentWants, DataComponentWantsBuilder> {
  int get foo;
  int get bar;
}
```

and have connect do:
```dart
DataComponentWants connect(AppState state) => new DataComponentWants(
  foo: state.foo,
  bar: state.bar,
);
```

davidmarne-wf: Since DataComponentWants is also a built, it should be comparable with the last result of connect, then we can limit redraws by only redrawing if the last result == new result

lib/src/component_declaration/redux_component.dart
davidmarne-wf: I'm confused about this, tracking _areStoreSubsBound and having all these teardown and setups it seems overly complicated to me. 
The reference to your store should never change, it seems to me that you would only setup subs on componentWillMount and remove them on componentWillUnmount.

davidmarne-wf:
I threw up a proposal on how I could see this working here:
https://github.com/davidmarne-wf/over_react/pull/1/files

The generics get a bit verbose, but it is necessary to get legit type safety.

jacehensley-wf:
@davidmarne-wf How does this look?

lib/src/component_declaration/redux_component.dart
```diff
@@ -47,117 +51,68 @@ abstract class ReduxUiProps<StoreT> extends UiProps {
...
-
-    _storeSubs = [];
-
-    for (var stream in redrawOn()) {
-      _storeSubs.add(stream.listen((_) => redraw()));
-    }
+    _storeSub = props.store.nextSubstate(connect).listen((Substate s) {
+      _connectedProps = s;
+      redraw();
+    });
```
johnbland-wf: I think the only problem with this is we can only listen to one substate so we'd be stuck if we wanted `state.count` and `state.todos`.

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,159 @@
...
+  ///         baz: baz,
+  ///       );
+  ///       MyState._();
+  ///
+  ///       int get foo;
+  ///       int get bar;
+  ///       int get baz;
+  ///     }
+  ///
+  ///     abstract class DataCoponentCaresAbout implements Built<DataCoponentCaresAbout, DataCoponentCaresAboutBuilder> {
```
johnbland-wf: `DataCoponent` -> `DataComponent`

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,159 @@
...
+  ///       render() {
+  ///         return Dom.dov()(
+  ///           connectedState.foo,
+  ///           connectedState.bar,
+  ///         );
+  ///       }
+  ///     }
+  ///
+  /// Related: [connectedState]
+  Substate connect(V state);
```
davidmarne-wf: I think it would be worth adding to the docstring that Substate must be a comparable object to avoid unnecessary redraws. Add a recommendation that Substate should be either primitive, built_value, built_collection, or other type of object that overrides the == operator.

davidmarne-wf: as well as a similar blurb to the ReduxUiComponent docstring

lib/src/component_declaration/redux_component.dart
johnbland-wf: Thoughts, @davidmarne-wf?

jacehensley-wf: No, your substate would be a built_value object like shown in the doc comment https://github.com/Workiva/over_react/pull/118/files/20dc99def5af7766e5045dbcc909ad3bb39db8f1#diff-cd6ab537d255abbf1ec247df0369120eR119

davidmarne-wf: i am of the opinion you define a new built_value and map state to that. so say your state is
```dart
abstract class State extends Built<State, StateBuilder> {
  int get count;
  BuiltList<Todo> todos;
  Map<String, String> options;
 ...
}
```
you define a model
```dart
abstract class SubstateMyComponentWants extends Built<SubstateMyComponentWants, SubstateMyComponentWantsBuilder> {
  int get count;
  BuiltList<Todo> todos;
 ...
}
```

then your connect is 
```dart
@override
SubstateMyComponentWants connect(State s) => new SubstateMyComponentWants((builder) => builder
  ..count = state.count
  ..todos = state.todos);
```

davidmarne-wf: I'ld be open to suggestions for multiple connectors though. I imagine it would look something like:
```dart
typedef Substate StateMapper<State, Substate>(State s);

Iterable<StateMapper<State, dynamic>> get connectors;

...

void _setUpSub() {
    _storeSubs.addAll(connectors.map((connect) => props.store.nextSubstate(connect).listen((_) {
      redraw();
    }));
  }
```

davidmarne-wf: then you wouldn't have connectedState, you would prob just expose a store.state getter and let the renderers pull data from that.

davidmarne-wf: or even simpler you could just have 

```dart
bool shouldUpdateOnReduxStateChange(State s);
```

that the consumer implements.

then if i wanna update on todos or count changes i do:

```dart
@override
bool shouldUpdateOnReduxStateChange(State prev, State next) =>  prev.todos != next.todos || prev.count != next.count
```

The upside to keeping how it is now, where you load everything into a new built_value, is this comparison is generated for the model

jacehensley-wf: I like the first option, with using built_value objects to contain multiple state values

johnbland-wf: That's fine.

jacehensley-wf:
@davidmarne-wf Added that doc comment :)

johnbland-wf:
Maybe this should be renamed `BuiltReduxUiComponent` since it is specific to built_redux only and not redux-dart.

jacehensley-wf:
That's too wordy IMO

johnbland-wf:
I agree. It'll be confusing to someone trying to use Redux directly when this won't work for them.

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+abstract class ReduxUiProps<V extends Built<V, B>, B extends Builder<V, B>,
+    A extends ReduxActions> extends UiProps {
+  String get _storePropKey => '${propKeyNamespace}store';
+
+  /// The [Store] prop defined by [V], [B], and [A].
+  Store<V, B, A> get store => props[_storePropKey];
+  set store(Store<V, B, A> value) => props[_storePropKey] = value;
+
+  /// The [ReduxActions] prop defined by [A].
+  A get actions => store.actions;
```
greglittlefield-wf: Can we add test coverage for this?

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  void componentWillMount() {
+    super.componentWillMount();
+    _setUpSub();
+  }
+
+  @mustCallSuper
+  @override
+  void componentWillReceiveProps(Map nextProps) {
+    super.componentWillReceiveProps(nextProps);
+    _tearDownSub();
```
greglittlefield-wf: Looks like this method is missing test coverage

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  }
+
+  @mustCallSuper
+  @override
+  void componentWillUpdate(Map nextProps, Map nextState) {
+    if (_storeSub == null) _setUpSub(nextProps);
```
greglittlefield-wf: Looks like this method is missing test coverage

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+
+abstract class _ReduxComponentMixin<
+    V extends Built<V, B>,
+    B extends Builder<V, B>,
+    A extends ReduxActions,
+    T extends ReduxUiProps<V, B, A>,
+    Substate> implements UiComponent<T> {
+  Substate _connectedState;
+
+  /// The substate values of the redux store that this component component subscribes to.
```
greglittlefield-wf: typo: `component component`

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  ///
+  /// It is reccommened to use this instead of `props.store.state`,
```
greglittlefield-wf: #nit comma should be a semicolon

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  ///       int get foo;
+  ///       int get bar;
+  ///     }
+  ///
+  ///     MyComponent extends ReduxUiComponent<MyState, MyStateBuilder, MyActions, MyProps, DataComponentCaresAbout> {
+  ///       @override
+  ///       DataComponentCaresAbout connect(state) => new DataComponentCaresAbout(foo: state.foo, bar: state.bar);
+  ///
+  ///       render() {
+  ///         return Dom.dov()(
```
greglittlefield-wf: typo: `dov`

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  ///           connectedState.foo,
+  ///           connectedState.bar,
+  ///         );
+  ///       }
+  ///     }
+  ///
+  /// __Note:__ [Substate] must be a comparable object to avoid unnecessary redraws,
```
greglittlefield-wf: #nit comma should be semicolon

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  /// it is reccomended that [Substate] either be a primitive, built_value, build_collection or an Object
```
greglittlefield-wf: Typo? Is `build_collection` supposed to be `built_collection`?

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+  /// that overrides `==`.
```
greglittlefield-wf: It'd probably be a good idea to mention overriding `hashCode` as well, since it's not immediately obvious to all developers that you need to override both and not just one.

lib/src/component_declaration/redux_component.dart
```diff
@@ -0,0 +1,171 @@
...
+
+  /// The substate values of the redux store that this component component subscribes to.
+  ///
+  /// It is reccommened to use this instead of `props.store.state`,
+  /// if you need to access other state values from the store then they should be connected with [connect].
+  ///
+  /// Related: [connect]
+  Substate get connectedState => _connectedState;
+
+  /// Subscribe to changes to the values from the redux store that this component cares about.
```
greglittlefield-wf: #nit This doesn't follow the Dart style guide convention around the first line of function doc comments.

Should be something like:
> Returns a subset of the values derived from the redux store that this component cares about.

lib/src/component_declaration/redux_component.dart
greglittlefield-wf: Might be good to add a code comment here that explains when this is null (i.e., for component props updates and not just any rerender)

test/over_react/component_declaration/redux_component_test.dart
```diff
@@ -0,0 +1,124 @@
...
+      var store = new Store<BaseState, BaseStateBuilder, BaseActions>(
+        baseReducerBuilder.build(),
+        baseState,
+        baseActions,
+      );
+      var renderedInstance = render(TestDefault()..store = store);
+      TestDefaultComponent component = getDartComponent(renderedInstance);
+
+      store.actions.trigger1();
+      await new Future.delayed(Duration.ZERO);
```
greglittlefield-wf: Is this necessary now that actions dispatch synchronously? Should we set the lower bound to the major version that changes that?

test/over_react/component_declaration/redux_component_test.dart
```diff
@@ -0,0 +1,124 @@
...
+      unmount(renderedInstance);
+
+      store.actions.trigger1();
+      await new Future.delayed(Duration.ZERO);
+
+      expect(component.numberOfRedraws, 1,
+          reason: 'component should no longer be listening after unmount');
+    });
+
+    test('subscribes to any state changes subscribed to in connect', () async {
```
greglittlefield-wf: This description would be clearer and more indicative of what's being tested if it read "only the state changes subscribed to in connect".

test/over_react/component_declaration/redux_component_test.dart
```diff
@@ -0,0 +1,124 @@
...
+      );
+      var renderedInstance = render((TestDefault()..store = store)());
+      TestDefaultComponent component = getDartComponent(renderedInstance);
+
+      store.actions.trigger1();
+      await new Future.delayed(Duration.ZERO);
+      expect(component.numberOfRedraws, 1);
+
+      component.props = TestDefault()..store = updatedStore;
+      component.redraw();
```
greglittlefield-wf: This should be actually rerendering the component and not just simulating a props update.

test/over_react/component_declaration/redux_component_test.dart
jacehensley-wf: I wanted to leave it open to both version to not be blocking, I don't think w_comments is on that major version yet

lib/src/component_declaration/redux_component.dart
jacehensley-wf: Done

lib/src/component_declaration/redux_component.dart
jacehensley-wf: Done

johnbland-wf:
One more attempt for a rename here. ;)

test/over_react/component_declaration/redux_component_test.dart
greglittlefield-wf: 👍 

If you end up making changes in this PR again, perhaps a todo comment would be good to remove these if we ever bump. Waiting extra long shouldn't hurt, but it would be nice to verify that the things we expect to be synchronous are indeed synchronous.

greglittlefield-wf:
+1

brianbolton-wf:
FYI @christianplazas-wf @garywhitlock-wf

aaronlademann-wf:
+1 refresh

greglittlefield-wf:
+1

jayudey-wf:
QA +1

- [ ] Testing instruction
- [ ] Dev +1's
- [ ] Dev/QA +10 with detail of what was tested
  - passing build with tests exercising new functionality
- [ ] Security review (if required)
- [ ] Unit tests created/updated
- [ ] All unit tests pass
- [ ] Rosie has run and reports properly the release the ticket will be included in
- [ ] Areas of regression noted

Merging into master.

PR #119, opened 2017-10-05
UIP-2762 More convenient ubiquitous access of DOM/aria props

greglittlefield-wf:
## Ultimate problem:
_I was bored after work yesterday 😄_

Adding aria/DOM props to components is a little clunky; you have to do something like this:
```dart
(Button()
  ..addProps(domProps()..draggable = true)
  ..addProps(ariaProps()
    ..controls = 'my_popover'
    ..labelledby = 'my_label'
  )
)('Click me')
```

What would be nicer is a syntax like this:
```dart
(Button()
  ..dom.draggable = true
  ..aria.controls = 'my_popover'
  ..aria.labelledby = 'my_label'
)('Click me')
```

## How it was fixed:
- Add `aria` and `dom` getters to `UbiquitousDomPropsMixin` that provide typed views into a given `UiProps` class that allow for reading/writing aria/DOM props.
- Add `doNotGenerate` option to `@Accessor()` annotation to allow declaring fields in accessor (props, props mixin, state, state mixin) classes that shouldn't be generated into key-value-pair-backed getters/setters.
    - Necessary to implement caching for `aria`/`dom` getters.

## Testing suggestions:
- Verify that tests pass.
- Render a component with the above syntax and verify that things work as expected.

This could break consumer code if they've declared props or state fields named `aria` or `dom`.

---

### DIFF
--- a/lib/src/component_declaration/annotations.dart
+++ b/lib/src/component_declaration/annotations.dart
@@ -243,12 +243,16 @@ class Accessor {
   /// The error message displayed when the accessor is not set.
   final String requiredErrorMessage;
 
+  /// Whether to skip generating an accessor for this field.
+  final bool doNotGenerate;
+
   const Accessor({
     this.key,
     this.keyNamespace,
     this.isRequired = false,
     this.isNullable = false,
     this.requiredErrorMessage,
+    this.doNotGenerate = false,
   });
 }
 
--- a/lib/src/component/prop_mixins.dart
+++ b/lib/src/component/prop_mixins.dart
@@ -17,7 +17,7 @@ library over_react.prop_mixins;
 
 // Must import these consts because they are used in the transformed code.
 // ignore: unused_import
-import 'package:over_react/over_react.dart' show PropDescriptor, ConsumedProps;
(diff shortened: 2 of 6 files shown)


aviary-wf:
## Raven
<details>

</details>

lib/src/component_declaration/annotations.dart
```diff
@@ -243,12 +243,16 @@ class Accessor {
   /// The error message displayed when the accessor is not set.
   final String requiredErrorMessage;
 
+  /// Whether to skip generating an accessor for this field.
+  final bool doNotGenerate;
```
jacehensley-wf: Should we add a constant annotation for this? `@doNotGenerate` obviously not needed for this PR but might be nice for later

greglittlefield-wf: Yeah, I thought about adding one but decided against it since this won't be used very often (we haven't had a need for it until now).

jacehensley-wf:
QA +10
- Build successful
- tests pass
----
- [X] Testing instruction
- [X] Dev +1's
- [X] All unit tests pass
- [X] Dependency Scan Clean

Merging.

PR #122, opened 2017-10-17
do not listen when warning that it cannot listen

toddbeckman-wf:
## Ultimate problem:

A check is made to determine whether a store is disposed or disposing, and a warning is provided when a store cannot be listened to, yet despite this warning, an attempt is made to listen anyway.

## How it was fixed:

Do not listen to the store when this warning is provided.

## Testing suggestions:

- Provide a store as a prop to a FluxComponent, and dispose that store before mounting the component.
- When attempting to mount, the warning is provided but no attempt to listen is made.

---

### DIFF
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -168,7 +168,10 @@ abstract class _FluxComponentMixin<TProps extends FluxUiProps> implements Batche
         'in a test environment, try adding an `await window.animationFrame;` before disposing your '
         'store.');
 
-      if (isDisposedOrDisposing) _logger.warning(message);
+      if (isDisposedOrDisposing) {
+        _logger.warning(message);
+        return;
+      }
 
       StreamSubscription subscription = store.listen(handler);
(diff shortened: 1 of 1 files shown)


aviary-wf:
## Raven
<details>

</details>

jacehensley-wf:
Didn't we make it still listen to avoid a breaking change?

greglittlefield-wf:
@jacehensley-wf Listening will result in an exception in that case, so changing this wouldn't be breaking. 

The check and warning was originally added to add more context to the error message, while maintaining existing behavior of throwing when in this state.

See original PR: https://github.com/Workiva/over_react/pull/72.

---

I'm going to raise the same concerns I did in that PR when it was suggested that we only warn, and don't throw on this exception. @toddbeckman-wf could you provide more context as to why you want it to fail silently instead? If this happens, then the component can't really recover, right?

toddbeckman-wf:
After discussion and working around the issue in https://github.com/Workiva/wdesk_sdk/pull/1749, it was decided this fix was not needed.

PR #130, opened 2017-12-29
CPLAT-6182 Use the resize observer API if it is available 

robbecker-wf:
## Overview:
The resize sensor uses scroll events and adds dom elements to detect resizes in events. This works but can cause extra CPU usage and event listeners.

## How it was fixed:
Use the resize observer API if it is available (coming in Chrome 64 Jan 23 2017). Otherwise fallback to existing functionality.
Chrome will start using this when the API becomes available. Other browsers will use the existing code. If we want to get rid of our implementation with scroll events, we could load the resize observer polyfill for browsers that need it when the wdesk shell loads.

## Testing suggestions:
 - [ ] Existing tests should all pass. 
 - [ ] Could manually try out the resize sensor example (pub serve and http://localhost:8080/examples/resize_sensor/)

The resize observer API gives sub-pixel accuracy instead of integer accuracy that we currently have. This *should* be ok but 🤷‍♂️ In the handler, the width and height are rounded to fit the existing `ResizeSensorEvent` integer types.


aviary-wf:
## Raven
<details>

</details>

lib/src/component/resize_sensor.dart
```diff
@@ -106,13 +107,32 @@ class ResizeSensorComponent extends UiComponent<ResizeSensorProps> with _SafeAni
...
   }
 
   @mustCallSuper
   @override
   void componentDidMount() {
+    if (ResizeObserver.supported) {
+      Element el = findDomNode(this);
+      _lastHeight = 0;
+      _lastWidth = 0;
+      ResizeObserver.observe(findDomNode(this), (el, x, y, w, h, t, b, l, r) {
```
johnbland-wf: Use `el` instead of finding the node again.

I'd also suggest a class method vs an anonymous one.

lib/src/component/resize_sensor.dart
```diff
@@ -161,7 +181,9 @@ class ResizeSensorComponent extends UiComponent<ResizeSensorProps> with _SafeAni
       (Dom.div()..style = _collapseSensorChildStyle)()
     );
 
-    var resizeSensor = (Dom.div()
+    // If we have the Resize Observer API, the resize sensor is a no-op
+    // otherwise, wire up the div, expand and collapse sensor just as before
+    var resizeSensor = ResizeObserver.supported ? null : (Dom.div()
```
johnbland-wf: `ResizeObserver.supported` will never be true here, right?

robbecker-wf: It **can** be true here. In the case where we are using the resize observer path, Ive changed the behavior in the `componentDidMount` and `render` methods. In render, it will still render its children but not the extra expandSensor and collapseSensor elements.

PR #131, opened 2018-01-05
Allow usage of ResizeObserver in ResizeSensor

jacehensley-wf:
## Ultimate problem:
Different approach to https://github.com/Workiva/over_react/pull/130 using this [ResizeObserver](https://github.com/JaceHensley/resize_observer_polyfill_wrapper) instead.

This approach will work in all browsers, but it doesn't use the built in functionality in Chrome.

## How it was fixed:
- Add a `props.useResizeObserver` that defaults to `false`
- Refactor `ResizeSensor` to render one of two internal impls
- Update tests

## Testing suggestions:
- Verify tests pass
- Pull this down and mess around with the new `ResizeObserver` component and verify that it works as expected.

- `ResizeSensor`

---
__FYA:__ @greglittlefield-wf @aaronlademann-wf @jacehensley-wf @clairesarsam-wf @joelleibow-wf
__FYI:__ @robbecker-wf


aviary2-wf:
## Raven
<details>

</details>

lib/src/component/resize/resize_props_mixin.dart
```diff
@@ -0,0 +1,87 @@
...
+  ///
+  ///     * [onInitialize] will never be called.
+  ///
+  /// * The sensors will be initialized/reset in the next animation frame after mount, as opposed to synchronously,
+  ///   helping to break up resulting layouts.
+  ///
+  /// Default: false
+  bool quickMount;
+
+  /// Whether to use a polyfilled `ResizeObserver`.
```
aaronlademann-wf: Probably should put a bit more meat on this bone to explain the benefits, drawbacks, etc.

lib/src/component/resize/resize_sensor.dart
```diff
@@ -0,0 +1,317 @@
...
+/// A wrapper component that detects when its parent is resized.
+///
+/// This component _must_ be put in a relative or absolutely positioned
+/// container.
+///
+///     (ResizeSensor()..onResize = () => print('resized'))(children)
+///
+/// See: <https://docs.workiva.org/web_skin_dart/latest/components/#resize-sensor>.
+@Factory()
+UiFactory<ResizeSensorProps> ResizeSensor;
```
aaronlademann-wf: IMO its a little confusing that this factory / props class / component class are named identically to the one we export.

pubspec.yaml
```diff
@@ -40,5 +42,6 @@ transformers:
       ]
   - test/pub_serve:
       $include: test/**_test{.*,}.dart
+  - dart_to_js_script_rewriter
```
aaronlademann-wf: ?

test/over_react_component_test.html
```diff
@@ -21,6 +21,7 @@
     <!-- javascript -->
     <script src="packages/react/react_with_addons.js"></script>
     <script src="packages/react/react_dom.js"></script>
+    <script src="packages/resize_observer_polyfill_wrapper/resize_observer_polyfill.js"></script>
```
aaronlademann-wf: Is this JS file necessary if `props.useResizeObserver` is `true`?

Is there any sort of validation within the component enforcing this?

If not - at the very least - there should be extensive documentation in the doc comment for `props.useResizeObserver` explaining that this is a requirement, IMO.

lib/src/component/resize/resize_sensor.dart
jacehensley-wf: I agree, do you have a suggestion?

test/over_react_component_test.html
jacehensley-wf: It's only necessary if that is `true`.  There's no validation here, just my underlying [package](https://github.com/JaceHensley/resize_observer_polyfill_wrapper)

robbecker-wf: So, would we need to include this JS file in the main shell .html file? Hopefully with feature detection and only load it if it needs the polyfill?

jacehensley-wf: The JS library that I wrapped never uses the native ResizeObserver, for whatever reason, so this JS file will always need to be loaded.

robbecker-wf: oh, that's weird. Isn't the whole point of using the browser built in resize observer to reduce code? This change just replaces the resize detection logic from our Dart polyfill to a JS implemented polyfill then?

jacehensley-wf: Kinda, but it's a lot cleaner of an impl IMO. It doesn't wrap the resizing element and place "resize sensors" around it. And I _think_ it's more performant. I want to try it out with comments and report back on that.

aaronlademann-wf: @jacehensley-wf IMO - if this script is necessary for the component to work when `props.useResizeObserver` is true, then the component should be responsible for dynamically loading / unloading that script into the DOM.

jacehensley-wf: But how does it know where the script is?

robbecker-wf: You could load it as a resource and inject it as a script?
See https://github.com/dart-lang/resource

jacehensley-wf: Interesting, I'll take a stab at that tonight

robbecker-wf: hmm I guess that makes more sense to do in the wrapper dart package

jacehensley-wf: The more I'm thinking about it the less I like the approach of injecting js without the user doing it. I think throwing an error and letting the user add the js file is the best approach.

PR #142, opened 2018-02-26
UIP-3062 Do not execute prop validation logic in dart2js code

aaronlademann-wf:
## Ultimate problem:
The `validateProps` method is currently called in `componentWillMount` and `componentWillReceiveProps` - even in production (dart2js) code.

This means that our internal `validateRequiredProps` is always called when a component mounts / updates - and any consumer validations are also always called if/when they override `validateProps`.

There is a not-insignificant performance cost to this, and there is no reason for prop validation to be happening in a dart2js scenario.  

## How it was fixed:
1. Wrap both calls to `validateProps` in an `assert` to improve performance in dart2js builds.
2. Update tests.

## Testing suggestions:
1. Pull this branch locally
2. Run `pub get` and `ddev test -p chrome`
3. Verify that all tests pass in chrome. 
    * _(NOTE: focusing the chrome window running the tests greatly speeds up test execution)_.

Tests?

---

> __FYA:__ @Workiva/ui-platform-pp @sebastianmalysa-wf @evanweible-wf @corwinsheahan-wf

### DIFF
--- a/lib/over_react.dart
+++ b/lib/over_react.dart
@@ -27,7 +27,7 @@ export 'package:react/react.dart' show
     SyntheticUIEvent,
     SyntheticWheelEvent;
 
-export 'package:react/react_client.dart' show setClientConfiguration, ReactElement, inReactDevMode;
+export 'package:react/react_client.dart' show setClientConfiguration, ReactElement;
 
 export 'src/component/abstract_transition.dart';
 export 'src/component/abstract_transition_props.dart';
--- a/lib/over_react.dart
+++ b/lib/over_react.dart
@@ -27,7 +27,7 @@ export 'package:react/react.dart' show
     SyntheticUIEvent,
     SyntheticWheelEvent;
 
-export 'package:react/react_client.dart' show setClientConfiguration, ReactElement;
+export 'package:react/react_client.dart' show setClientConfiguration, ReactElement, inReactDevMode;
 
 export 'src/component/abstract_transition.dart';
 export 'src/component/abstract_transition_props.dart';
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -197,17 +197,15 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component imple
 
   /// Validates that props with the `@requiredProp` annotation are present.
   void validateRequiredProps(Map appliedProps) {
-    if (inReactDevMode) {
-      consumedProps?.forEach((ConsumedProps consumedProps) {
-        consumedProps.props.forEach((PropDescriptor prop) {
-          if (!prop.isRequired) return;
-          if (prop.isNullable && appliedProps.containsKey(prop.key)) return;
-          if (!prop.isNullable && appliedProps[prop.key] != null) return;
+    consumedProps?.forEach((ConsumedProps consumedProps) {
+      consumedProps.props.forEach((PropDescriptor prop) {
+        if (!prop.isRequired) return;
+        if (prop.isNullable && appliedProps.containsKey(prop.key)) return;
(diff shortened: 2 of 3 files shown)


aviary2-wf:
## Module Findings
No security relevant content detected. Please review for security relevance and request security review as needed.

jacehensley-wf:
I'm worried this is a breaking change, especially since we have to update OverReact Test to accommodate these changes

jacehensley-wf:
So debug and _unset_ are the same behavior, correct?

clairesarsam-wf:
@greglittlefield-wf That seems like a solid idea, though since this is a change with a decent performance impact based on @aaronlademann-wf 's profiling I would really like to be able to have this  as default behavior that consumers don't have to make changes to pull in. Though maybe with unified wdesk that flag would really only need to be set in one place to matter?

greglittlefield-wf:
Ooh, another idea is to detect whether we're using the development or production version of React, and only validate props for the development version (just like how propTypes works in React JS).

jacehensley-wf:
That would be really cool! Is there a variable that React sets when it's in development mode?

greglittlefield-wf:
I don't think there is one that's considered stable; we could add one within the correct JS bundles in react-dart, though.

jacehensley-wf:
Now that is really interesting. We could do the same for test mode and the test ids

greglittlefield-wf:
Yup! Though it'd probably be good to have more granular control as to what's enabled.

For instance, in automated tests (functional or perf-related), you might want to be running code with prod-like performance, but still have test IDs enabled.

lib/over_react.dart
```diff
@@ -27,7 +27,7 @@ export 'package:react/react.dart' show
     SyntheticUIEvent,
     SyntheticWheelEvent;
 
-export 'package:react/react_client.dart' show setClientConfiguration, ReactElement;
+export 'package:react/react_client.dart' show setClientConfiguration, ReactElement, inReactDevMode;
```
greglittlefield-wf: Why are we exporting this? 

I can't really think of a use-case other consumers would have for this that isn't related to `validateProps.`

I see this flag being a means to have behavioral parity with the prop validation that exists in React JS, and I would be concerned about exposing it for use other use-cases that shouldn't be tied to which React build is being used (and should instead be tied to some other flag that's more appropriate).

lib/src/component_declaration/component_base.dart
```diff
@@ -195,16 +195,19 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component imple
...
+  /// Validates that props with the `@requiredProp` annotation are present.
   void validateRequiredProps(Map appliedProps) {
-    consumedProps?.forEach((ConsumedProps consumedProps) {
-      consumedProps.props.forEach((PropDescriptor prop) {
-        if (!prop.isRequired) return;
-        if (prop.isNullable && appliedProps.containsKey(prop.key)) return;
-        if (!prop.isNullable && appliedProps[prop.key] != null) return;
-
-        throw new PropError.required(prop.key, prop.errorMessage);
+    if (inReactDevMode) {
```
greglittlefield-wf: Why do we have a check in this method but not in `validateProps`? Seems like it should be both or neither.

I lean toward neither, to allow consumers to still explicitly perform validation if desired when using the prod bundle, but that validation being turned off "by default" due to the checks in lifecycle methods.

corwinsheahan-wf:
QA +10 CI passes

corwinsheahan-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10 with detail of what was tested
- [x] Security review (if required)
- [x] All unit tests pass
- [x] Rosie has run and reports properly the release the ticket will be included in

Merging into master.
@Workiva/release-management-p

PR #154, opened 2018-05-04
AF-232 Expose render and unmountComponentAtNode

aaronlademann-wf:
## Ultimate problem:
We currently expose most of the commonly consumed methods from the `react` package, but not anything from its `react_dom.dart`.  This forces consumers to import `package:react/react_dom.dart`, and therefore declare `react` as a dependency for their application.

However, this library is designed to act as an improved consumption experience for `react` consumers, which should include insulation from internal breaking changes without having to change the upper bound of a dependency.

## How it was fixed:
1. Proxy the `react_dom.render` and `react_dom.unmountComponentAtNode` methods from `package:react/react_dom.dart`, with improved function / return signatures.
2. Update all internal consumption, and consumption documentation.
3. Write tests.

__For consumers, this should be the extent of the changes necessary to stop depending directly on `react`:__

```diff
- import 'package:react/react_dom.dart' as react_dom;
+ import 'package:over_react/react_dom.dart' as react_dom;
```

## Testing suggestions:
- [ ] Passing CI build

None expected

---

> __FYA:__ @greglittlefield-wf @clairesarsam-wf @sebastianmalysa-wf @dustinlessard-wf

### DIFF
--- a/lib/react_dom.dart
+++ b/lib/react_dom.dart
@@ -30,27 +30,31 @@ import 'dart:html';
 import 'package:over_react/over_react.dart';
 import 'package:react/react_dom.dart' as react_dom show render, unmountComponentAtNode;
 
-/// Renders the provided [vDomComponent] into the DOM in the provided [mountNode]
+/// Renders the provided [element] into the DOM mounted in the provided [mountNode]
 /// and returns a reference to it based on its type:
 ///
-/// 1. Returns an [Element] if [vDomComponent] is a [Dom] component _(e.g. [Dom.div])_.
-/// 2. Returns a React `Component` if [vDomComponent] is a composite component.
-/// 3. Returns `null` if [vDomComponent] is null, or if [vDomComponent] is a stateless component.
+/// 1. Returns an [Element] if [element] is a DOM component _(e.g. [Dom.div])_.
+/// 2. Returns a React `Component` if [element] is a composite component.
 ///
-/// > If the [vDomComponent] was previously rendered into the [mountNode], this will perform an update on it and only
-///   mutate the DOM as necessary to reflect the latest React component.
+/// > __Throws__ if [element] or [mountNode] are `null`.
 ///
-/// > Use [unmountComponentAtNode] to unmount the instance.
+/// If the [element] was previously rendered into the [mountNode], this will perform an update on it and only
+/// mutate the DOM as necessary to reflect the latest React component.
+///
+/// Use [unmountComponentAtNode] to unmount the instance.
 ///
 /// > Proxies [react_dom.render].
-dynamic render(ReactElement vDomComponent, Element mountNode) {
-  return react_dom.render(vDomComponent, mountNode);
+dynamic render(ReactElement element, Element mountNode) {
+  return react_dom.render(element, mountNode);
 }
 
-/// Removes a React `Component` that was mounted via [render] from the DOM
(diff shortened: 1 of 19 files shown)


aviary-wf:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://dev.workiva.net/docs/information-security/aviary#configuring-aviary) coverage of security relevant code
---

lib/react_dom.dart
```diff
@@ -0,0 +1,58 @@
...
+
+import 'dart:html';
+
+import 'package:over_react/over_react.dart';
+import 'package:react/react_dom.dart' as react_dom show render, unmountComponentAtNode;
+
+/// Renders the provided [vDomComponent] into the DOM in the provided [mountNode]
+/// and returns a reference to it based on its type:
+///
+/// 1. Returns an [Element] if [vDomComponent] is a [Dom] component _(e.g. [Dom.div])_.
```
greglittlefield-wf: #nit I think we can just say `DOM component`, since it doesn't necessarily have to come from `[Dom]`.

lib/react_dom.dart
```diff
@@ -0,0 +1,58 @@
...
+/// 2. Returns a React `Component` if [vDomComponent] is a composite component.
+/// 3. Returns `null` if [vDomComponent] is null, or if [vDomComponent] is a stateless component.
+///
+/// > If the [vDomComponent] was previously rendered into the [mountNode], this will perform an update on it and only
+///   mutate the DOM as necessary to reflect the latest React component.
+///
+/// > Use [unmountComponentAtNode] to unmount the instance.
+///
+/// > Proxies [react_dom.render].
+dynamic render(ReactElement vDomComponent, Element mountNode) {
```
greglittlefield-wf: I think it `vDomComponent` should be renamed to `element` to be consistent with our other naming conventions (or `nextElement`, which is what they call the variable in the React JS code).

lib/react_dom.dart
```diff
@@ -0,0 +1,58 @@
...
+  if (vDomComponent == null) return null;
```
greglittlefield-wf: We don't need this check, since React handles this case. Plus, this would prevent the behavior of effectively unmounting a component by rendering `null`. Could you add a test for that case?

aaronlademann-wf: @greglittlefield-wf it appears that the doc comment is not accurate.

If passed null, [`ReactDom.render`]() throws an `InvariantViolation`:

```
'Invariant Violation: ReactDOM.render(): Invalid component element.'
```

I've added a test to verify that... but as far as I can tell - we don't have a way to create stateless components (or at least not one in which the return value would be `ReactElement`), so I've removed that bit from the doc comment.

aaronlademann-wf:
@greglittlefield-wf all feedback addressed.

lib/react_dom.dart
greglittlefield-wf: We don't have a way to create those, but that could change in the future, and you could also potentially get one from a JS library, so I think we should have retained that information. Not a big deal, though.

sebastianmalysa-wf:
QA +1

- [x] Testing instruction
- [x] Dev +1's
- [x] Dev/QA +10 with detail of what was tested
    - CI passes 
    - pulled down the branch locally - all tests pass
- [x] All unit tests pass

PR #151, opened 2018-05-03
AF-223 Begin preparations for Dart SDK 2

aaronlademann-wf:
## Ultimate problem:
Dart SDK 2 is on its way, and over_react has a number of lints / hints / warnings that are easy enough to address ahead of time.

## How it was fixed:
* I addressed all of the lints / hints / warnings / deprecations that show up using the latest Dart 2 SDK build.
    > Note: The stuff involving `react` callable classes is not addressed here.  That work will be done in a separate PR (which requires a change / release of `react`).

## Testing suggestions:
Passing CI

## Potential areas of regression:
None expected

---

> __FYA:__ @Workiva/app-frameworks

PR #155, opened 2018-05-08
[2.0.0-dev - dart2] AF-233 Consume react-dart callable class fixes

aaronlademann-wf:
> Depends on https://github.com/cleandart/react-dart/pull/142

### Problem(s):
1. In the Dart 2 SDK, the assignment of an object with a call method to a function type will now be treated as an implicit closurization of the `.call` method of the object.  This is not compatible with how `ReactComponentFactory` works.
2. The `MapBase` additions from #151are no longer the best way to satisfy the missing concrete `Map` class impls.
3. Misc. housekeeping items need to be addressed to get the `2.0.0-dev` branch ready to support further Dart 2 SDK testing / `2.0.0-dev.<n>` releases as the Dart team moves closer and closer to a production release of the Dart 2 SDK.

### Solution(s):

1. Consume the https://github.com/cleandart/react-dart/pull/142 callable class fixes
2. The `MapBase` additions from #151 were reverted and the new fields / getters / methods from that are present on the `Map` class in the Dart 2 SDK were instead implemented individually so we don't add a bunch of stuff from `dart:collection` that we don't actually need.
3. Address housekeeping items
    1. Update Dart SDK dependency range to `>=2.0.0-dev.0 <=2.0.0-dev.49.0`
    2. Address any lints / hints / errors found when the analysis server in Dart `w.0.0-dev.49.0` runs

### Testing Instructions:

1. Switch your Dart SDK to `2.0.0-dev.49.0`
3. Serve the examples using the DDC:
    ```bash
    pub serve --web-compiler=dartdevc
    ```
4. Load `http://localhost:8080` and play around with the example components.
5. Verify that the examples work as expected without producing run time errors

### Known Issues:

There are some issues present that, upon some investigation, appear to be unrelated to these changes:

1. There are numerous test failures centered around type casting.  I am unsure if this is a `test` package issue, or if we need to do away with some of our "shared" test suites.

---

__FYA:__ @greglittlefield-wf @clairesarsam-wf 

__FYI:__ @dustinlessard-wf @sebastianmalysa-wf @robbecker-wf


aviary3-wk:
## Security Insights

## Action Items
* Verify [`aviary.yaml`](https://dev.workiva.net/docs/information-security/aviary#configuring-aviary) coverage of security relevant code
---

aaronlademann-wf:
@greglittlefield-wf I don't think we'll be able to get a passing travis build until https://github.com/cleandart/react-dart/pull/142/files is merged / released as `5.0.0-dev.alpha.0`.  Travis doesn't seem to like git dependencies

aaronlademann-wf:
Also, once https://github.com/Workiva/smithy-runner-dart/pull/48 goes through we should be able to get the smithy build passing as well.

lib/src/component_declaration/built_redux_component.dart
```diff
@@ -96,7 +96,7 @@ abstract class BuiltReduxUiComponent<
 
   @mustCallSuper
   @override
-  void redraw([callback()]) {
+  void redraw([Function callback]) {
```
greglittlefield-wf: This should be changed back (is this leftover from when it was typed to `Function` previously in react-dart?)

lib/src/component_declaration/component_base.dart
```diff
@@ -700,17 +662,51 @@ abstract class _OverReactMapViewBase<K, V> {
   Map<K, V> get _map;
 }
 
+/// A class that mimics `MapBase` without all the additional functionality we don't need.
+///
+/// > Necessary for eventual upgrade to the Dart 2 SDK.
+abstract class _OverReactMapBase<K, V> extends Object implements MapViewMixin<K, V>, Map<K, V> {
+  // Manually implement members from `Map` that appear in the Dart 2 SDK
+  // without having to extend from `MapBase`.
```
greglittlefield-wf: #nit I think this comment implies the wrong thing; we don't want to extend from `MapBase` because we want to just proxy that map.

Perhaps this class should be called `_MapProxy` or `_OverReactMapView`

lib/src/component_declaration/component_base.dart
```diff
@@ -700,17 +662,51 @@ abstract class _OverReactMapViewBase<K, V> {
...
+  @override void addEntries(Iterable<MapEntry<K, V>> newEntries) => _map.addEntries(newEntries);
+  @override void removeWhere(bool predicate(K key, V value)) => _map.removeWhere(predicate);
+  @override V update(K key, V update(V value), {V ifAbsent()}) => _map.update(key, update, ifAbsent: ifAbsent);
+  @override void updateAll(V update(K key, V value)) => _map.updateAll(update);
+  @override Map<RK, RV> cast<RK, RV>() => _map.cast<RK, RV>();
+  @override Map<RK, RV> retype<RK, RV>() => _map.retype<RK, RV>();
+
+  // Manually implement members from `MapViewMixin`,
+  // since mixing that class in doesn't play well with the DDC.
+  // TODO find out root cause and reduced test case.
```
greglittlefield-wf: I think this block of comments is now outdated and can be removed, right?

lib/src/component_declaration/component_base.dart
```diff
@@ -503,9 +483,9 @@ typedef PropsModifier(Map props);
 ///
 /// > Note: Implements [MapViewMixin] instead of extending it so that the abstract [Props] declarations
 /// don't need a constructor. The generated implementations can mix that functionality in.
-abstract class UiProps extends MapBase
+abstract class UiProps<K, V> extends _OverReactMapBase<K, V>
```
greglittlefield-wf: Is there a reason generic parameters were added to `UiProps`/`UiState` (and other places)? I'm concerned that that might introduce type complexity in props class inheritances (which can be problematic in dart2js; see https://github.com/dart-lang/sdk/issues/14729) without providing much benefit.

lib/src/component_declaration/component_base.dart
```diff
@@ -514,28 +494,10 @@ abstract class UiProps extends MapBase
...
-  @override bool containsValue(Object value) => _map.containsValue(value);
-  @override void forEach(void action(key, value)) { _map.forEach(action); }
-  @override bool get isEmpty => _map.isEmpty;
-  @override bool get isNotEmpty => _map.isNotEmpty;
-  @override int get length => _map.length;
-  @override Iterable get keys => _map.keys;
-  @override remove(Object key) => _map.remove(key);
-  @override Iterable get values => _map.values;
-
   // Manually implement members from `StateMapViewMixin`,
```
greglittlefield-wf: (I think `StateMapViewMixin` is a typo and should say `PropsMapViewMixin`.)

Could we try mixing in `PropsMapViewMixin instead and see how it works with Dart 2's DDC?

lib/src/util/react_wrappers.dart
```diff
@@ -281,7 +281,7 @@ react.Component getDartComponent(/* ReactElement|ReactComponent|Element */ insta
     return null;
   }
 
-  assert(() {
```
greglittlefield-wf: Was passing functions to `assert` removed in Dart 2 or something?

test/over_react/component/abstract_transition_test.dart
```diff
@@ -556,7 +556,7 @@ main() {
           // Have to use MutationObserver since setting the state to PRE_SHOWING
           // synchronously updates the state to SHOWING.
           var transitionAttrMutations = <List<String>>[];
-          var observer = new MutationObserver((records, observer) {
+          var observer = new MutationObserver((List<dynamic> records, observer) {
```
greglittlefield-wf: Could use a comment that this is `<dynamic>` to work around a Dart HTML internals issue in the DDC

test/over_react/component/abstract_transition_test.dart
```diff
@@ -745,7 +745,7 @@ class TransitionerComponent extends AbstractTransitionComponent<TransitionerProp
   List<TransitionPhase> transitionPhasesSet = [];
 
   @override
-  void setState(dynamic newState, [callback()]) {
+  void setState(dynamic newState, [Function callback]) {
```
greglittlefield-wf: I think this change can be reverted now since this typing change is no longer present in `react-dart`.

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -250,35 +250,38 @@ main() {
...
-
-        expect(calls, [
-          [expectedProps, 0, 0, 0, 0],
-          [expectedProps, 1, 0, 0, 0],
-          [expectedProps, 1, 2, 0, 0],
-          [expectedProps, 1, 2, 3, 0],
-          [expectedProps, 1, 2, 3, 4],
-        ]);
-      });
+      // TODO: The typing for `customFactory` is an analyzer error in Dart 2 (dev.49), is this DDC workaround + test still necessary or can it be removed?
```
greglittlefield-wf: This line of the comment can at least be removed, since the `NOTE` below is more up to date.

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -250,35 +250,38 @@ main() {
...
+      //
+      // NOTE: This is no longer supportable in Dart 2 (should advertise as a breaking change in release notes)
```
greglittlefield-wf: Actually, can we just remove this test? We don't intend on doing anything with it, and we know this is going to break.

test/over_react/component_declaration/flux_component_test/default.dart
```diff
@@ -14,7 +14,7 @@ class TestDefaultComponent extends FluxUiComponent {
   render() => Dom.div()();
 
   @override
-  void setState(_, [callback()]) {
+  void setState(_, [Function callback]) {
```
greglittlefield-wf: This and its friends should be reverted as well.

test/over_react/shared/map_proxy_tests.dart
```diff
@@ -50,12 +50,12 @@ void mapProxyTests(Map mapProxyFactory(Map proxiedMap)) {
       var testMap = {};
       when(mock.addAll(any)).thenReturn('value');
 
-      expect(proxy.addAll(testMap), isNull);
+      expect(() => proxy.addAll(testMap), isNull);
```
greglittlefield-wf: Huh, you can pass in a function and `expect` will evaluate it and match the result? Is this documented somewhere?

test/over_react/shared/map_proxy_tests.dart
```diff
@@ -83,7 +83,7 @@ void mapProxyTests(Map mapProxyFactory(Map proxiedMap)) {
     test('forEach', () {
       var callback = (key, value) {};
 
-      expect(proxy.forEach(callback), isNull);
+      expect(() => proxy.forEach(callback), isNull);
       verify(mock.forEach(callback));
     });
```
greglittlefield-wf: We should add tests for the new Dart 2 Map members as well

test/over_react/util/handler_chain_util_test.dart
```diff
@@ -17,12 +17,13 @@ library handler_chain_util_test;
 import 'package:over_react/over_react.dart';
 import 'package:test/test.dart';
 
+typedef dynamic ChainFunction([_, __, ___, ____, _____, ______]);
```
greglittlefield-wf: I think this ended up being unused.

test/over_react/shared/map_proxy_tests.dart
aaronlademann-wf: That I don't know for sure @greglittlefield-wf ... I did this to address a dart 2 analyzer error, and the test didn't fail.  I didn't dig too much further into it.

lib/src/util/react_wrappers.dart
aaronlademann-wf: @greglittlefield-wf the assert arg has to return a bool in dart 2.

test/over_react/shared/map_proxy_tests.dart
aaronlademann-wf: @greglittlefield-wf agreed, but currently, these tests are skipped in the DDC because they make use of mirrors.

test/over_react/shared/map_proxy_tests.dart
aaronlademann-wf: In addition, these test are currently not run in the DDC because they us mirrors.

greglittlefield-wf: It looks like it doesn't work. This is causing failures in Dart 1:
```
00:16 +245 -20: [Chrome] test/over_react_component_declaration_test.dart: component base: UiState provides Map functionality: proxies the Map member: forEach [E]
  Expected: null
    Actual: <Closure: () => void from: () => proxy[$forEach](callback)>
  
  package:test                                 expect
  over_react/shared/map_proxy_tests.dart 86:7  test.test.dart.fn
```

We'll have to find a different workaround for this issue; perhaps we just shouldn't assert the return value. Or just comment out the tests until we're running them in Dart 2.

test/over_react/shared/map_proxy_tests.dart
greglittlefield-wf: Yeah, at some point, we need to update these tests to use DDC-compatible mocks, or to not use mocks at all. 

We do need some kind of test coverage around these methods; we should at least stub out these tests with FIXMEs or something so that we add them by the time we release these changes as stable.

lib/src/component_declaration/component_base.dart
```diff
@@ -514,25 +494,7 @@ abstract class UiProps extends MapBase
     }
   }
 
-  // Manually implement members from `MapViewMixin`,
```
greglittlefield-wf: This branch looks so much better than what we currently have! 

There are a few more things that could be cleaned up, though, and since they're hard to explain I've just made a PR into this branch: https://github.com/Workiva/over_react/pull/158.

test/over_react/shared/map_proxy_tests.dart
aaronlademann-wf: @greglittlefield-wf this branch is targeting an SDK version of 2.0.0-dev.49.0, I wouldn't expect a lot of tests in this branch to pass using the dart 1 SDK.

aaronlademann-wf:
@greglittlefield-wf I've merged in your changes, and your comments about the `() => proxy` thing not working in Dart 1 - I'm not sure they're really relevant since this is merging into the 2.0.0 branch which is a dart 2 SDK "working" branch.

I have tests passing locally - although there is likely still some work to be done getting them passing in Travis.  I think this is a good first step though, and I'd like to get this merged unless there are major objections by you or @clairesarsam-wf.

greglittlefield-wf:
@aaronlademann-wf I'd rather us just comment out the tests than check in invalid changes just to fix syntax errors.

> We do need some kind of test coverage around these methods; we should at least stub out these tests with FIXMEs or something so that we add them by the time we release these changes as stable.

I also think we should stub those in so we don't forget about it.

clairesarsam-wf:
+10

- [x] Examples page renders as expected with no runtime errors

PR #196, opened 2018-10-26
AF-2922: add defaultProps getter to UiProp

sebastianmalysa-wf:
## Ultimate problem:

With the new builder, the implementation of ```typedPropsFactory()``` will no longer be available on the component class defined by the consumer – it will only be available on the generated component class. As a consequence, all instances of ```new FooComponent().getDefaultProps()``` will break.

## How it was fixed:

Add a ```defaultProps``` getter on the ```UiProps``` as an alternative way to retrieve a component's default props since instantiating the component class directly is an anti-pattern anyway.

## Testing suggestions:

- [ ] Code changes and test updates make sense
- [ ] CI Passes 

---

@greglittlefield-wf @corwinsheahan-wf  @evanweible-wf

### DIFF
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -691,6 +691,7 @@ abstract class UiProps extends MapBase
 
   Function get componentFactory;
 
+  /// The default props for this component brought in from the [componentFactory].
   Map get componentDefaultProps => componentFactory is ReactDartComponentFactoryProxy
       // ignore: avoid_as
       ? (componentFactory as ReactDartComponentFactoryProxy).defaultProps
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -691,7 +691,7 @@ abstract class UiProps extends MapBase
 
   Function get componentFactory;
 
-  Map get defaultProps => componentFactory is ReactDartComponentFactoryProxy
+  Map get componentDefaultProps => componentFactory is ReactDartComponentFactoryProxy
       // ignore: avoid_as
       ? (componentFactory as ReactDartComponentFactoryProxy).defaultProps
       : const {};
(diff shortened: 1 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

greglittlefield-wf:
+1

Might be good to do a quick check to make sure no one's declaring a `defaultProps` prop before releasing.

sebastianmalysa-wf:
Looks like ```defaultProps``` is initialized on prop mixins in several places, for example: https://github.com/Workiva/truss/blob/7100e7346d59b4208d9b4a03d92d178657c9fa53/lib/src/notification_manager/internal/components/shared_notification_manager_props.dart#L8

lib/src/component_declaration/component_base.dart
```diff
@@ -690,6 +690,11 @@ abstract class UiProps extends MapBase
   }
 
   Function get componentFactory;
+
+  Map get componentDefaultProps => componentFactory is ReactDartComponentFactoryProxy
```
greglittlefield-wf: This should have a doc comment

lib/src/component_declaration/component_base.dart
```diff
@@ -690,6 +690,11 @@ abstract class UiProps extends MapBase
...
+      // ignore: avoid_as
+      ? (componentFactory as ReactDartComponentFactoryProxy).defaultProps
+      : const {};
```
greglittlefield-wf: Can we add a test case for this case? 
e.g.,
```dart
expect(Dom.div().componentDefaultProps, equals({}));
```

corwinsheahan-wf:
QA +1
* CI Passes
* Added tests offer coverage of changes
* Tested with codemod updates in WSD. Works correctly in tandem with codemod changes implemented in WSD.

Merging into master.
@Workiva/release-management-p

PR #200, opened 2018-11-13
AF-3082: transform static meta fields

sebastianmalysa-wf:
Proposed solution to https://jira.atl.workiva.net/browse/AF-3082

## Ultimate problem:

In anticipation of the Dart 2 transition, all props and state classes will have a static meta getter that is implemented to point to some symbol that needs to be generated, for example:

```
class FooProps {
  static const PropsMeta meta = $metaForFooProps;
  ...
}
```

The transformer will need to transform this meta static field inline:

```
class Foo Props{
  static const PropsMeta meta = $Props(FooProps);
```

## How it was fixed:

- identify static members with the ```meta``` variable name and replace the initializer to use the appropriate ```Props()``` or ```State()```  implementation within the source file . 

## Testing suggestions:

Code changes and tests make sense 
CI passes 

---

@corwinsheahan-wf @evanweible-wf @greglittlefield-wf

### DIFF
--- a/lib/src/transformer/impl_generation.dart
+++ b/lib/src/transformer/impl_generation.dart
@@ -414,16 +414,6 @@ class ImplGenerator {
         .forEach((_field) {
           final field = _field as FieldDeclaration; // ignore: avoid_as
 
-          annotations.Accessor accessorMeta = instantiateAnnotation(field, annotations.Accessor);
-
-          if (accessorMeta?.doNotGenerate == true) {
-            logger.fine('Skipping generation of field `$field`.',
-                span: getSpan(sourceFile, field)
-            );
-
-            return;
-          }
-
           final name = typedMap.node.name.name;
           final metaType = type == AccessorType.props ? 'Props' : 'State';
 
--- a/lib/src/transformer/impl_generation.dart
+++ b/lib/src/transformer/impl_generation.dart
@@ -410,6 +410,31 @@ class ImplGenerator {
     Map constants = {};
 
     typedMap.node.members
+        .where((member) => member is FieldDeclaration && member.isStatic)
+        .forEach((_field) {
+          final field = _field as FieldDeclaration; // ignore: avoid_as
+
+          annotations.Accessor accessorMeta = instantiateAnnotation(field, annotations.Accessor);
+
+          if (accessorMeta?.doNotGenerate == true) {
+            logger.fine('Skipping generation of field `$field`.',
+                span: getSpan(sourceFile, field)
+            );
+
+            return;
+          }
+
+          final name = typedMap.node.name.name;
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -256,6 +256,55 @@ main() {
               reason: 'should preserve existing inheritance');
         });
 
+        test('with static PropsMeta declaration', () {
```
corwinsheahan-wf: we should also have a test which verifies this functionality on props/state mixins:

```
test('with static PropsMeta declaration in PropsMixin', () {
  final transformedLine = 'static const PropsMeta meta = \$Props(FooPropsMixin);';
  setUpAndGenerate('''
    @PropsMixin()
    class FooPropsMixin {
      static const PropsMeta meta = \$metaForFooPropsMixin;
    }
    '''
  );
  var transformedSource = transformedFile.getTransformedText();
  expect(transformedSource, contains(transformedLine));
});

File: test/vm_tests/transformer/impl_generation_test.dart [hunk 2]
```diff
@@ -256,6 +256,55 @@ main() {
...
+           
+           @Component()
+           class FooComponent {
+            render() => null;
+           }
+           '''
+          );
+
+          var transformedSource = transformedFile.getTransformedText();
+          expect(transformedSource, contains(transformedLine));
```
corwinsheahan-wf: I think it might make sense to also verify that the original meta line is not longer present to ensure that the transformer is replacing and not simply adding a line.

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -256,6 +256,55 @@ main() {
...
+        });
+
+        test('with static StateMeta declaration', () {
```
corwinsheahan-wf: I think we can combine the StateMeta and PropsMeta tests into 1 test, since any instance of a component will have at least a props class with a `meta` field.

lib/src/transformer/impl_generation.dart
```diff
@@ -409,6 +409,31 @@ class ImplGenerator {
     Map keyConstants = {};
     Map constants = {};
 
+    typedMap.node.members
+        .where((member) => member is FieldDeclaration && member.isStatic)
+        .forEach((_field) {
+          final field = _field as FieldDeclaration; // ignore: avoid_as
+
+          annotations.Accessor accessorMeta = instantiateAnnotation(field, annotations.Accessor);
```
corwinsheahan-wf: Can we skip lines 417-425? I don't think there should ever be an instance where a `doNotGenerate()` annotation is added to a `meta` field

PR #201, opened 2018-11-15
AF-3083: Transformer should emit an empty part file for libraries containing the .over_react.g.dart part directive

sebastianmalysa-wf:
Proposed solution for https://jira.atl.workiva.net/browse/AF-3083

## Ultimate problem:

The builder requires a part file, but the transformer does not. The updated boilerplate rollout will include a part directive that includes this new generated part, but that file will not exist when running in Dart 1.

To handle this, the transformer should output an empty file for each of these parts so that it does not fail at runtime.

## How it was fixed:

- find and remove the ignore comment and the associated part directive required by the builder 

## Testing suggestions:

Code changes and tests make sense 
CI passes 

@corwinsheahan-wf @evanweible-wf @greglittlefield-wf


aviary2-wf:
## Security Insights

## Action Items
---

lib/transformer.dart
```diff
@@ -98,6 +98,26 @@ class WebSkinDartTransformer extends Transformer implements LazyTransformer {
     TransformedSourceFile transformedFile = new TransformedSourceFile(sourceFile);
     TransformLogger logger = new JetBrainsFriendlyLogger(transform.logger);
 
+    var ignoreCommentPattern = new RegExp(r'\/\/\s?ignore:\s?uri_does_not_exist,\s?uri_has_not_been_generated');
+    var partPattern = new RegExp(r'(\s*?)part\s(.*?).overReact.g.dart(.;)');
```
greglittlefield-wf: Not sure if this was already considered, but another option would be to emit an empty `.g.dart` file from the transformer, which is probably more robust than trying to do this transformation.

Just putting it out there.

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -864,6 +864,80 @@ main() {
...
+
+        var transformedSource = transformedFile.getTransformedText();
+
+        expect(transformedSource, contains(ignoreLine));
+      });
+
+
+      test('ignore line and part directive is removed from component file', () {
+        final ignoreLine = '\/\/ ignore: uri_does_not_exist, uri_has_not_been_generated';
+        final partLine = "part \'component\.overReact\.g\.dart\';";
```
evanweible-wf: Needs one more test to verify that the part directive is removed even if the ignore comment isn't found

lib/transformer.dart
```diff
@@ -98,6 +98,25 @@ class WebSkinDartTransformer extends Transformer implements LazyTransformer {
...
```
evanweible-wf: I think this regex should actually enforce that the `part` directive starts at the very beginning of the line so that we avoid a commented out part directive. My suggestion would be:

```dart 
var partPattern = new RegExp(r'''^part\s+['"].+.overReact.g.dart['"];''');
```

sebastianmalysa-wf:
Working on adding tests.

corwinsheahan-wf:
@sebastianmalysa-wf to keep with dart naming conventions, we're gonna have the output extension be `over_react.g.dart` instead of `overReact.g.dart`. We'll have to update this PR to look for the former instead of the latter.

test/vm_tests/transformer/test_data/componentWithPartDirective.dart
```diff
@@ -0,0 +1,17 @@
+import 'package:over_react/over_react.dart';
+
+// ignore: uri_does_not_exist, uri_has_not_been_generated
+part 'foo.over_react.g.dart';
+// ignore: uri_does_not_exist, uri_has_not_been_generated
+part 'bar.over_react.g.dart';
```
corwinsheahan-wf: We'll only ever have 1 `over_react.g.dart` part directive in a file, so we should remove lines 5-6

lib/transformer.dart
```diff
@@ -98,6 +98,19 @@ class WebSkinDartTransformer extends Transformer implements LazyTransformer {
...
 
+    var partPattern = new RegExp(r'''part\s+['"](.+.over_react.g.dart)['"];''');
+    var partFilename = partPattern.firstMatch(sourceFile.getText(0));
+
+    // For Dart 1 compatibility an empty generated part file will be created when a file contains
+    // the part directive pointing to the generated file the new builder requires.
+    if (partFilename != null) {
+        var sourceFileDirectory = p.dirname(sourceFile.url.toFilePath());
+        var partFilePath = p.join(sourceFileDirectory, partFilename.group(1));
+        var asset = new Asset.fromString(new AssetId(transform.primaryInput.id.package, partFilePath), '');
```
greglittlefield-wf: Shouldn't this contain a `part of` directive?

Or, does it not matter since dart2js doesn't care? If so, fine by me 🤷‍♂️

test/vm_tests/transformer/transformer_test.dart
```diff
@@ -15,17 +15,56 @@
...
+    WebSkinDartTransformer initWithConfig(Map config) =>
+        new WebSkinDartTransformer.asPlugin(new BarbackSettings(config, BarbackMode.DEBUG));
+
+    test('declares the correct input assets', () {
+      expect(initWithConfig({}).isPrimary(new AssetId(fakePackage, 'web/src/demo_components/button.dart')), isTrue);
+      expect(initWithConfig({}).isPrimary(new AssetId(fakePackage, 'web/src/index.html')), isFalse);
+      expect(initWithConfig({}).isPrimary(new AssetId(fakePackage, 'dart_test.yaml')), isFalse);
+    });
+
+    test('outputs an empty file when dart 2 boiler plate compatible part directive is found', () async {
```
greglittlefield-wf: Should we have a test for when an empty file should not be emitted?

lib/transformer.dart
evanweible-wf: Yeah it should, good catch.

```dart
var contents = "part of '${p.basename(sourceFile.url.toFilePath())}';";
var asset = new Asset.fromString(new AssetId(transform.primaryInput.id.package, partFilePath), contents);
```

@sebastianmalysa-wf seb we'll want to do this ^

test/vm_tests/transformer/transformer_test.dart
```diff
@@ -15,17 +15,80 @@
...
+      List<Asset> fileAssets = verify(mockTransform.addOutput(captureThat(isDartFile))).captured;
+
+      expect(fileAssets.length, equals(2));
+      expect(fileAssets[0].id.toString(), equals('testId|foo.over_react.g.dart'));
+      expect(fileAssets[1].id.toString(), equals('testId|component_with_part_directive.dart'));
+      expect(await fileAssets[0].readAsString(), equals(generatedPartFileContents));
+    });
+
+    test('does not output a generated part file when dart 2 boiler plate compatible part directive is not found', () async {
+      AssetId fakeInputFileAssetId = new AssetId('testId', 'component_with_part_directive.dart');
```
evanweible-wf:
```suggestion
      AssetId fakeInputFileAssetId = new AssetId('testId', 'component_without_part_directive.dart');
```

test/vm_tests/transformer/transformer_test.dart
```diff
@@ -15,17 +15,80 @@
...
+      when(mockTransform.primaryInput).thenReturn(inputFile);
+      when(inputFile.readAsString())
+          .thenReturn(new File.fromUri(Uri.parse('test/vm_tests/transformer/test_data/component_without_part_directive.dart')).readAsString());
+
+      await initWithConfig({}).apply(mockTransform);
+
+      List<Asset> fileAssets = verify(mockTransform.addOutput(captureThat(isDartFile))).captured;
+
+      expect(fileAssets.length, equals(1));
+      expect(fileAssets[0].id.toString(), equals('testId|component_with_part_directive.dart'));
```
evanweible-wf:
```suggestion
      expect(fileAssets[0].id.toString(), equals('testId|component_without_part_directive.dart'));
```

test/vm_tests/transformer/transformer_test.dart
```diff
@@ -15,17 +15,80 @@
...
+      expect(fileAssets[0].id.toString(), isNot(equals('testId|foo.over_react.g.dart')));
```
corwinsheahan-wf: This test seems redundant with line 87. We should verify that the length of fileAssets has not grown.

evanweible-wf:
QA +1
- [x] CI passes
- [x] All tests passed in WSD branch with part directives added to every file with over_react code

    ```diff
    +dependency_overrides:
    +  over_react:
    +    git:
    +      url: git@github.com:sebastianmalysa-wf/over_react.git
    +      ref: AF-3083
    ```

    ```
    23:44 +27294: All tests passed!
    ```

@Workiva/release-management-p

PR #203, opened 2018-11-23
AF-3084: Transformer should handle new boilerplate for Props and State classes

sebastianmalysa-wf:
## Ultimate problem:

With the new boilerplate, props and state classes (including abstract variants) will have a dual-class setup:

 - a private class that defines the interface via uninitialized fields
 - a public class that is essentially a stub and needs to be transformed to include the concrete accessor implementations

## How it was fixed:

- When parsing declarations if a class annotated with ```@Props()``` or ```@State()``` is private look for the public ```Props|State``` class and add that to the ```declarationMap``` instead. 
- Remove the ```with _$<Name>(Props|State)AcessorsMixin``` from the public signature

## Testing suggestions:

- CI passes
- Code changes and tests make sense 

> __FYA:__ @corwinsheahan-wf @evanweible-wf @greglittlefield-wf

### DIFF
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -53,21 +53,21 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
-    var privatePropsAndStateClassPattern = new RegExp(r'^(?:\@Props\(\)\s*|\@State\(\)\s*)(?:abstract\s)?class\s(_\$\w*?)\s');
-
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
-        var match = privatePropsAndStateClassPattern.firstMatch(member.toString());
+        var publicMember;
 
-        if ((name == 'Props' || name == 'State') && match != null) {
-          var publicName = match.group(1).replaceFirst('_\$', '');
-          var publicMember = unit.declarations.firstWhere((CompilationUnitMember member) => member is ClassDeclaration && member.name.name == publicName, orElse: () => null);
-
-          declarationMap[name]?.add(publicMember ?? member);
-        } else {
-          declarationMap[name]?.add(member);
+        if (member is ClassDeclaration && member.name.name.startsWith('_\$')) {
+          unit.declarations.forEach((CompilationUnitMember innerLoopMember) {
+            // Look for a matching public class.
+            if (innerLoopMember is ClassDeclaration && member.name.name.substring(2) == innerLoopMember.name.name) {
+              publicMember = innerLoopMember;
+            }
+          });
         }
+
+        declarationMap[name]?.add(publicMember ?? member);
       });
     });
 
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -53,12 +53,12 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
-    var propsClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
+    var propsAndStateClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
 
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
-        var match = propsClassPattern.firstMatch(member.toString());
+        var match = propsAndStateClassPattern.firstMatch(member.toString());
 
         if ((name == 'Props' || name == 'State') && match != null) {
           var publicMember = unit.declarations.where((CompilationUnitMember member) => member.toString().contains(match.group(1).substring(2)));
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -53,7 +53,7 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
-    var privatePropsAndStateClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
+    var privatePropsAndStateClassPattern = new RegExp(r'^(?:\@Props\(\)\s*|\@State\(\)\s*)(?:abstract\s)?class\s(_\$\w*?)\s');
 
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
@@ -61,9 +61,10 @@ class ParsedDeclarations {
         var match = privatePropsAndStateClassPattern.firstMatch(member.toString());
 
         if ((name == 'Props' || name == 'State') && match != null) {
-          var publicMember = unit.declarations.where((CompilationUnitMember member) => member.toString().contains(match.group(1).substring(2)));
+          var publicName = match.group(1).replaceFirst('_\$', '');
+          var publicMember = unit.declarations.firstWhere((CompilationUnitMember member) => member is ClassDeclaration && member.name.name == publicName, orElse: () => null);
 
-          declarationMap[name]?.add(publicMember.first);
+          declarationMap[name]?.add(publicMember ?? member);
         } else {
           declarationMap[name]?.add(member);
         }
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -53,12 +53,12 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
-    var propsAndStateClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
+    var privatePropsAndStateClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
 
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
-        var match = propsAndStateClassPattern.firstMatch(member.toString());
+        var match = privatePropsAndStateClassPattern.firstMatch(member.toString());
 
         if ((name == 'Props' || name == 'State') && match != null) {
           var publicMember = unit.declarations.where((CompilationUnitMember member) => member.toString().contains(match.group(1).substring(2)));
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -58,13 +58,19 @@ class ParsedDeclarations {
         var name = annotation.name.toString();
         var publicMember;
 
-        if (member is ClassDeclaration && member.name.name.startsWith('_\$')) {
-          unit.declarations.forEach((CompilationUnitMember innerLoopMember) {
-            // Look for a matching public class.
-            if (innerLoopMember is ClassDeclaration && member.name.name.substring(2) == innerLoopMember.name.name) {
-              publicMember = innerLoopMember;
+        if (name == 'Props' || name == 'State' || name == 'AbstractProps' || name == 'AbstractState') {
+          if (member is ClassDeclaration && member.name.name.startsWith('_\$')) {
+            var matchingPublicPropsOrStateClasses =
+              unit.declarations.where((CompilationUnitMember innerLoopMember) =>
+                innerLoopMember is ClassDeclaration && member.name.name.substring(2) == innerLoopMember.name.name);
+
+            if (matchingPublicPropsOrStateClasses.isEmpty) {
+              error('${member.name.name} must have an accompanying public class within the '
+                  'same file for Dart 2 builder compatibility, but one was not found.', getSpan(sourceFile, member));
+            } else {
+              publicMember = matchingPublicPropsOrStateClasses.first;
             }
-          });
+          }
         }
 
         declarationMap[name]?.add(publicMember ?? member);
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -53,11 +53,20 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
+    var propsClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
+
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
+        var match = propsClassPattern.firstMatch(member.toString());
 
-        declarationMap[name]?.add(member);
+        if ((name == 'Props' || name == 'State') && match != null) {
+          var publicMember = unit.declarations.where((CompilationUnitMember member) => member.toString().contains(match.group(1).substring(2)));
+
+          declarationMap[name]?.add(publicMember.first);
+        } else {
+          declarationMap[name]?.add(member);
+        }
       });
     });
 
--- a/lib/src/transformer/declaration_parsing.dart
+++ b/lib/src/transformer/declaration_parsing.dart
@@ -32,6 +32,7 @@ import 'package:transformer_utils/transformer_utils.dart';
 /// * Any number of mixins: `@PropsMixin()`, `@StateMixin()`
 class ParsedDeclarations {
   factory ParsedDeclarations(CompilationUnit unit, SourceFile sourceFile, TransformLogger logger) {
+    const companionPrefix = r'_$';
     bool hasErrors = false;
 
     void error(String message, [SourceSpan span]) {
@@ -60,9 +61,9 @@ class ParsedDeclarations {
         var name = annotation.name.toString();
 
         if (name == 'Props' || name == 'State' || name == 'AbstractProps' || name == 'AbstractState') {
-          if (member is ClassDeclaration && member.name.name.startsWith('_\$')) {
+          if (member is ClassDeclaration && member.name.name.startsWith(companionPrefix)) {
             final className = member.name.name;
-            final companionName = member.name.name.substring(2);
+            final companionName = member.name.name.substring(companionPrefix.length);
             final companionClass = unit.declarations.firstWhere(
               (innerLoopMember) => innerLoopMember is ClassDeclaration && innerLoopMember.name.name == companionName,
               orElse: () => null);
--- a/lib/src/transformer/impl_generation.dart
+++ b/lib/src/transformer/impl_generation.dart
@@ -92,7 +92,10 @@ class ImplGenerator {
         // The public props class signature includes a with <PropsClass>AccessorsMixin clause
         // for dart 2 builder compatibility. But in Dart 1, the transformer is able to generate
         // the concrete accessors inline without a separate mixin. For this reason, the transformer
-        // first removes this with clause and then generates the concrete accessors.
+        // removes this with clause and then generates the concrete accessors. To prevent the
+        // with clause being removed unnecessarily a check is preformed to identify if the class
+        // has an annotation since the public class added for dart 2 builder compatibility will not be
+        // annotated.
         if (declarations.props.node.metadata.isEmpty && declarations.props.node.withClause != null) {
           transformedFile.remove(getSpan(sourceFile, declarations.props.node.withClause));
         }
@@ -223,10 +226,13 @@ class ImplGenerator {
         final String stateName = declarations.state.node.name.toString();
         final String stateImplName = '$generatedPrefix${stateName}Impl';
 
-        // The public state class signature includes a with <PropsClass>AccessorsMixin clause
+        // The public state class signature includes a with <StateClass>AccessorsMixin clause
         // for dart 2 builder compatibility. But in Dart 1, the transformer is able to generate
         // the concrete accessors inline without a separate mixin. For this reason, the transformer
-        // first removes this with clause and then generates the concrete accessors.
+        // removes this with clause and then generates the concrete accessors. To prevent the
+        // with clause being removed unnecessarily a check is preformed to identify if the class
+        // has an annotation since the public class added for dart 2 builder compatibility will not be
+        // annotated.
         if (declarations.state.node.metadata.isEmpty && declarations.state.node.withClause != null) {
           transformedFile.remove(getSpan(sourceFile, declarations.state.node.withClause));
         }
@@ -373,10 +379,32 @@ class ImplGenerator {
     //   Abstract Props/State implementations
     // ----------------------------------------------------------------------
     declarations.abstractProps.forEach((abstractPropsClass) {
+      // The public abstract props class signature includes a with <AbstractPropsClass>AccessorsMixin clause
+      // for dart 2 builder compatibility. But in Dart 1, the transformer is able to generate
+      // the concrete accessors inline without a separate mixin. For this reason, the transformer
+      // removes this with clause and then generates the concrete accessors. To prevent the
+      // with clause being removed unnecessarily a check is preformed to identify if the class
+      // has an annotation since the public class added for dart 2 builder compatibility will not be
+      // annotated.
+      if (abstractPropsClass.node.metadata.isEmpty && abstractPropsClass.node.withClause != null) {
+        transformedFile.remove(getSpan(sourceFile, abstractPropsClass.node.withClause));
+      }
+
       generateAccessors(AccessorType.props, abstractPropsClass);
     });
 
     declarations.abstractState.forEach((abstractStateClass) {
+      // The public abstract state class signature includes a with <AbstractStateClass>AccessorsMixin clause
+      // for dart 2 builder compatibility. But in Dart 1, the transformer is able to generate
+      // the concrete accessors inline without a separate mixin. For this reason, the transformer
+      // removes this with clause and then generates the concrete accessors. To prevent the
+      // with clause being removed unnecessarily a check is preformed to identify if the class
+      // has an annotation since the public class added for dart 2 builder compatibility will not be
+      // annotated.
+      if (abstractStateClass.node.metadata.isEmpty && abstractStateClass.node.withClause != null) {
+        transformedFile.remove(getSpan(sourceFile, abstractStateClass.node.withClause));
+      }
+
       generateAccessors(AccessorType.state, abstractStateClass);
     });
   }
--- a/lib/src/transformer/impl_generation.dart
+++ b/lib/src/transformer/impl_generation.dart
@@ -736,10 +736,10 @@ class ImplGenerator {
 
 enum AccessorType {props, state}
 
-/// Check if the passed in class declaration is null, if not, remove it's with clause.
+/// Check if the passed in class declaration is null, if not, remove its with clause.
 ///
 /// The public Props|State|AbstractProps|AbstractState class signatures includes a with
-/// <PropsClass>AccessorsMixin clause for dart 2 builder compatibility. But in Dart 1,
+/// <Class>AccessorsMixin clause for dart 2 builder compatibility. But in Dart 1,
 /// the transformer is able to generate the concrete accessors inline without a separate
 /// mixin. For this reason, the transformer removes the with clause from the public class
 /// signatures.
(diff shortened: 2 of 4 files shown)


test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,81 @@ main() {
...
+
+        test('with builder compatible private state class', () {
+          final originalPrivateFooPropsLine = 'class _\$FooState extends UiState {}';
+          final originalPublicFooPropsLine = 'class FooState extends UiState with _\$FooStateAccessorsMixin implements _\$FooState';
+          final transformedFooPropsLine = 'class FooState extends UiState  implements _\$FooState';
+          final fooStateImplExtendsPublicClass = 'class _\$FooStateImpl extends FooState';
+          final fooStateImplExtendsPrivateClass = 'class _\$FooStateImpl extends _\$FooState';
+
+          preservedLineNumbersTest('''
+           @Factory()
```
aaronlademann-wf: #nit indentation is off here

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,81 @@ main() {
...
+          expect(transformedSource.contains(originalPublicFooPropsLine), isFalse);
+          expect(transformedSource, contains(transformedFooPropsLine));
+          expect(transformedSource, contains(fooPropsImplExtendsPublicClass));
+          expect(transformedSource.contains(fooPropsImplExtendsPrivateClass), isFalse);
+        });
+
+        test('with builder compatible private state class', () {
+          final originalPrivateFooPropsLine = 'class _\$FooState extends UiState {}';
+          final originalPublicFooPropsLine = 'class FooState extends UiState with _\$FooStateAccessorsMixin implements _\$FooState';
+          final transformedFooPropsLine = 'class FooState extends UiState  implements _\$FooState';
```
aaronlademann-wf: These three vars should be renamed, replacing `*Props` with `*State`

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,81 @@ main() {
...
+           @Component()
+           class FooComponent {
+            render() => null;
+           }
+           '''
+          );
+
+          var transformedSource = transformedFile.getTransformedText();
+          expect(transformedSource, contains(originalPrivateFooPropsLine));
+          expect(transformedSource.contains(originalPublicFooPropsLine), isFalse);
```
aaronlademann-wf: Could also do 

```dart
expect(transformedSource, isNot(contains(originalPublicFooPropsLine)));
```

for better consistency with the other expectations.

lib/src/transformer/impl_generation.dart
```diff
@@ -415,7 +424,7 @@ class ImplGenerator {
           final field = _field as FieldDeclaration; // ignore: avoid_as
 
           final name = typedMap.node.name.name;
-          final metaType = type == AccessorType.props ? 'Props' : 'State';
+          final metaType = isProps ? 'Props' : 'State';
```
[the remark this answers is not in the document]
sebastianmalysa-wf: ```isProps``` is set above here: https://github.com/Workiva/over_react/pull/203/files/5093aee0d5894136e9e96291fc7d02f2126371d9#diff-40cb89047341733783a79c631ef44984R411 I figured I'd just reuse it here.

lib/src/transformer/declaration_parsing.dart
```diff
@@ -53,11 +53,20 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
+    var privatePropsAndStateClassPattern = new RegExp(r'^.*class\s(_[$].*?\s)');
```
evanweible-wf: Instead of allowing anything before the `class` keyword, we should be more specific to avoid a commented-out class declaration from matching. Also changed up the class name capturing a bit:

```suggestion
    var privatePropsAndStateClassPattern = new RegExp(r'^(?:abstract )?class\s(_\$\w*?)\s');
```

lib/src/transformer/declaration_parsing.dart
```diff
@@ -53,11 +53,20 @@ class ParsedDeclarations {
...
+
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
+        var match = privatePropsAndStateClassPattern.firstMatch(member.toString());
+
+        if ((name == 'Props' || name == 'State') && match != null) {
+          var publicMember = unit.declarations.where((CompilationUnitMember member) => member.toString().contains(match.group(1).substring(2)));
```
evanweible-wf: I think this part could be broken up a bit to make the intention a bit more clear:

```dart
var publicName = match.group(1).replaceFirst('_$', '');
var publicMember = unit.declarations.firstWhere((CompilationUnitMember member) => member is ClassOrMixinDeclaration && member.name.name == publicName), orElse: () => null);
declarationMap[name]?.add(publicMember ?? member);
```

lib/src/transformer/impl_generation.dart
```diff
@@ -88,6 +88,15 @@ class ImplGenerator {
       // We can safely make this abstract, since we already have a runtime warning when it's
       // instantiated.
       if (!declarations.props.node.isAbstract) {
+
+        // Remove the with clause from the public props class signature for dart 2 builder compatibility.
+        // In Dart 2 there will be a dual class set up requiring a private and public props class. The private
+        // annotated class defines the interface via uninitialized fields. The public un-annotated class is a
+        // stub and needs to be transformed to include the concrete accessor implementations.
```
evanweible-wf: This comment should explicitly explain why we're removing the with clause. Maybe something like this:

> The public props class signature includes a `with <PropsClass>AccessorsMixin` clause for dart 2 builder compatibility. But in Dart 1, the transformer is able to generate the concrete accessors inline without a separate mixin. For this reason, the transformer first removes this with clause and then generates the concrete accessors.

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -140,6 +140,26 @@ main() {
           expect(declarations.declaresComponent, isTrue);
         });
 
+        test('a component with builder compatible annotated private props class', () {
+          setUpAndParse('''
+            @Factory()    UiFactory<FooProps> Foo;
+            class FooProps {}
```
evanweible-wf: This class should follow the same setup that we will have in practice where it extends the private class and mixes in an accessors mixin.

```dart
class FooProps extends _$FooProps with _$FooPropsAccessorsMixin {}
```

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -162,6 +182,29 @@ main() {
           expect(declarations.declaresComponent, isTrue);
         });
 
+        test('a stateful component with builder compatible annotated private state class', () {
+          setUpAndParse('''
+            @Factory()    UiFactory<FooProps> Foo;
+            @Props()      class FooProps {}
+            class FooState {}
```
evanweible-wf: Same comment here

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,80 @@ main() {
           expect(transformedSource, contains(transformedLine));
         });
 
+        test('with builder compatible private props class', () {
+          final originalPrivateFooPropsLine = 'class _\$FooProps extends UiProps {}';
+          final originalPublicFooPropsLine = 'class FooProps extends UiProps with _\$FooPropsAccessorsMixin implements _\$FooProps';
```
evanweible-wf: We are no longer extending `UiProps`/`UiState` and implementing the private class, instead we're just extending the private class:

```suggestion
          final originalPublicFooPropsLine = 'class FooProps extends _\$FooProps with _\$FooPropsAccessorsMixin {';
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,80 @@ main() {
...
+          final transformedFooPropsLine = 'class FooProps extends UiProps  implements _\$FooProps';
```
evanweible-wf:
```suggestion
          final transformedFooPropsLine = 'class FooProps extends _\$FooProps {';
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,80 @@ main() {
...
+          final fooPropsImplExtendsPublicClass = 'class _\$FooPropsImpl extends FooProps';
+          final fooPropsImplExtendsPrivateClass = 'class _\$FooPropsImpl extends _\$FooProps';
+
+          preservedLineNumbersTest('''
+            @Factory()
+            UiFactory<FooProps> Foo;
+        
+            class FooProps extends UiProps with _\$FooPropsAccessorsMixin implements _\$FooProps {
```
evanweible-wf:
```suggestion
            class FooProps extends _\$FooProps with _\$FooPropsAccessorsMixin {
```

test/vm_tests/transformer/impl_generation_test.dart
evanweible-wf: Same changes for the State test below

lib/src/transformer/declaration_parsing.dart
```diff
@@ -53,11 +53,21 @@ class ParsedDeclarations {
       key_stateMixin:        <CompilationUnitMember>[],
     };
 
+    var privatePropsAndStateClassPattern = new RegExp(r'^(?:\@Props\(\)\s*|\@State\(\)\s*)(?:abstract\s)?class\s(_\$\w*?)\s');
+
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
+        var match = privatePropsAndStateClassPattern.firstMatch(member.toString());
```
evanweible-wf: Talked about this offline, but we'll need to do a nested for-loop check here to find the public-pair instead of using a regex so that we don't have to account for all the possible variations (annotation params, extra annotations, comments, etc.)

lib/src/transformer/declaration_parsing.dart
```diff
@@ -53,11 +53,21 @@ class ParsedDeclarations {
...
+
+        if ((name == 'Props' || name == 'State') && match != null) {
```
evanweible-wf: Also check for `AbstractProps` and `AbstractState`

lib/src/transformer/declaration_parsing.dart
```diff
@@ -56,8 +56,18 @@ class ParsedDeclarations {
     unit.declarations.forEach((CompilationUnitMember member) {
       member.metadata.forEach((annotation) {
         var name = annotation.name.toString();
+        var publicMember;
 
-        declarationMap[name]?.add(member);
+        if (member is ClassDeclaration && member.name.name.startsWith('_\$')) {
+          unit.declarations.forEach((CompilationUnitMember innerLoopMember) {
```
corwinsheahan-wf: We'll probably want some validation to make sure that there is only 1 instance of a public props class in the file. We could do:
```suggestion
          var matchingPublicPropsOrStateClasses = 
          	unit.declarations.where((CompilationUnitMember innerLoopMember) => 
          	innerLoopMember is ClassDeclaration && 
				member.name.name.substring(2) == innerLoopMember.name.name);
```	

Then ensure that the length of `matchingPUblicPropsOrStateClasses` is 1 and assign it to `publicMember`. If the length is not 1, we should tell consumers that they should not write their own public props classes (or delete the one that is there), and instead rely on the code modded ones.

corwinsheahan-wf: Also, looks like this code runs on all annotations, we should only search for the public props/state class if the annotation is one of `@Props()`, `@State()`, `@AbstractProps()`, `@AbstractState()`

lib/src/transformer/declaration_parsing.dart
```diff
@@ -56,8 +56,18 @@ class ParsedDeclarations {
...
+            // Look for a matching public class.
+            if (innerLoopMember is ClassDeclaration && member.name.name.substring(2) == innerLoopMember.name.name) {
+              publicMember = innerLoopMember;
+            }
+          });
+        }
+
+        declarationMap[name]?.add(publicMember ?? member);
```
corwinsheahan-wf: Might be a good idea to create either a separate key in this map for the public props members, or add it to a different data structure entirely. This would allow the removing of the `with` clause to be more DRY and more robust, since we can just iterate over those class declarations directly instead of picking them out when iterating over props, abstract props, state, and abstract state in `impl_generation.dart`.

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -140,6 +140,45 @@ main() {
           expect(declarations.declaresComponent, isTrue);
         });
 
+        test('a component with a private \$ prefixed props class without a matching public class', () {
```
corwinsheahan-wf: We should have this test, but verify that the transformer throws an error when there is no accompanying public props class.

sebastianmalysa-wf:
+1 to Evan's commits

lib/src/transformer/impl_generation.dart
```diff
@@ -681,3 +735,17 @@ class ImplGenerator {
...
+// The public Props|State|AbstractProps|AbstractState class signatures includes a with
+// <PropsClass>AccessorsMixin clause for dart 2 builder compatibility. But in Dart 1,
+// the transformer is able to generate the concrete accessors inline without a separate
+// mixin. For this reason, the transformer removes this with clause and then generates
+// the concrete accessors. To prevent the with clause being removed unnecessarily a check
+// is preformed to identify if the class has an annotation since the public class added
+// for dart 2 builder compatibility will not be annotated.
+void removeWithClauseIfNecessary(ClassDeclaration declaration, SourceFile sourceFile, TransformedSourceFile transformedFile) {
+  if (declaration == null) return;
+  if (declaration.metadata.isEmpty && declaration.withClause != null) {
```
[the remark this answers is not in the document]
evanweible-wf: I missed this, we actually don't need the `metadata` check or the `withClause` anymore since we're capturing both the consumer-defined node as well as the companion node. In all of the usages of this method, we only ever pass the companion node so we can safely remove the with clause as long as the node itself isn't null

sebastianmalysa-wf:
@greglittlefield-wf and @evanweible-wf would it be better for us to add integration tests to the builder_compat_transformer_integration branch once all the transformer pieces are in place so we can test all of the boiler plate at that time? Or do you think it makes more sense to add integration tests as we go?

evanweible-wf:
@sebastianmalysa-wf I'd be okay with either option, I'll defer to @greglittlefield-wf. I was already planning on ensuring test coverage in the last step of the integration branch so that we cover existing boilerplate and new boilerplate

greglittlefield-wf:
Doesn't matter to me; however you guys want to schedule the work as long as it gets done before release.

lib/src/transformer/impl_generation.dart
```diff
@@ -736,16 +736,30 @@ class ImplGenerator {
...
+///     @Props()
+///     class _$FooProps extends UiProps {}
+///
+/// The builder is responsible for generating the _$FooPropsAccessorsMixin found in FooProps
+/// with clause, but since the transformer can inline concrete accessors _$FooPropsAccessorsMixin
+/// is not required and needs to be removed.
 void removeWithClauseIfNecessary(ClassDeclaration declaration, SourceFile sourceFile, TransformedSourceFile transformedFile) {
   if (declaration == null) return;
-  if (declaration.metadata.isEmpty && declaration.withClause != null) {
+  else {
```
evanweible-wf: Since we're returning in the above if statement, you don't need an `else` clause here.

lib/src/transformer/impl_generation.dart
```diff
@@ -736,16 +736,30 @@ class ImplGenerator {
...
 enum AccessorType {props, state}
 
-// The public Props|State|AbstractProps|AbstractState class signatures includes a with
-// <PropsClass>AccessorsMixin clause for dart 2 builder compatibility. But in Dart 1,
-// the transformer is able to generate the concrete accessors inline without a separate
-// mixin. For this reason, the transformer removes this with clause and then generates
-// the concrete accessors. To prevent the with clause being removed unnecessarily a check
-// is preformed to identify if the class has an annotation since the public class added
-// for dart 2 builder compatibility will not be annotated.
+/// Check if the passed in class declaration is null, if not, remove it's with clause.
```
evanweible-wf: s/it's/its

lib/src/transformer/impl_generation.dart
```diff
@@ -736,16 +736,30 @@ class ImplGenerator {
...
+///
+/// The public Props|State|AbstractProps|AbstractState class signatures includes a with
+/// <PropsClass>AccessorsMixin clause for dart 2 builder compatibility. But in Dart 1,
```
evanweible-wf: s/PropsClass/Class (since it could be props or state)

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+              render() => null;
+            }
+            
+            class AbstractStateProps extends _\$AbstractStateProps with _\$AbstractFooStateAccessorsMixin {
+              // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+              static const AbstractFooStateMeta meta = \$metaForAbstractStateProps;
+            }
+              
+            @AbstractState() 
+            class _\$AbstractStateProps {
```
evanweible-wf:
```suggestion
            abstract class _\$AbstractStateProps {
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+
+          preservedLineNumbersTest('''
+            @Factory()
+            UiFactory<FooProps> Foo;
+            
+            @Props()
+            class FooProps extends UiProps {}
+            
+            @Component()
+            class FooComponent {
```
evanweible-wf: Should be able to ditch the factory, props, and component for this test since we only care about the abstract state

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+              render() => null;
+            }
+          
+            class AbstractFooProps extends _\$AbstractFooProps with _\$AbstractFooPropsAccessorsMixin {
+              // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+              static const AbstractFooPropsMeta meta = \$metaForAbstractFooProps;
+            }
+            
+            @AbstractProps() 
+            class _\$AbstractFooProps {
```
evanweible-wf:
```suggestion
            abstract class _\$AbstractFooProps {
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+            
+            @Props()
+            class FooProps extends UiProps {}
+            
+            @Component()
+            class FooComponent {
+              render() => null;
+            }
+          
+            class AbstractFooProps extends _\$AbstractFooProps with _\$AbstractFooPropsAccessorsMixin {
```
evanweible-wf:
```suggestion
            abstract class AbstractFooProps extends _\$AbstractFooProps with _\$AbstractFooPropsAccessorsMixin {
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+            
+            @Props()
+            class FooProps extends UiProps {}
+            
+            @Component()
+            class FooComponent {
+              render() => null;
+            }
+            
+            class AbstractStateProps extends _\$AbstractStateProps with _\$AbstractFooStateAccessorsMixin {
```
evanweible-wf:
```suggestion
            abstract class AbstractFooState extends _\$AbstractFooState with _\$AbstractFooStateAccessorsMixin {
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+          final transformedFooStateLine = 'class AbstractFooProps extends _\$AbstractFooProps';
+
+          preservedLineNumbersTest('''
+            @Factory()
+            UiFactory<FooProps> Foo;
+            
+            @Props()
+            class FooProps extends UiProps {}
+            
+            @Component()
```
evanweible-wf: Should be able to ditch the factory, props, and component for this test since we only care about the abstract props

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+            class FooComponent {
+              render() => null;
+            }
+            
+            class AbstractStateProps extends _\$AbstractStateProps with _\$AbstractFooStateAccessorsMixin {
+              // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+              static const AbstractFooStateMeta meta = \$metaForAbstractStateProps;
```
evanweible-wf:
```suggestion
              static const StateMeta meta = \$metaForAbstractFooState;
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+          var transformedSource = transformedFile.getTransformedText();
+          expect(transformedSource, contains(originalPrivateClassLine));
+          expect(transformedSource, isNot(contains(originalPublicClassLine)));
+          expect(transformedSource, contains(transformedFooStateLine));
+        });
+
+        test('with builder-compatible dual-class abstract state setup', () {
+          final originalPrivateClassLine = 'class _\$AbstractStateProps {';
+          final originalPublicClassLine = 'class AbstractStateProps extends _\$AbstractStateProps with _\$AbstractStatePropsAccessorsMixin {';
+          final transformedFooStateLine = 'class AbstractStateProps extends _\$AbstractStateProps';
```
evanweible-wf:
```suggestion
          final transformedFooStateLine = 'abstract class AbstractFooState extends _\$AbstractFooState';
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+
+          var transformedSource = transformedFile.getTransformedText();
+          expect(transformedSource, contains(originalPrivateClassLine));
+          expect(transformedSource, isNot(contains(originalPublicClassLine)));
+          expect(transformedSource, contains(transformedFooStateLine));
+        });
+
+        test('with builder-compatible dual-class abstract state setup', () {
+          final originalPrivateClassLine = 'class _\$AbstractStateProps {';
+          final originalPublicClassLine = 'class AbstractStateProps extends _\$AbstractStateProps with _\$AbstractStatePropsAccessorsMixin {';
```
evanweible-wf:
```suggestion
          final originalPublicClassLine = 'abstract class AbstractFooState extends _\$AbstractFooState with _\$AbstractFooStateAccessorsMixin {';
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+          );
+
+          var transformedSource = transformedFile.getTransformedText();
+          expect(transformedSource, contains(originalPrivateClassLine));
+          expect(transformedSource, isNot(contains(originalPublicClassLine)));
+          expect(transformedSource, contains(transformedFooStateLine));
+        });
+
+        test('with builder-compatible dual-class abstract state setup', () {
+          final originalPrivateClassLine = 'class _\$AbstractStateProps {';
```
evanweible-wf:
```suggestion
          final originalPrivateClassLine = 'abstract class _\$AbstractFooState {';
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+            class FooProps extends UiProps {}
+            
+            @Component()
+            class FooComponent {
+              render() => null;
+            }
+          
+            class AbstractFooProps extends _\$AbstractFooProps with _\$AbstractFooPropsAccessorsMixin {
+              // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+              static const AbstractFooPropsMeta meta = \$metaForAbstractFooProps;
```
evanweible-wf:
```suggestion
              static const PropsMeta meta = \$metaForAbstractFooProps;
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+          expect(transformedSource, isNot(contains(originalPublicFooStateLine)));
+          expect(transformedSource, contains(transformedFooStateLine));
+          expect(transformedSource, contains(fooStateImplExtendsPublicClass));
+          expect(transformedSource, isNot(contains(fooStateImplExtendsPrivateClass)));
+        });
+
+        test('with builder-compatible dual-class abstract props setup', () {
+          final originalPrivateClassLine = 'class _\$AbstractFooProps {';
+          final originalPublicClassLine = 'class AbstractFooProps extends _\$AbstractFooProps with _\$AbstractFooPropsAccessorsMixin {';
+          final transformedFooStateLine = 'class AbstractFooProps extends _\$AbstractFooProps';
```
evanweible-wf: add `abstract` to both of these

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -397,6 +397,157 @@ main() {
...
+        
+            class FooProps extends _\$FooProps with _\$FooPropsAccessorsMixin {
+              // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+              static const PropsMeta meta = \$metaForFooProps;
+            }
+           
+            @Props()
+            class _\$FooProps extends UiProps {}
+            
+            @AbstractProps() class AbstractFooProps {
```
evanweible-wf: Only testing props here, so we should drop the `AbstractProps`

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -420,6 +497,71 @@ main() {
...
+              UiFactory<FooProps> Foo;
+              
+              @Props()      
+              class FooProps {}
+              
+              @Component()  
+              class FooComponent {}
+              
+              @AbstractState() 
+              class _\$AbstractStateProps {}
```
evanweible-wf:
```suggestion
              abstract class _\$AbstractStateProps {}
```

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -420,6 +497,71 @@ main() {
...
+              UiFactory<FooProps> Foo;
+              
+              @Props()      
+              class FooProps {}
+              
+              @Component()  
+              class FooComponent {}
+              
+              @AbstractProps() 
+              class _\$AbstractFooProps {}
```
evanweible-wf:
```suggestion
              abstract class _\$AbstractFooProps {}
```

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -222,14 +285,28 @@ main() {
...
           expect(declarations.abstractState[1].meta, new isInstanceOf<annotations.AbstractState>());
 
           expectEmptyDeclarations(abstractState: false);
           expect(declarations.declaresComponent, isFalse);
         });
 
+        test('abstract state class with builder-compatible dual-class setup', () {
+          setUpAndParse('''
+            @AbstractState() class _\$AbstractFooState {}
+            class AbstractFooState {}
```
evanweible-wf:
```suggestion
            abstract class AbstractFooState {}
```

test/vm_tests/transformer/declaration_parsing_test.dart
```diff
@@ -205,14 +254,28 @@ main() {
...
           expect(declarations.abstractProps[1].meta, new isInstanceOf<annotations.AbstractProps>());
 
           expectEmptyDeclarations(abstractProps: false);
           expect(declarations.declaresComponent, isFalse);
         });
 
+        test('abstract props class with builder-compatible dual-class setup', () {
+          setUpAndParse('''
+            @AbstractProps() class _\$AbstractFooProps {}
+            class AbstractFooProps {}
```
evanweible-wf:
```suggestion
            abstract class AbstractFooProps {}
```

PR #204, opened 2018-11-26
AF-3086: transformer generates empty $ sign prefixed mixin classes

sebastianmalysa-wf:
## Ultimate problem:

With the builder-compat boilerplate, Props and State mixin classes are renamed to include a $ prefix with the assumption that the actual class with concrete accessor implementations will be generated.

```dart
@PropsMixin()
abstract class FooPropsMixin {
+  // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+  static const PropsMeta meta = $metaForBarPropsMixin;
+
  Map get props;
  
  String fooProp;
}

// --- mixin usages ---

@Props()
class BarProps extends UiProps
-     with FooPropsMixin {
+     with
+         FooPropsMixin,
+         // TODO: AF-#### This will be removed once the transition to Dart 2 is complete.
+         // ignore: mixin_of_non_class,undefined_class
+         $FooPropsMixin {
  // ...
}
```

Previously, the transformer would inline these implementations over top of the fields. To support this setup, the transformer needs to inline an implementation of the public mixin class that extends the $-prefixed class. With this approach, the mixin clauses for consuming these mixin classes can be left as-is.

## How it was fixed:

- When a $ sign prefixed mixin is found in the props|state class with clause an empty abstract $ sign prefixed mixin classes is generated.

## Testing suggestions:

- CI passes 
- Code changes and tests make sense 

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @evanweible-wf @corwinsheahan-wf

### DIFF
--- a/lib/src/transformer/impl_generation.dart
+++ b/lib/src/transformer/impl_generation.dart
@@ -167,6 +167,16 @@ class ImplGenerator {
       // ----------------------------------------------------------------------
       generateAccessors(AccessorType.props, declarations.props);
 
+      // Generate an empty abstract $ sign prefixed props mixins when found since Dart 2 builder compatible
+      // boiler plate prefixes props mixins with a $ sign and adds the mixin to props class via the with clause.
+      if (declarations.props.node.withClause != null) {
+        declarations.props.node.withClause.mixinTypes.forEach((type) {
+          if (type.toString().startsWith('\$')) {
+            transformedFile.insert(sourceFile.location(declarations.factory.node.end), '\n\nabstract class $type {}');
+          }
+        });
+      }
+
       final String propKeyNamespace = getAccessorKeyNamespace(declarations.props);
 
       implementations
@@ -211,6 +221,17 @@ class ImplGenerator {
       //   State implementation
       // ----------------------------------------------------------------------
       if (declarations.state != null) {
+
+        // Generate an empty abstract $ sign prefixed state mixins when found since Dart 2 builder compatible
+        // boiler plate prefixes state mixins with a $ sign and adds the mixin to state class via the with clause.
+        if (declarations.state.node.withClause != null) {
+          declarations.state.node.withClause.mixinTypes.forEach((type) {
+            if (type.toString().startsWith('\$')) {
+              transformedFile.insert(sourceFile.location(declarations.factory.node.end), '\n\nabstract class $type {}');
+            }
+          });
+        }
+
         final String stateName = declarations.state.node.name.toString();
         final String stateImplName = '$generatedPrefix${stateName}Impl';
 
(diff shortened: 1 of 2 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/transformer/impl_generation.dart
```diff
@@ -167,6 +167,22 @@ class ImplGenerator {
       // ----------------------------------------------------------------------
       generateAccessors(AccessorType.props, declarations.props);
 
+      // Remove $ sign prefixed props mixins when found in the props class with clause.
+      if (declarations.props.node.withClause != null &&
+          declarations.props.node.withClause.mixinTypes.any((type) => type.toString().startsWith('\$'))) {
```
evanweible-wf: I don't think this is going to be sufficient. This entire section is within a `if (declarations.declaresComponent) {` block, and it's only checking with clauses from the props and state classes. But PropsMixins and StateMixins can be used anywhere. I think we'll need to take a slightly different approach where we search every file for with clauses, unfortunately, something like:

1. Check file for class declarations
2. Check mixinTypes for the dual-setup (`<SomeMixinType>` AND `$<SomeMixinType>`)
3. For each pair found, remove the `$`-prefixed mixin

evanweible-wf: Also I think we should be able to add a test case for this (class in a file without a component declaration) that should fail right now

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -328,6 +328,63 @@ main() {
...
+            class FooProps extends UiProps with FooPropsMixin,
+            // TODO: AF-#### This will be removed once the transition to Dart 2 is complete.
+            // ignore: mixin_of_non_class,undefined_class
+            \$FooPropsMixin, BarPropsMixin, \$BarPropsMixin {}
+
+            @State()
+            class FooState extends UiState with FooStateMixin,
+            // TODO: AF-#### This will be removed once the transition to Dart 2 is complete.
+            // ignore: mixin_of_non_class,undefined_class
+            \$FooStateMixin {}
```
evanweible-wf: Since the implementation generation is no longer relying on props or state classes that mixin specific props or state mixins, I think we can ditch this boilerplate just to slim down and simplify the test a bit. Really all we need to test is that all instances of props mixins and state mixins have their `$`-prefixed versions generated, so a test with 2 props mixins and 2 state mixins in the same file should be sufficient

evanweible-wf: In other words, you can ditch the component, factory, props, and state pieces and then add another state mixin

PR #206, opened 2018-12-04
AF-1896 Add PropsMeta and StateMeta classes as alternative to $Props/$PropKeys

evanweible-wf:
**_DEPENDS ON https://github.com/Workiva/over_react/pull/203_**

## Ultimate problem:
The existing API for obtaining a list of props or a list of prop keys from a props class is via `$Props` or `$PropKeys`, but these options rely on the transformer to replace these instances inline with actual lists. Going forward in Dart2 with a builder, this approach will not work.

The alternative is to introduce `PropsMeta` and `StateMeta` classes and (eventually) have the builder generate a static `meta` field on the generated version of every consumer-defined props and state class. This meta field will map to a `PropsMeta` or `StateMeta` instance that is generated by the builder.

This PR introduces this approach ahead of the Dart 2 transition by adding the classes and updating the transformer to generate the meta instances for every props and state class, which will allow consumers to migrate away from `$Props()` and `$PropKeys()` usages ahead of time.

## How it was fixed:
- Transformer now generates a `_$<Class>Meta` class for every class annotated with one of:
  - `@Props()`
  - `@State()`
  - `@PropsMixin()`
  - `@StateMixin()`
  - `@AbstractProps()`
  - `@AbstractState()`

    As a part of the Dart1->Dart2 transition, props, state, abstract props, and abstract state classes will be renamed to have a `_$` prefix and an accompanying public class will be either generated by the builder, or during the transition, generated by a codemod script and committed. The transformer takes this into account and will strip the `_$` prefix from the annotated classes so that it generates the correct meta instance name.

- ~~All props and state classes in `over_react` itself have been updated to include the static meta field, and all references to `$Props()` and `$PropKeys()` have been updated accordingly. This was done so that existing tests can verify that the new `PropsMeta` and `StateMeta` classes and the accompanying transformations are working correctly.~~
  - Instead of doing this, we're going to run the complete codemod as the last commit in the series of commits that are required to update over_react transformer for builder compatibility.

## Testing suggestions:
- TBD

- `$Props()` and `$PropKeys()` usages

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @kealjones-wk @evanweible-wf @maxwellpeterson-wf

### DIFF
--- a/lib/src/component/abstract_transition.dart
+++ b/lib/src/component/abstract_transition.dart
@@ -21,6 +21,8 @@ import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart';
 import 'package:over_react/component_base.dart' as component_base;
 
+class TestPropsMeta extends PropsMeta {}
+
 @AbstractProps()
 abstract class AbstractTransitionProps extends UiProps with TransitionPropsMixin {}
 
--- a/lib/src/component/abstract_transition.dart
+++ b/lib/src/component/abstract_transition.dart
@@ -21,8 +21,6 @@ import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart';
 import 'package:over_react/component_base.dart' as component_base;
 
-class TestPropsMeta extends PropsMeta {}
-
 @AbstractProps()
 abstract class AbstractTransitionProps extends UiProps with TransitionPropsMixin {}
 
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -760,11 +760,16 @@ abstract class MapViewMixin<K, V> implements _OverReactMapViewBase<K, V> {
   Iterable<V> get values => _map.values;
 }
 
+abstract class _Descriptor {
+  String get key;
+}
+
 /// Provides a representation of a single `prop` declared within a [UiProps] subclass or props mixin.
 ///
 /// > Related: [StateDescriptor]
-class PropDescriptor {
+class PropDescriptor implements _Descriptor {
   /// The string key associated with the `prop`.
+  @override
   final String key;
   /// Whether the `prop` is required to be set.
   final bool isRequired;
@@ -779,8 +784,9 @@ class PropDescriptor {
 /// Provides a representation of a single `state` declared within a [UiState] subclass or state mixin.
 ///
 /// > Related: [PropDescriptor]
-class StateDescriptor {
+class StateDescriptor implements _Descriptor {
   /// The string key associated with the `state`.
+  @override
   final String key;
   /// Whether the `state` is required to be set.
   ///
@@ -808,4 +814,92 @@ class ConsumedProps {
   final List<String> keys;
 
   const ConsumedProps(this.props, this.keys);
+}
+
+abstract class AccessorMeta<T extends _Descriptor> {
+  List<T> get fields;
+  List<String> get keys;
+}
+
+/// Metadata for the prop fields declared in a specific props class--
+/// a class annotated with @[Props], @[PropsMixin], @[AbstractProps], etc.
+/// for which prop accessors are generated.
+///
+/// This metadata includes map key values corresponding to these fields, which
+/// is used in [UiComponent.consumedPropKeys], as well as other prop
+/// configuration done via @[Accessor]/@[requiredProp]/etc., which is used to
+/// perform prop validation within [UiComponent] lifecycle methods.
+///
+/// This metadata is generated as part of the over_react builder, and should be
+/// exposed like so:
+///     @Props()
+///     class FooProps {
+///       static const PropsMeta meta = $metaForFooProps;
+///
+///       String foo;
+///
+///       @Accessor(isRequired: true, key: 'custom_key', keyNamespace: 'custom_namespace')
+///       int bar;
+///     }
+///
+/// What the metadata looks like:
+///     main() {
+///       print(FooProps.meta.keys); // [FooProps.foo, custom_namespace.custom_key]
+///       print(FooProps.meta.props.map((p) => p.isRequired); // (false, true))
+///     }
+///
+/// _See also: [getPropKey]_
+class PropsMeta implements ConsumedProps, AccessorMeta<PropDescriptor> {
+  /// Rich views of prop field declarations.
+  ///
+  /// This includes string keys, and required prop validation related fields.
+  @override
+  final List<PropDescriptor> fields;
+
+  /// Top-level accessor of string keys of props stored in [fields].
+  @override
+  final List<String> keys;
+
+  const PropsMeta({this.fields, this.keys});
+
+  @override
+  List<PropDescriptor> get props => fields;
+}
+
+/// Metadata for the state fields declared in a specific state class--
+/// a class annotated with @[State], @[StateMixin], @[AbstractState], etc.
+/// for which state accessors are generated.
+///
+/// This metadata includes map key values corresponding to these fields, which
+/// is used to perform state validation within [UiComponent] lifecycle methods.
+///
+/// This metadata is generated as part of the over_react builder, and should be
+/// exposed like so:
+///     @State()
+///     class FooState {
+///       static const StateMeta meta = $metaForFooState;
+///
+///       String foo;
+///
+///       @Accessor(key: 'custom_key', keyNamespace: 'custom_namespace')
+///       int bar;
+///     }
+///
+/// What the metadata looks like:
+///     main() {
+///       print(FooState.meta.keys); // [FooState.foo, custom_namespace.custom_key]
+///       print(FooState.meta.fields.map((p) => p.key); // [FooState.foo, custom_namespace.custom_key]
+///     }
+class StateMeta implements AccessorMeta<StateDescriptor> {
+  /// Rich views of state field declarations.
+  ///
+  /// This includes string keys, and required state validation related fields.
+  @override
+  final List<StateDescriptor> fields;
+
+  /// Top-level accessor of string keys of props stored in [fields].
+  @override
+  final List<String> keys;
+
+  const StateMeta({this.fields, this.keys});
 }
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -888,7 +888,7 @@ class PropsMeta implements ConsumedProps, AccessorMeta<PropDescriptor> {
 /// What the metadata looks like:
 ///     main() {
 ///       print(FooState.meta.keys); // [FooState.foo, custom_namespace.custom_key]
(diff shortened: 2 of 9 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/component_declaration/component_base.dart
```diff
@@ -809,3 +815,39 @@ class ConsumedProps {
 
   const ConsumedProps(this.props, this.keys);
 }
+
+abstract class AccessorMeta<T extends _Descriptor> {
+  List<T> get fields;
+  List<String> get keys;
+}
+
+class PropsMeta implements ConsumedProps, AccessorMeta<PropDescriptor> {
```
evanweible-wf: Need to add doc comments for this class and `StateMeta`.. but I'm honestly not sure how we should describe these classes. @greglittlefield-wf could I get an assist from you?

lib/src/transformer/impl_generation.dart
```diff
@@ -617,6 +608,24 @@ class ImplGenerator {
         sourceFile.location(typedMap.node.leftBracket.end),
         staticVariablesImpl
     );
+
+    final name = typedMap.node.name.name;
+    final metaClassName = '$generatedPrefix${name}Meta';
+    final metaStructName = type == AccessorType.props
+        ? 'PropsMeta'
+        : 'StateMeta';
+    final metaInstanceName = '${publicGeneratedPrefix}metaFor$name'; // FIXME validate boilerplate to avoid typos
```
evanweible-wf: This note was from the original implementation, @greglittlefield-wf what validation did you have in mind?

lib/src/transformer/impl_generation.dart
```diff
@@ -617,6 +608,24 @@ class ImplGenerator {
...
+     var output = new StringBuffer();
+    output.writeln('/// A class that allows us to reuse generated code from the accessors class.');
+    output.writeln('/// This is only used by other generated code, and can be simplified if needed.');
+    output.writeln('class $metaClassName {');
+    output.writeln(staticVariablesImpl);
+    output.writeln('}');
+    output.writeln('const $metaStructName $metaInstanceName = const $metaStructName(');
+    output.writeln('  fields: $metaClassName.$constantListName,'); // todo don't include generated prefix
+    output.writeln('  keys: $metaClassName.$keyListName,'); // todo don't include generated prefix
```
evanweible-wf: These todos were also from the original implementation, but I don't understand them. @greglittlefield-wf is this saying that we should drop the `$` prefix from `$props`/`$state`/`$propKeys`/`$stateKeys`?

lib/src/transformer/impl_generation.dart
greglittlefield-wf: I believe this is referring to validating that the `meta` variable is referencing the correct generated variable.

For example, for `FooProps`,
```dart
class FooProps {
  static const PropsMeta meta = $metaForFooProps;
}
```
making sure they don't have `= $metaForBarProps` or something, which could easily happen as a form of copy-paste error.

lib/src/transformer/impl_generation.dart
greglittlefield-wf: That's a good question... whatever it is, I don't think it's super important, though; this code looks good to me as-is.

It might be referring to the fact that the variables are just copied from the generated accessors class to the `$metaClassName` class, prefix and all, when that's unnecessary.

Or, it could be referring to something related to the simplification noted in the generated doc comment on 620.

lib/src/component_declaration/component_base.dart
greglittlefield-wf: How about something like this?
```dart
/// Metadata for the prop fields declared in a specific props class--
/// a class annotated with @[Props], @[PropsMixin], @[AbstractProps], etc. 
/// for which prop accessors are generated.
///
/// This metadata includes map key values corresponding to these fields,
/// which is used in [UiComponent.consumedPropKeys],
/// as well as other prop configuration done via @[Accessor]/@[requiredProp]/etc.,
/// which is used to perform prop validation within [UiComponent] lifecycle methods.
///
/// This metadata is generated as part of the over_react builder, and should
/// be exposed like so:
///     @Props()
///     class FooProps {
///       static const PropsMeta meta = $metaForFooProps;
///     
///       String foo;
///
///       @Accessor(isRequired: true, key: 'custom_key', keyNamespace: 'custom_namespace')
///       int bar;
///     }
///
/// What the metadata looks like:
///     main() {
///       print(FooProps.meta.keys); // [FooProps.foo, custom_namespace.custom_key]
///       print(FooProps.meta.props.map((p) => p.isRequired); // (false, true))
///     }
///
/// _See also: [getPropKey]_
```

lib/src/transformer/impl_generation.dart
greglittlefield-wf: > I don't think it's super important, though

For more context, I usually use "fixme" comments if it's important, not todos 😄

lib/src/component/abstract_transition.dart
```diff
@@ -21,6 +21,8 @@ import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart';
 import 'package:over_react/component_base.dart' as component_base;
 
+class TestPropsMeta extends PropsMeta {}
```
greglittlefield-wf: Is this left over from debugging?

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -652,6 +731,33 @@ main() {
           });
         });
       });
+
+      group('static meta field', () {
```
greglittlefield-wf: It would be sweet if we also had an integration test for these.

It looks like we don't have any for `$PropKeys`/`$Props`, but there's this test that tests them indirectly: https://github.com/Workiva/over_react/blob/master/test/over_react/component_declaration/transformer_integration_tests/do_not_generate_accessor_integration_test.dart#L44

Thoughts on adding a test there for good measure?
```dart
        expect(DoNotGenerateAccessorTestProps.meta.propKeys, [
          contains('generated1Prop'),
          contains('generated2Prop'),
          contains('explicitlyGeneratedProp'),
        ], reason: 'should only include generated props');

        expect(DoNotGenerateAccessorTestProps.meta.map((prop) => prop.key).toList(), [
          contains('generated1Prop'),
          contains('generated2Prop'),
          contains('explicitlyGeneratedProp'),
        ], reason: 'should only include generated props');
```

test/vm_tests/transformer/impl_generation_test.dart
```diff
@@ -652,6 +731,33 @@ main() {
...
+            expect(transformedFile.getTransformedText(), contains(expectedMetaForInstance));
+          });
+        }
+
+        testStaticMetaField('props class', StaticMetaTest.props);
+        testStaticMetaField('state class', StaticMetaTest.state);
+        testStaticMetaField('props mixin', StaticMetaTest.propsMixin);
+        testStaticMetaField('state mixin', StaticMetaTest.stateMixin);
+        testStaticMetaField('abstract props', StaticMetaTest.abstractProps);
+        testStaticMetaField('abstract state', StaticMetaTest.abstractState);
```
greglittlefield-wf: So DRY 🤤

lib/src/component/abstract_transition.dart
evanweible-wf: yes it is 😨

lib/src/component_declaration/component_base.dart
```diff
@@ -809,3 +815,91 @@ class ConsumedProps {
...
+///       print(FooState.meta.fields.map((p) => p.key); // [FooState.foo, custom_namespace.custom_key]
+///     }
+class StateMeta implements AccessorMeta<StateDescriptor> {
+  /// Rich views of state field declarations.
+  ///
+  /// This includes string keys, and required state validation related fields.
+  @override
+  final List<StateDescriptor> fields;
+
+  /// Top-level accessor of string keys of props stored in [fields].
```
evanweible-wf: s/props/state

lib/src/component_declaration/component_base.dart
```diff
@@ -809,3 +815,91 @@ class ConsumedProps {
...
+///       String foo;
+///
+///       @Accessor(key: 'custom_key', keyNamespace: 'custom_namespace')
+///       int bar;
+///     }
+///
+/// What the metadata looks like:
+///     main() {
+///       print(FooState.meta.keys); // [FooState.foo, custom_namespace.custom_key]
+///       print(FooState.meta.fields.map((p) => p.key); // [FooState.foo, custom_namespace.custom_key]
```
evanweible-wf: should use `s` instead of `p`

lib/src/transformer/impl_generation.dart
```diff
@@ -574,49 +591,95 @@ class ImplGenerator {
...
+      );
+    }
+    
+    final name = typedMap.node.name.name;
+    final metaClassName = '$generatedPrefix${name}Meta';
+    final metaStructName = type == AccessorType.props
+        ? 'PropsMeta'
+        : 'StateMeta';
+    final metaInstanceName = '${publicGeneratedPrefix}metaFor$name'; // FIXME validate boilerplate to avoid typos
+     var output = new StringBuffer();
```
corwinsheahan-wf: #nit spacing

lib/src/transformer/impl_generation.dart
corwinsheahan-wf: validation implemented here: https://github.com/Workiva/over_react/pull/206/files#diff-92afc98c256ac39698eadfad4f3a3ce4R209

lib/src/transformer/impl_generation.dart
```diff
@@ -574,49 +591,95 @@ class ImplGenerator {
...
+    output.writeln('/// A class that allows us to reuse generated code from the accessors class.');
+    output.writeln('/// This is only used by other generated code, and can be simplified if needed.');
+    output.writeln('class $metaClassName {');
+    output.writeln(staticVariablesImpl);
+    output.writeln('}');
+    output.writeln('const $metaStructName $metaInstanceName = const $metaStructName(');
+    output.writeln('  fields: $metaClassName.$constantListName,'); // todo don't include generated prefix
+    output.writeln('  keys: $metaClassName.$keyListName,'); // todo don't include generated prefix
```
corwinsheahan-wf: #nit delete the `todo`'s

test/vm_tests/transformer/impl_generation_test.dart
corwinsheahan-wf: Can we get tests that verify the validation in the declaration parsing for Props/State meta is working properly?

evanweible-wf:
Merging into integration branch, will open a new PR with the entire integration branch against master along with some final changes, additional tests, and updates to the over_react component code itself.

PR #227, opened 2019-02-01
CPLAT-4031: Make builder compatible with Dart 2 only boilerplate

corwinsheahan-wf:
## Ultimate problem:
Currently the `2.0.0-dev` branch only contains support for building on the [transition boilerplate](https://github.com/Workiva/over_react/blob/master/doc/dart2_migration.md#temporary-transitional-boilerplate). 

## How it was fixed:
* Add support for Dart 2 only boilerplate by generating consumable companion class for `@Props()`, `@State()`, `@AbstractProps()` and `@AbstractState()` annotated classes, as well as generated consumable class for props mixins. 
* Is still backwards compatible with the transition boilerplate. 
## Testing suggestions:
* CI Passes (tests and examples added)

Everything

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @evanweible-wf @maxwellpeterson-wf @sebastianmalysa-wf @georgelesica-wf


aviary-wf:
## Security Insights

## Action Items
---

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -96,33 +117,24 @@ class ParsedDeclarations {
...
+          final companionName = member.name.name.substring(companionPrefix.length);
+          final companionClass = unit.declarations.firstWhere(
+                  (innerMember) =>
+              innerMember is ClassDeclaration && innerMember.name.name == companionName,
+              orElse: () => null);
+
+          final hasCompanionClass = companionClass != null;
+          if (hasCompanionClass) {
+            updateCompanionClass(name, true);
+            validateMetaField(companionClass, isPropsClass(name) ? 'PropsMeta': 'StateMeta');validateMetaField(companionClass, isPropsClass(name) ? 'PropsMeta': 'StateMeta');
```
greglittlefield-wf: Looks like this statement was duplicated

lib/src/builder/generation/impl_generation.dart
```diff
@@ -570,20 +574,21 @@ class ImplGenerator {
     }
 
     final consumerClassName = _classNameFromNode(node);
-    final accessorsMixinName = '\$$consumerClassName';
+    final accessorsMixinName = consumerClassName.startsWith(generatedPrefix) ?
```
greglittlefield-wf: #nit Should this be `companionPrefix`? (I know it's not declared in this file, but it would be nice to be consistent between this and impl_generation.dart)

lib/src/builder/generation/impl_generation.dart
```diff
@@ -652,9 +660,37 @@ class ImplGenerator {
         ..writeln())
         .toString();
   }
+
+  String _generateConsumablePropsOrStateClass(AccessorType type, NodeWithMeta node) {
+    if (!(node is PropsOrStateNode)) {
```
greglittlefield-wf: When would this evaluate to false (other than if the value is `null`)?

It seems like `node` could be typed as `PropsOrStateNode`.

lib/src/builder/generation/impl_generation.dart
```diff
@@ -652,9 +660,37 @@ class ImplGenerator {
...
+    final className = _classNameFromNode(node);
+    final typeParameters = (node.node as ClassDeclaration).typeParameters;
+    final typeParamsOnClass = typeParameters?.toSource() ?? '';
+    final typeParamsOnSuper = removeBoundsFromTypeParameters(typeParameters);
+    final consumableClassName = _publicPropsOrStateClassNameFromConsumerClassName(className);
+
+    final classKeywords = '${type == AccessorType.abstractProps || type == AccessorType.abstractState ? 'abstract ' : ''}class';
+    final isProps = type == AccessorType.props || type == AccessorType.abstractProps;
+    return (StringBuffer()
+      ..writeln('$classKeywords $consumableClassName$typeParamsOnClass extends $className$typeParamsOnSuper with ${className}AccessorsMixin$typeParamsOnSuper {')
```
greglittlefield-wf: Can we use `_accessorsMixinNameFromConsumerName` here?

lib/src/builder/generation/impl_generation.dart
```diff
@@ -652,9 +660,37 @@ class ImplGenerator {
...
+        .toString();
+  }
+
+  String _generateStaticMetaDecl(String consumableClassName, bool isProps) {
+    return '  static const ${isProps ? 'Props' : 'State'}Meta meta = ${_metaConstantName(consumableClassName)};';
+  }
 }
 
-enum AccessorType {props, state}
+enum AccessorType {props, state, abstractProps, abstractState}
```
greglittlefield-wf: There are still a handful of places where there are checks like:
```dart
final isProps = type == AccessorType.props;
```
Even if they're not problematic right now, they might be brittle against future changes.

What about changing this to something like:
```dart
class AccessorType {
  final bool isProps;
  final bool isAbstract;
  const AccessorType(bool isProps, bool isAbstract);

  static const AccessorType props = const AccessorType(true, false);
  static const AccessorType state = const AccessorType(false, false);
  static const AccessorType abstractProps = const AccessorType(true, true);
  static const AccessorType abstractState = const AccessorType(false, true);
}
```
and updating those checks:
```diff
-type == AccessorType.props
+type.isProps
```

lib/src/builder/generation/impl_generation.dart
evanweible-wf: also, for future reference, `is!`:
```
if (node is! PropsOrStateNode) {
```

lib/src/builder/generation/impl_generation.dart
```diff
@@ -570,20 +574,21 @@ class ImplGenerator {
     }
 
     final consumerClassName = _classNameFromNode(node);
-    final accessorsMixinName = '\$$consumerClassName';
+    final accessorsMixinName = consumerClassName.startsWith(generatedPrefix) ?
+        _publicPropsOrStateClassNameFromConsumerClassName(consumerClassName) :
+        '\$$consumerClassName';
```
evanweible-wf: My understanding is that the accessor mixins should be private, but I think both of these are returning values that are prefixed with `$` instead of `_$`. That being said, it appears to work as expected when using the builder locally.. so maybe I'm just missing something.

test/over_react/component_declaration/builder_integration_tests/abstract_accessor_integration_test.dart
```diff
@@ -0,0 +1,240 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+import 'package:over_react/over_react.dart';
+import 'package:test/test.dart';
+
+// ignore: uri_has_not_been_generated
+part 'abstract_accessor_integration_test.over_react.g.dart';
```
evanweible-wf: Just occurred to me that with build to source, we won't need the ignore comment here.

greglittlefield-wf: Did you mean to say "build to cache"?

evanweible-wf: I meant to say build to source – we've been experimenting with it and I was able to confirm this weekend that it works and is backwards-compat with Dart 1. Our hope is that it would provide a perf benefit because it also means we only have to build the root package and not all packages

lib/src/builder/generation/declaration_parsing.dart
corwinsheahan-wf: @evanweible-wf I was thinking about that this weekend. I don't think we should error, since there's really no issue with that class defining a static meta field in the `_$` prefixed class and the extending class. I think a warning is appropriate.

lib/src/builder/generation/impl_generation.dart
corwinsheahan-wf: How about `buildTargetPrefix`, since anything prefixed with `_$` will have an analogous built class. I'm not in love with it, since it overloads the "build target" terminology, but I can't think of anything better atm 🤷‍♂️ 

Thoughts?

lib/src/builder/generation/impl_generation.dart
corwinsheahan-wf: This would evaluate to false if the node is a mixin. See the class declarations here: https://github.com/Workiva/over_react/pull/227/files#diff-c832e6c6bfeea9404f15c0a9f034ff61R462

Props, State, AbstractProps, and AbstractState annotated classes have an accompanying consumable class in the transition boilerplate. Once we're on Dart 2, we will generate that class. The `PropsOrStateNode` class provides a field for whether this class already exists, so we can be backwards compatible with the transition boilerplate. Props and State Mixins never had this accompanying companion class.

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -30,6 +30,10 @@ class ParsedDeclarations {
   factory ParsedDeclarations(CompilationUnit unit, SourceFile sourceFile, Logger logger) {
```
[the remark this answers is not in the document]
corwinsheahan-wf: The builder does factory validation here: https://github.com/Workiva/over_react/blob/00b04578b6862a6b51ca9522305fa1271bab28c3/lib/src/builder/generation/declaration_parsing.dart#L291

I'll add validation on the part directive.

test/over_react/component_declaration/builder_integration_tests/abstract_accessor_integration_test.dart
evanweible-wf: For posterity, we're sticking with build to cache over source.

lib/src/builder/generation/impl_generation.dart
corwinsheahan-wf: That's true for props/state/abstractprops/abstractstate. This code is just for props/state mixins. For props/state mixins, the accessors mixin is either `$FooPropsMixin` or `FooPropsMixin`.

lib/src/builder/generation/impl_generation.dart
greglittlefield-wf: What about something like `privateSourcePrefix`? I was trying to find a precedent for referring to these names, and it looks like built_value uses the term "source" https://github.com/google/built_value.dart/tree/master/built_value_generator/lib/src

Other "companion" alternative name ideas:
- input/buildInput
- template/buildTemplate
- boilerplate
- precursor

corwinsheahan-wf: I like `privateSourcePrefix`

test/vm_tests/builder/impl_generation_test.dart
```diff
@@ -81,352 +81,528 @@ main() {
...
 
-            test('contains string keys', () {
-              expect(implGenerator.outputContentsBuffer.toString(), contains(
-                  '  static const String _\$key__someField___\$$className = \'$className.someField\';\n'
-                  '  static const String _\$key__foo___\$$className = \'$className.foo\';\n'
-                  '  static const String _\$key__bar___\$$className = \'$className.bar\';\n'
-                  '  static const String _\$key__baz___\$$className = \'$className.baz\';\n'));
-            });
+          test('component', () {
+            generateFromSource(OverReactSrc.props(backwardsCompatible: backwardsCompatible).source);
```
evanweible-wf: This isn't actually verifying anything is it?

lib/src/builder/generation/impl_generation.dart
```diff
@@ -600,25 +601,77 @@ class ImplGenerator {
...
+      // Otherwise, fields which are not synthetic have concrete getters/setters
+      // generated for them, so don't copy those over either. We also don't want
+      // to copy anything that would conflict with what the builder already may
+      // generate, namely `props`, `state`, and `meta`
+      final isValid = !_isStaticFieldOrMethod(member)
+          && !(member is FieldDeclaration && !member.isSynthetic)
+          && !_memberHasName(member, 'props')
+          && !_memberHasName(member, 'state')
+          && !_memberHasName(member, 'meta');
+//      logger.warning('''checking if member is valid for copying
```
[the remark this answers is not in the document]
aaronlademann-wf: # LOG

test/vm_tests/builder/impl_generation_test.dart
corwinsheahan-wf: Actually, in the `tearDown` for the containing group, there's two verifications: `verifyNoErrorLogs()`, and `verifyImplGenerationIsValid()`.

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -321,18 +333,18 @@ class ParsedDeclarations {
...
-        error(
-            'Factory variables are stubs for the generated factories, and should not have initializers '
-                'unless initialized with a valid variable name for Dart 2 builder compatibility. '
-                'Should be:\n    $expectedInitializer}',
-            getSpan(sourceFile, variable.initializer)
-        );
+        if ((variable?.initializer?.toString() ?? '') != expectedInitializer) {
+          error(
+              'Factory variables are stubs for the generated factories, and should not have initializers '
+                  'unless initialized with a valid variable name for Dart 2 builder compatibility. '
```
evanweible-wf: This message is leaking some legacy. I think we can drop the "...should not have initializers unless initialized with..." part and just tell them that they "should be initialized with <expected initializer>".

evanweible-wf:
@corwinsheahan-wf Just noticed this when reading the docs updates - any reason the generated factory implementation needs to be public?

https://github.com/Workiva/over_react/blob/afbee2fffca0611e0be3c6b3ef0efdef54e967af/lib/src/builder/generation/impl_generation.dart#L526

corwinsheahan-wf:
@evanweible-wf That does need to be public. When a component is a subtype of another component, the super component's factory is referenced as part of the factory for the sub component:
```
parentType: _$SuperComponentFactory,
```

evanweible-wf:
QA +1
- [x] CI passes
- [x] Verified locally against two internal consumers using backwards-compat boilerplate and the dart2-only boilerplate.

@Workiva/release-management-p

PR #228, opened 2019-02-04
CPLAT-4038 Update transformer to support dart2 builder outputs to source.

evanweible-wf:
## Ultimate problem:
We're going to configure the over_react dart2 builder to build to source, meaning that the generated `.over_react.g.dart` outputs will be included alongside package source files. Previously, the transformer was assuming that these part files would not exist, so it would generate no-op files where necessary to ensure the build succeeds. Now that they may actually exist, the transformer needs to remove everything in them except the `part of` directive. This will not affect runtime functionality at all – the transformer is already providing (effectively) the same code that exists in these part files, just in different locations.

## How it was fixed:
- Don't output a new asset for a `.over_react.g.dart` part file if the asset already exists (would be a duplicate output transformer error)
- Transform all `.over_react.g.dart` files to remove everything after the `part of` directive.

## Testing suggestions:
- Consumer test this change to ensure no regressions in the transformer behavior.

- Transformer.

---

> __FYA:__ @greglittlefield-wf @aaronlademann-wf @kealjones-wk @evanweible-wf @maxwellpeterson-wf @sebastianmalysa-wf @georgelesica-wf

### DIFF
--- a/lib/over_react.dart
+++ b/lib/over_react.dart
@@ -27,7 +27,7 @@ export 'package:react/react.dart' show
     SyntheticUIEvent,
     SyntheticWheelEvent;
 
-export 'package:react/react_client.dart' show setClientConfiguration, ReactElement;
+export 'package:react/react_client.dart' show setClientConfiguration, ReactElement, ReactComponentFactoryProxy;
 
 export 'src/component/abstract_transition.dart';
 export 'src/component/abstract_transition_props.dart';
--- a/lib/transformer.dart
+++ b/lib/transformer.dart
@@ -98,64 +98,73 @@ class WebSkinDartTransformer extends Transformer implements LazyTransformer {
     TransformedSourceFile transformedFile = new TransformedSourceFile(sourceFile);
     TransformLogger logger = new JetBrainsFriendlyLogger(transform.logger);
 
-    var partPattern = new RegExp(r'''part\s+['"](.+\.over_react\.g\.dart)['"];''');
-    var partFilename = partPattern.firstMatch(sourceFile.getText(0));
-
-    // For Dart 1 compatibility an empty generated part file will be created when a file contains
-    // the part directive pointing to the generated file the new builder requires.
-    if (partFilename != null) {
-      var sourceFileDirectory = p.dirname(transform.primaryInput.id.path);
-      var sourceFilename = p.basename(transform.primaryInput.id.path);
-      var partFilePath = p.join(sourceFileDirectory, partFilename.group(1));
-      var contents = "part of '$sourceFilename';";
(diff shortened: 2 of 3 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/over_react.dart
```diff
@@ -27,7 +27,7 @@ export 'package:react/react.dart' show
     SyntheticUIEvent,
     SyntheticWheelEvent;
 
-export 'package:react/react_client.dart' show setClientConfiguration, ReactElement;
+export 'package:react/react_client.dart' show setClientConfiguration, ReactElement, ReactComponentFactoryProxy;
```
evanweible-wf: This is a bonus change that I'm including in this PR. In 2.0.0-dev, a breaking change to the type of `UiProps.componentFactory` from `Function` to the more specific type of `ReactComponentFactoryProxy`. To help with the transition, we should export this symbol on 1.x as well so that consumers get valid overrides when using the builder's generated outputs on a Dart1/over_react1.

lib/transformer.dart
```diff
@@ -98,64 +98,73 @@ class WebSkinDartTransformer extends Transformer implements LazyTransformer {
     TransformedSourceFile transformedFile = new TransformedSourceFile(sourceFile);
```
evanweible-wf: Most of the diff in this file is just whitespace (indenting one level to be within the `else` block of the first check).

lib/transformer.dart
```diff
@@ -98,64 +98,73 @@ class WebSkinDartTransformer extends Transformer implements LazyTransformer {
...
+      var partPattern = new RegExp(r'''part\s+['"](.+\.over_react\.g\.dart)['"];''');
+      var partFilename = partPattern.firstMatch(sourceFile.getText(0));
+
+      // For Dart 1 compatibility an empty generated part file will be created when a file contains
+      // the part directive pointing to the generated file the new builder requires.
+      if (partFilename != null) {
+        var sourceFileDirectory = p.dirname(transform.primaryInput.id.path);
+        var partFilePath = p.join(sourceFileDirectory, partFilename.group(1));
+        var partAssetId = new AssetId(transform.primaryInput.id.package, partFilePath);
+        if (!(await transform.hasInput(partAssetId))) {
```
evanweible-wf: This is how we prevent a duplicate asset output transformer error when dealing with builder outputs being committed to source.

corwinsheahan-wf:
We are no longer going to explore building to source. This can cause problems propagating builder updates. For example, for any root package, all upstream packages will need to commit generated fixes for the root package to consume a builder update.

PR #249, opened 2019-02-18
CPLAT-3616 Warn consumers about props / state mutation

joebingham-wk:
## Ultimate problem:
With the upcoming change to the component_base class, setting the state / props directly will cause breakages. To give consumers a heads up, we should emit warnings in the browser console when they directly mutate props / state values inside a component instance.

## How it was fixed:
Adding a class called WarnOnUpdate that watches for setting props / state directly and throws an error in dev mode using an Assert.

## Testing suggestions:
Run the test suite. Consider any ways the tests could return false results or miss a use case.

None that are known as it's only in dev that the error will be thrown.

### DIFF
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -245,7 +245,7 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component imple
     var unwrappedProps = this.unwrappedProps;
     var typedProps = _typedPropsCache[unwrappedProps];
     if (typedProps == null) {
-      typedProps = typedPropsFactory(unwrappedProps);
+      typedProps = typedPropsFactory(WarnOnModify(unwrappedProps));
       _typedPropsCache[unwrappedProps] = typedProps;
     }
     return typedProps;
@@ -415,7 +415,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
     var unwrappedState = this.unwrappedState;
     var typedState = _typedStateCache[unwrappedState];
     if (typedState == null) {
-      typedState = typedStateFactory(unwrappedState);
+      typedState = typedStateFactory(WarnOnModify(unwrappedState));
       _typedStateCache[unwrappedState] = typedState;
     }
     return typedState;
@@ -444,6 +444,18 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   // ----------------------------------------------------------------------
 }
 
+class WarnOnModify<K, V> extends MapView<K, V> {
+  WarnOnModify(Map componentData): super(componentData);
+  
+  @override
+  operator []=(K key, V value) {
+    assert(
+      super[key] != value,
+      'Mutating the state or props directly will soon be impossible without causing breakages.'
+    );
+    super[key] = value;
+  }
+}
 
 /// A `dart.collection.MapView`-like class with strongly-typed getters/setters for React state.
 ///
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -446,7 +446,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
 
 class WarnOnModify<K, V> extends MapView<K, V> {
   WarnOnModify(Map componentData): super(componentData);
-  
+
   @override
   operator []=(K key, V value) {
     assert(
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -16,6 +16,7 @@ library over_react.component_declaration.component_base;
 
 import 'dart:async';
 import 'dart:collection';
+import 'dart:html';
 
 import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart';
@@ -245,7 +246,7 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component imple
     var unwrappedProps = this.unwrappedProps;
     var typedProps = _typedPropsCache[unwrappedProps];
     if (typedProps == null) {
-      typedProps = typedPropsFactory(WarnOnModify(unwrappedProps));
+      typedProps = typedPropsFactory(inReactDevMode ? _WarnOnModify(unwrappedProps, true) : unwrappedProps);
       _typedPropsCache[unwrappedProps] = typedProps;
     }
     return typedProps;
@@ -415,7 +416,7 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
     var unwrappedState = this.unwrappedState;
     var typedState = _typedStateCache[unwrappedState];
     if (typedState == null) {
-      typedState = typedStateFactory(WarnOnModify(unwrappedState));
+      typedState = typedStateFactory(inReactDevMode ? _WarnOnModify(unwrappedProps, false) : unwrappedProps);
       _typedStateCache[unwrappedState] = typedState;
     }
     return typedState;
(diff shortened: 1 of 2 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -827,6 +836,13 @@ main() {
 
             expect(stateBeforeChange, isNot(same(stateAfterChange)));
           });
+
+          test('warns against setting state directly', () {
+
```
kealjones-wk: #nit This new line isn't needed

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +444,18 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
...
 }
 
+class WarnOnModify<K, V> extends MapView<K, V> {
+  WarnOnModify(Map componentData): super(componentData);
+
+  @override
+  operator []=(K key, V value) {
+    assert(
+      false,
+      'Mutating the state or props directly will soon be impossible without causing breakages.'
```
aaronlademann-wf:
```suggestion
      '''
      Direct mutation of component props / state values directly is unsupported - and often 
      causes unexpected behavior(s) within UiComponent / UiStatefulComponent implementations.
      
      In the soon-to-be-released UiComponentV2 / UiStatefulComponentV2 classes, 
      direct mutation of these values will result in a runtime error.
      
      The mutation of props should be done by passing in a new value from a 
      parent / wrapper component or data store.
      
      The mutation of state should be done by calling setState(newState()..someKey = someValue).
      '''
```

@greglittlefield-wf I would love your thoughts on the accuracy / wording of this message.

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +444,18 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   // ----------------------------------------------------------------------
 }
 
+class WarnOnModify<K, V> extends MapView<K, V> {
```
aaronlademann-wf: This should be private [_(add an underscore to the beginning of the name)_](https://www.dartlang.org/guides/language/language-tour#important-concepts).

lib/src/component_declaration/component_base.dart
greglittlefield-wf: Looks good, but I think we should keep the error message terse that consumers don't glaze over it, and as few lines as possible so that it doesn't blow up the logs.

I think we should also include more information on what was mutated so that the consumer knows where to fix it. might also be nice to have more directed versions for props/state:
```
props["$key"] was updated incorrectly. Never mutate this.props directly, as it can cause unexpected behavior; props must be updated only by passing in new values when rerendering this component.

This will throw in UiComponentV2 (to be released as part of the React 16 upgrade).
```
```
state["$key"] was updated incorrectly. Never this.state directly, as it can cause unexpected behavior; state must be updated only via setState.

This will throw in UiComponentV2 (to be released as part of the React 16 upgrade).
```

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +444,18 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
...
+  WarnOnModify(Map componentData): super(componentData);
+
+  @override
+  operator []=(K key, V value) {
+    assert(
```
greglittlefield-wf: This shouldn't be an assert, since we don't want this to throw and break existing consumer code during development. 

```dart
operator []=(K key, V value) {
  window.console.warn(...)
  super[key] = value;
}
```

lib/src/component_declaration/component_base.dart
```diff
@@ -245,7 +245,7 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component imple
     var unwrappedProps = this.unwrappedProps;
     var typedProps = _typedPropsCache[unwrappedProps];
     if (typedProps == null) {
-      typedProps = typedPropsFactory(unwrappedProps);
+      typedProps = typedPropsFactory(WarnOnModify(unwrappedProps));
```
greglittlefield-wf: Since we don't want to emit this warning in prod, we can wrap these in checks so that this map is used only in dev mode:
```
typedProps = typedPropsFactory(inReactDevMode ? WarnOnModify(unwrappedProps) : unwrappedProps);
``` 
Doing it here as opposed to in WarnOnModify has the advantage of not allocating extra objects unnecessarily in prod.

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +445,38 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
...
 }
 
+class _WarnOnModify<K, V> extends MapView<K, V> {
+
+  //Used to customize warning based on whether the data is props or state
+  bool isProps;
+
+  String message;
+
+  _WarnOnModify(Map componentData, bool isProps): super(componentData){
```
kealjones-wk: Not required to change this but just to let you know, Dart Constructors are pretty smart... You can actually set class level fields without a body by doing this: `_WarnOnModify(Map componentData, bool this.isProps): super(componentData);` Cool right?

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +445,38 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
...
+    this.isProps = isProps;
+  }
+
+  @override
+  operator []=(K key, V value) {
+    if (isProps) {
+      message =
+        '''
+          props["$key"] was updated incorrectly. Never mutate this.props directly, as it can cause unexpected behavior; props must be updated only by passing in new values when rerendering this component.
```
kealjones-wk: Our team follows the rules that lines should never exceed 120 characters in length, including comments and strings. You can set up a column marker in your IDE. These lines should all be wrapped manually onto new lines before they hit 120 characters.
```suggestion
          props["$key"] was updated incorrectly. Never mutate this.props directly, as it can cause unexpected behavior; 
          props must be updated only by passing in new values when re-rendering this component.
```

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +445,38 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
...
+
+          This will throw in UiComponentV2 (to be released as part of the React 16 upgrade).
+        ''';
+    } else {
+      message =
+        '''
+          state["$key"] was updated incorrectly. Never mutate this.state directly, as it can cause unexpected behavior; state must be updated only via setState.
```
kealjones-wk: Same goes for here.
```suggestion
          state["$key"] was updated incorrectly. Never mutate this.state directly, as it can cause unexpected behavior; 
          state must be updated only via setState.
```

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +445,38 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   // ----------------------------------------------------------------------
 }
 
+class _WarnOnModify<K, V> extends MapView<K, V> {
+
```
kealjones-wk: #nit Unnecessary new line

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +445,37 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
   // ----------------------------------------------------------------------
 }
 
+class _WarnOnModify<K, V> extends MapView<K, V> {
+  //Used to customize warning based on whether the data is props or state
+  bool isProps;
+
+  String message;
+
+  _WarnOnModify(Map componentData, this.isProps): super(componentData);
```
kealjones-wk:
```suggestion
  _WarnOnModify(Map componentData, bool this.isProps): super(componentData);
```

lib/src/component_declaration/component_base.dart
```diff
@@ -444,6 +445,37 @@ abstract class UiStatefulComponent<TProps extends UiProps, TState extends UiStat
...
+
+  @override
+  operator []=(K key, V value) {
+    if (isProps) {
+      message =
+        '''
+          props["$key"] was updated incorrectly. Never mutate this.props directly, as it can cause unexpected behavior;
```
greglittlefield-wf: Does this currently print out to the console all indented? If so, we should wrap it in an unindent.

greglittlefield-wf:
+1, will merge pending travis build

greglittlefield-wf:
+10

PR #280, opened 2019-04-01
CPLAT-5152: Optimize builder's use of asset reads to avoid unnecessary rebuilds

evanweible-wf:
# [CPLAT-5152](https://jira.atl.workiva.net/browse/CPLAT-5152)
Issue Status

## Motivation
  
We've seen unexpectedly large rebuild times in codebases that consume this package and its builder. After some digging, it has been determined that the main reason for this is due to how the over_react builder is currently implemented:

The `PartBuilder` util from the `source_gen` package always retrieves the `LibraryElement` for each file that is a library (i.e. not a part file). This requires that the analysis session resolves the entire element tree for that file, which is expensive but more importantly results in reads on all transitively imported files. These reads are then stored in the asset graph as inputs for that build step, which consequently means that the over_react builder will be re-run on many files unnecessarily because the inputs have changed.

## Changes
  
The over_react builder only needs the AST for each file, so the solution here is to replace our usage of PartBuilder with a custom Builder that manually reads the library source for each BuildStep and calls the analyzer API `parseCompilationUnit()` to get the info needed to generate the `.over_react.g.dart` part files.

## Perf Improvement

Some numbers to demonstrate the performance improvement delivered by this change:

**Average clean build time:**

> Averaged over 5 runs.
> `pub run build_runner clean && pub run build_runner build`

Project | Before | After
------- | ------ | -----
web_skin_dart | 122s | 107s
wdesk_sdk | 112s | 45s

**Average re-build time:**

> Averaged over 5 runs. Rebuilt after changing one entry point file in `lib/`
> that has a large number of imports/exports.
>
> Change file -> `pub run build_runner build`

Project | File | Before | After
------- | ---- | ------ | -----
web_skin_dart | `lib/ui_core.dart` | 42.9s | 21.22s
wdesk_sdk | `lib/app_infrastructure.dart` | 63s | 2.2s

#### Release Notes
  
Optimize the over_react builder to avoid unnecessary asset reads. Informal testing in a large codebase that consumes `over_react` has shown in the worst case a speed up on hot-rebuilds from ~1 minute to ~2 seconds.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/builder/builder.dart
```diff
@@ -1,8 +1,144 @@
...
+
+    final outputs = <String>[];
+    final libraryGenerator = new ImplGenerator(log, librarySourceFile)
+      ..generate(libraryDeclarations);
+    outputs.add(libraryGenerator.outputContentsBuffer.toString());
+
+    final parts = libraryUnit.directives
+      .whereType<PartDirective>()
+      // Ignore `.over_react.g.dart` parts - that's what we're generating.
+      .where((part) => !part.uri.stringValue.endsWith(outputExtension));
```
corwinsheahan-wf: We may also need to ignore parts from other generated files. That was an issue with the old approach, did you verify that it's not an issue here?

lib/src/builder/builder.dart
```diff
@@ -1,8 +1,144 @@
...
+    for (final part in parts) {
+      final partId = new AssetId.resolve(
+        part.uri.stringValue,
+        from: buildStep.inputId);
+      final partSource = await buildStep.readAsString(partId);
+      final partSourceFile = new SourceFile.fromString(
+        partSource,
+        url: idToPackageUri(partId));
+      CompilationUnit partUnit;
+      try {
```
corwinsheahan-wf: This code is very similar to the parsing and generating for the main library code above. What's the reason not to DRY this up?

lib/src/builder/builder.dart
```diff
@@ -1,8 +1,144 @@
...
+    final outputId = buildStep.inputId.changeExtension(outputExtension);
+    final partOf = "'${p.basename(buildStep.inputId.uri.toString())}'";
+
+    final buffer = new StringBuffer()
+      ..writeln('// GENERATED CODE - DO NOT MODIFY BY HAND')
+      ..writeln()
+      ..writeln('part of $partOf;')
+      ..writeln()
+      ..writeln(_headerLine)
+      ..writeln('// OverReactGenerator')
```
corwinsheahan-wf: #nit Personally, I'd advocate for not using the same header as the generator. It could lead to confusion if a consumer is examining our outputs.

lib/src/builder/builder.dart
evanweible-wf: Yeah, the main difference is that I have to parse the compilation unit for the library earlier so that I can check the directives for a PartOfDirective (i.e. to see if the file is actually a part file) so that we can bail early. I can probably dry it up though

lib/src/builder/builder.dart
evanweible-wf: Fair enough, I was going for minimal changes to generated code but I guess that doesn't really matter since we're building to cache.

lib/src/builder/builder.dart
evanweible-wf: I didn't see any issues, but this reminds me that I meant to add an another check for each part directive so that we only parse the compilation units if the part file contains over react declarations. Currently this is only doing that check on the parent library file

corwinsheahan-wf: I think it would be safer to ignore anything ending in `.g.dart` and explicitly not support over_react annotations in generated files. We could run into a situation where we're parsing a file that doesn't exist yet.

evanweible-wf:
@greglittlefield-wf @corwinsheahan-wf Just pushed a commit to tidy up the builder implementation a bit.

> Have we let the Dart team know about this performance issue?
Not yet, I want to do a few more tests with a simpler builder and simpler codebase that still uses `buildStep.inputLibrary` and then I plan on asking in their gitter chat about that cost in general. Will probably end up opening an issue in the `build` project around documenting those potential pitfalls as well as an issue in `source_gen` since there is currently no way to opt out of that behavior with their utils.

lib/src/builder/builder.dart
```diff
@@ -1,8 +1,136 @@
...
+    final outputs = <String>[];
+    void generateForFile(String source, AssetId id, CompilationUnit unit) {
+      if (!ParsedDeclarations.mightContainDeclarations(source)) {
+        return;
+      }
+      final sourceFile = new SourceFile.fromString(
+        source, url: idToPackageUri(id));
+      final declarations = new ParsedDeclarations(unit, sourceFile, log);
+      if (declarations.hasErrors) {
+        log.severe('There was an error parsing the file declarations for file: ${buildStep.inputId}');
```
evanweible-wf: Shoot, this should be `$id` instead of `${buildStep.inputId}`

evanweible-wf: Fixed.

lib/src/builder/builder.dart
```diff
@@ -33,88 +25,93 @@ class OverReactBuilder extends Builder {
       return;
```
corwinsheahan-wf: I'm not sure how feasible it is without actually examining the source of the part files, but it would be nice to get a warning as a consumer if I've got a parent library file whose parts contain over_react components, but which is missing the `*.over_react.g.dart` part directive.

evanweible-wf: Addressed in 93d4593

evanweible-wf:
Some numbers to demonstrate the performance improvement delivered by this change:

**Average clean build time:**

Project | Before | After
------- | ------ | -----
web_skin_dart | 122s | 107s
wdesk_sdk | 112s | 45s

> Averaged over 5 runs.

---

**Average re-build time:**

Project | File | Before | After
------- | ---- | ------ | -----
web_skin_dart | `lib/ui_core.dart` | 42.9s | 21.22s
wdesk_sdk | `lib/app_infrastructure.dart` | 63s | 2.2s

>  Averaged over 5 runs. Rebuilt after changing one entry point file in `lib/` that has a large number of imports/exports.

lib/src/builder/builder.dart
```diff
@@ -21,10 +21,6 @@ class OverReactBuilder extends Builder {
   @override
   FutureOr<void> build(BuildStep buildStep) async {
     final source = await buildStep.readAsString(buildStep.inputId);
-    if (!_mightContainDeclarations(source)) {
```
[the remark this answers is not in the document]
greglittlefield-wf: This diff is outdated, but still applies to the new version of the file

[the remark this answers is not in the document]
evanweible-wf: Without this change, we are no longer providing a warning in the scenario where a consumer has over_react-annotated code but did not include the `.over_react.g.dart` part file (which is something that the `PartBuilder` from source_gen did for us.

I didn't post these numbers but I did run a comparison between this commit and the previous commit and there was no noticeable difference in clean build or rebuild times on WSD or wdesk_sdk. I was a bit surprised by that, but my best guess is that the file reads and the AST parsing are cached.. or maybe they're just both fast? I'm not totally sure.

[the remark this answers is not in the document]
greglittlefield-wf: That's a bit surprising to me, too. I know file reads are pretty fast, but I expected AST parsing to not be super cheap. 🤷‍♂️ 

I'm fine with it as-is, then!

evanweible-wf:
QA +1
- [x] CI passes
- [x] Consumer tested with WSD and wdesk_sdk

evanweible-wf:
@Workiva/release-management-p

PR #300, opened 2019-05-22
CPLAT-5909 Update error boundary component to use the init method

joebingham-wk:
## Motivation
  
The `init` lifecycle method was introduced in part to allow for set to be state prior to component mounting. With the current code generation script, the typing conflicts because the generated code expects a `JsBackedMap` while react-dart accepts any `Map`. This results in a runtime error.
## Changes
  
- Modify the generation script to cast the passed in `Map` to a `JsBackedMap`.
- Update the stateful component test to use the `init` method to ensure it behaves the same as `getInitialState`.
- Replace `getInitialState` with `init` within `error_boundary_mixins.dart`.
## Review

  

Please review: 
@aaronlademann-wf @greglittlefield-wf @kealjones-wk 
### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] Pull this branch down 
      - [ ] Run `webdev serve web`
      - [ ] Verify the error boundary component works while using the `init()` lifecycle method. 

## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

lib/src/builder/generation/impl_generation.dart
```diff
@@ -247,8 +247,9 @@ class ImplGenerator {
             ..writeln()
             ..writeln('  @override')
             ..writeln('  set state(Map value) {')
-            ..writeln('    super.state = value;')
-            ..writeln('    _cachedTypedState = typedStateFactoryJs(value);')
+            ..writeln('    JsBackedMap jsBackedValue = JsBackedMap.from(value);')
```
greglittlefield-wf: This is creating a copy of the state map, which we want to avoid when possible:
- for performance (one of the reasons Component2 is faster is fewer map copies).
- to ensure the state value is in sync with `this.state`


Sorry, we should have groomed this ticket before working on it. I think instead of making this setter suboptimal, we should drop support for initializing state via `this.state = ` and instead create a method in Component2 like `initalizeState(Map value)` or something that should be used instead. Thoughts?

If we do go that route, I think it would be nice here to add an assert that it is indeed a JsBackedMap in case someone tries to do it the React JS way anyways.
```dart
assert(value is JsBackedMap, 
    'Component2.state may only be set to a JsBackedMap, '
        'and must not be set outside of react-dart internals.')
```

joebingham-wk: Oh that makes sense. Sorry about that - I feel like I should have seen that one. 

For avoiding `this.state =`, we already overrode the `[]=` operator to discourage setting state / props directly in https://github.com/Workiva/over_react/pull/249, so that solution makes sense and seems consistent. Implementation wise, it would be implementing the logic in react-dart and then exposing in in over_react, right @greglittlefield-wf ?

lib/src/builder/generation/impl_generation.dart
```diff
@@ -247,6 +247,9 @@ class ImplGenerator {
             ..writeln()
             ..writeln('  @override')
             ..writeln('  set state(Map value) {')
+            ..writeln('    assert(value is JsBackedMap, ')
+            ..writeln('      \'Component2.state may only be set to a JsBackedMap, \'')
+            ..writeln('        \'and must not be set outside of the react-dart internals.\');')
```
aaronlademann-wf: Can we make this message for helpful to the consumer by instructing them to use the `initializeState` method instead?

joebingham-wk: Yep, good call. Because it's the general state setter I updated the message to also mention setState. Let me know if the new message could be improved as well!

joebingham-wk:
Awesome, that explanation helps a lot. Thank you!

aaronlademann-wf:
+10

PR #301, opened 2019-05-29
CPLAT-4843 Check for Legacy Lifecycle Methods in Component2 Declaration

joebingham-wk:
## Motivation
  
Because of the inheritance structure of `Component2`, it is possible for a consumer to utilize lifecycle methods that no longer exist within React 16. To prohibit this, a check was needed to stop the builder if a component inheriting from `Component2` was calling one those methods.
## Changes
  
- Add a check in `declaration_parsing.dart` looking for the legacy lifecycle methods within `Component2`.
- Fix `ToggleButton` because it was breaking the build process.
- Add Tests
## Review

  

Please review: 
@aaronlademann-wf @greglittlefield-wf @kealjones-wk @sydneyjodon-wk 
### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] Pull down branch and make sure the `ToggleButton` example still behaves as expected.

## Merge Checklist

### DIFF
--- a/lib/src/builder/generation/declaration_parsing.dart
+++ b/lib/src/builder/generation/declaration_parsing.dart
@@ -282,7 +282,7 @@ class ParsedDeclarations {
       );
     }
 
-    // Ensure that Component2 declarations do not use removed lifecycle methods.
+    // Ensure that Component2 declarations do not use legacy lifecycle methods.
 
     if (declarationMap[key_component2].isNotEmpty) {
       if (declarationMap[key_component2].first is ClassDeclaration) {
--- a/lib/src/builder/generation/declaration_parsing.dart
+++ b/lib/src/builder/generation/declaration_parsing.dart
@@ -299,9 +299,11 @@ class ParsedDeclarations {
 
           if (method != null) {
             error(unindent(
-                '''Error within ${firstComponent2Member.name.name}.
+                '''
+                Error within ${firstComponent2Member.name.name}.
         
-                When using Component2, a class cannot use $method because React 16 has removed $method and renamed it UNSAFE_$method.
+                When using Component2, a class cannot use ${method.name} because React 16 has removed ${method.name} 
+                and renamed it UNSAFE_${method.name}.
                 
                 $helpMessage
                 
--- a/lib/src/builder/generation/declaration_parsing.dart
+++ b/lib/src/builder/generation/declaration_parsing.dart
@@ -16,6 +16,7 @@ import 'package:analyzer/analyzer.dart';
 import 'package:logging/logging.dart';
 import 'package:over_react/src/builder/util.dart';
 import 'package:over_react/src/component_declaration/annotations.dart' as annotations;
+import 'package:over_react/src/util/string_util.dart';
 import 'package:source_span/source_span.dart';
 import 'package:transformer_utils/transformer_utils.dart' show getSpan, NodeWithMeta;
 
@@ -297,14 +298,16 @@ class ParsedDeclarations {
           final method = firstComponent2Member.getMethod(methodName);
 
           if (method != null) {
-            error('''Error within ${firstComponent2Member.name.name}.
+            error(unindent(
+                '''Error within ${firstComponent2Member.name.name}.
         
                 When using Component2, a class cannot use $method because React 16 has removed $method and renamed it UNSAFE_$method.
                 
                 $helpMessage
                 
                 See https://reactjs.org/docs/react-component.html#legacy-lifecycle-methods for additional information.   
-                ''');
+                '''
+            ));
           }
         });
       }
--- a/lib/src/builder/generation/declaration_parsing.dart
+++ b/lib/src/builder/generation/declaration_parsing.dart
@@ -285,21 +285,26 @@ class ParsedDeclarations {
     // Ensure that Component2 declarations do not use legacy lifecycle methods.
 
     if (declarationMap[key_component2].isNotEmpty) {
-      if (declarationMap[key_component2].first is ClassDeclaration) {
-        ClassDeclaration componentObject = declarationMap[key_component2].first;
-        List<String> legacyLifecycleEvents =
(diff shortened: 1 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -282,6 +282,29 @@ class ParsedDeclarations {
...
+        List<String> legacyLifecycleEvents =
+            ['componentWillReceiveProps', 'componentWillMount', 'componentWillUpdate'];
+
+        legacyLifecycleEvents.forEach((lifecycle) {
+          dynamic method = componentObject.getMethod(lifecycle);
+
+          if (method != null) {
+            error(
+                'Error within ${componentObject.name.name}. \n\nWhen using Component2, a class cannot '
+                    'use $lifecycle because React 16 has removed $lifecycle and renamed it UNSAFE_$lifecycle. '
```
aaronlademann-wf: I'd like to see this error message suggest the replacement for the legacy methods.

We could pull that off by mapping the legacy lifecycle method to its replacement like so:

```dart
if (declarationMap[key_component2].isNotEmpty) {
  final firstComponent2Member = declarationMap[key_component2].first;
  if (firstComponent2Member is ClassDeclaration) {
    Map<String, String> legacyLifecycleMethodsMap = {
      'componentWillReceiveProps': 'Use getDerivedStateFromProps instead.',
      'componentWillMount': 'Use init instead.',
      'componentWillUpdate': 'Use getSnapshotBeforeUpdate instead.',
    };

    legacyLifecycleMethodsMap.forEach((methodName, helpMessage) {
      final method = firstComponent2Member.getMethod(methodName);

      if (method != null) {
        error('''Error within ${firstComponent2Member.name.name}.
        
            When using Component2, a class cannot use $method because React 16 has removed $method and renamed it UNSAFE_$method.
            
            $helpMessage
            
            See https://reactjs.org/docs/react-component.html#legacy-lifecycle-methods for additional information.   
            ''');
      }
    });
  }
}
```

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -282,6 +282,29 @@ class ParsedDeclarations {
       );
     }
 
+    // Ensure that Component2 declarations do not use legacy lifecycle methods.
+
+    if (declarationMap[key_component2].isNotEmpty) {
+      if (declarationMap[key_component2].first is ClassDeclaration) {
+        ClassDeclaration componentObject = declarationMap[key_component2].first;
+        List<String> legacyLifecycleEvents =
+            ['componentWillReceiveProps', 'componentWillMount', 'componentWillUpdate'];
```
aaronlademann-wf: I just noticed that `componentWillMount` is not deprecated.  I'm assuming we should do that in react-dart 5.0.0-wip, and recommend the use of `init` instead?

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -282,6 +282,29 @@ class ParsedDeclarations {
       );
     }
 
+    // Ensure that Component2 declarations do not use legacy lifecycle methods.
+
+    if (declarationMap[key_component2].isNotEmpty) {
+      if (declarationMap[key_component2].first is ClassDeclaration) {
+        ClassDeclaration componentObject = declarationMap[key_component2].first;
```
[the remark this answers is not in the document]
joebingham-wk: 💯 Ah, I see. That's cool!

lib/src/builder/generation/declaration_parsing.dart
joebingham-wk: Yep, great idea. Also, thank you for improving the comment formatting. That looks much better!

lib/src/builder/generation/declaration_parsing.dart
joebingham-wk: It's not deprecated in react-dart 5.0.0-wip, but it is in 5.1.0-wip. Since 5.0.0 doesn't have the new lifecycle methods, would the deprecation messages be just to give heads up not to rely on any of the legacy methods?

aaronlademann-wf: Hmmm... thats weird... when I pulled this branch down yesterday I didn't see any deprecation messages for `componentWillMount`...  Let me double-check.

aaronlademann-wf: nothing to see here... just had some cached deps that were out of date. all good!

lib/src/builder/generation/declaration_parsing.dart
```diff
@@ -282,6 +282,34 @@ class ParsedDeclarations {
...
+          'componentWillReceiveProps': 'Use getDerivedStateFromProps instead.',
+          'componentWillMount': 'Use init instead.',
+          'componentWillUpdate': 'Use getSnapshotBeforeUpdate instead.',
+        };
+
+        legacyLifecycleMethodsMap.forEach((methodName, helpMessage) {
+          final method = firstComponent2Member.getMethod(methodName);
+
+          if (method != null) {
+            error('''Error within ${firstComponent2Member.name.name}.
```
aaronlademann-wf: This is my bad @joebingham-wk ... but with `'''` comments, in order for the indentation of the source to not influence the output, you have to wrap the string block in `unindent()`, which you can import from `import 'package:over_react/src/util/string_util.dart';`.

Otherwise... the error output in console will look like this...

image

joebingham-wk: No problem - I've worked with comment blocks being logged out before and should have caught that. Thank you for pointing it out again!

joebingham-wk: @aaronlademann-wf that's done! Also, I updated the error message to use `method.name` to keep the message cleaner and not include the method annotations / body. Let me know if the other way was better though

aaronlademann-wf:

PR #305, opened 2019-06-17
CPLAT-2281 optimize copy unconsumed props

joebingham-wk:
## Motivation
  
To increase the performance of forwarding props, we are adding the `addUnconsumedProps` and `addUnconsumedDomProps` methods.
## Changes
  
- Add the `addUnconsumedProps` and `addUnconsumedDomProps` methods
- Deprecate the `copyUnconsumedProps` and `copyUnconsumedDomProps` methods
- Change `_validDomProps` from a `SplayTree` to a `Set`
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/component_declaration/component_base/component_base_2.dart
+++ b/lib/src/component_declaration/component_base/component_base_2.dart
@@ -120,8 +120,7 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
     forwardUnconsumedProps(this.props, propsToUpdate: props, keySetsToOmit:
         consumedPropKeys, onlyCopyDomProps: true);
   }
-
-
+  
   /// Returns a copy of this component's props with React props optionally omitted, and
   /// with the specified [keysToOmit] and [keySetsToOmit] omitted.
   @override
--- a/lib/src/component_declaration/component_base/component_base_2.dart
+++ b/lib/src/component_declaration/component_base/component_base_2.dart
@@ -87,7 +87,7 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   void addUnconsumedProps(UiProps props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    props.addProps(copyProps(keySetsToOmit: consumedPropKeys));
+    props.addEntries(copyPropsToForwardIntoMap(this.props, keySetsToOmit: consumedPropKeys));
   }
 
   /// Returns a copy of this component's props with keys found in [consumedProps] and non-DOM props omitted.
@@ -109,7 +109,8 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   void addUnconsumedDomProps(UiProps props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    props.addProps(copyProps(onlyCopyDomProps: true, keySetsToOmit: consumedPropKeys));
+    props.addEntries(copyPropsToForwardIntoMap(this.props, keySetsToOmit:
+        consumedPropKeys, onlyCopyDomProps: true));
   }
 
 
--- a/lib/src/component_declaration/component_base/component_base_2.dart
+++ b/lib/src/component_declaration/component_base/component_base_2.dart
@@ -87,7 +87,8 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   void addUnconsumedProps(UiProps props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    props.addEntries(copyPropsToForwardIntoMap(this.props, keySetsToOmit: consumedPropKeys));
+    copyPropsToForwardIntoMap(this.props, addProp: props.addProp,
+        keySetsToOmit: consumedPropKeys);
   }
 
   /// Returns a copy of this component's props with keys found in [consumedProps] and non-DOM props omitted.
@@ -109,8 +110,8 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   void addUnconsumedDomProps(UiProps props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    props.addEntries(copyPropsToForwardIntoMap(this.props, keySetsToOmit:
-        consumedPropKeys, onlyCopyDomProps: true));
+    copyPropsToForwardIntoMap(this.props, addProp: props.addProp, keySetsToOmit:
+        consumedPropKeys, onlyCopyDomProps: true);
   }
 
 
--- a/lib/src/component_declaration/component_base/component_base_2.dart
+++ b/lib/src/component_declaration/component_base/component_base_2.dart
@@ -84,10 +84,10 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
     return copyProps(keySetsToOmit: consumedPropKeys);
   }
 
-  void addUnconsumedProps(UiProps props) {
+  void addUnconsumedProps(Map props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    forwardUnconsumedProps(this.props, addProp: props.addProp,
+    forwardUnconsumedProps(this.props, addProp: (props as UiProps).addProp,
         keySetsToOmit: consumedPropKeys);
   }
 
@@ -108,10 +108,10 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
     return copyProps(onlyCopyDomProps: true, keySetsToOmit: consumedPropKeys);
   }
 
-  void addUnconsumedDomProps(UiProps props) {
+  void addUnconsumedDomProps(Map props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    forwardUnconsumedProps(this.props, addProp: props.addProp, keySetsToOmit:
+    forwardUnconsumedProps(this.props, addProp: (props as UiProps).addProp, keySetsToOmit:
         consumedPropKeys, onlyCopyDomProps: true);
   }
 
--- a/lib/src/component_declaration/component_base/component_base_2.dart
+++ b/lib/src/component_declaration/component_base/component_base_2.dart
@@ -73,9 +73,9 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   ///
   /// > Related [copyUnconsumedDomProps]
   ///
-  /// __Deprecated.__ Use [addUnconsumedProps] instead within
-  /// [modifyProps] (rather than [addProps]). Will be removed in the `4.0.0`
-  /// release.
+  /// __Deprecated.__ Use [addUnconsumedProps] within
+  /// [modifyProps] (rather than [addProps]) instead. Will be removed in the
+  /// `4.0.0` release.
   @override
   @Deprecated('4.0.0')
   Map copyUnconsumedProps() {
@@ -87,7 +87,7 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   void addUnconsumedProps(UiProps props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    copyPropsToForwardIntoMap(this.props, addProp: props.addProp,
+    forwardUnconsumedProps(this.props, addProp: props.addProp,
         keySetsToOmit: consumedPropKeys);
   }
 
@@ -97,8 +97,9 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   ///
   /// > Related [copyUnconsumedProps]
   ///
-  /// __Deprecated.__ Use [addUnconsumedDomProps] instead within
-  /// [modifyProps] (rather than [addProps]). Will be removed in the `4.0.0` release.
+  /// __Deprecated.__ Use [addUnconsumedDomProps] within
+  /// [modifyProps] (rather than [addProps]) instead. Will be removed in the
+  /// `4.0.0` release.
   @override
   @Deprecated('4.0.0')
   Map copyUnconsumedDomProps() {
@@ -110,7 +111,7 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
   void addUnconsumedDomProps(UiProps props) {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
-    copyPropsToForwardIntoMap(this.props, addProp: props.addProp, keySetsToOmit:
+    forwardUnconsumedProps(this.props, addProp: props.addProp, keySetsToOmit:
         consumedPropKeys, onlyCopyDomProps: true);
   }
 
--- a/lib/src/component_declaration/component_base/component_base_2.dart
+++ b/lib/src/component_declaration/component_base/component_base_2.dart
@@ -84,7 +84,18 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
     return copyProps(keySetsToOmit: consumedPropKeys);
   }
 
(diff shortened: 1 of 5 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/util/map_util.dart
```diff
@@ -78,6 +117,6 @@ Map<String, dynamic> newStyleFromProps(Map props) {
   return existingStyle == null ? <String, dynamic>{} : new Map.from(existingStyle);
 }
 
-SplayTreeSet _validDomProps = new SplayTreeSet()
+Set _validDomProps = new Set()
```
greglittlefield-wf: Actually, I think we should use `HashSet`, since it's faster than `Set`, which is a `LinkedHashSet`; see https://github.com/Workiva/wdesk_sdk/pull/3822

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,45 @@ Map getPropsToForward(Map props, {
...
+  bool onlyCopyDomProps: false,
+  Iterable keysToOmit,
+  Iterable<Iterable> keySetsToOmit,
+  Function addProp,
+}) {
+  props.forEach((key, value) {
+    if (onlyCopyDomProps) {
+      bool shouldAddProp = false;
+
+      /// todo: work in a way to skip checks if a single check returns true?
```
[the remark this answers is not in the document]
joebingham-wk: Ah yep, that's exactly what I wanted to do!

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,45 @@ Map getPropsToForward(Map props, {
...
+
+    if (keysToOmit != null && keysToOmit.contains(key)) return;
+
+    if (keySetsToOmit != null) {
+      keySetsToOmit.forEach((Iterable keySets) {
+        if (keySets.contains(key)) return;
+      });
+    }
+
+    if (omitReactProps && ['key', 'ref', 'children'].contains(key)) return;
```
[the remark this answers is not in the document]
joebingham-wk: 😮

lib/src/component_declaration/component_base/component_base_2.dart
```diff
@@ -72,25 +72,55 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
...
+  /// `4.0.0` release.
   @override
+  @Deprecated('4.0.0')
   Map copyUnconsumedProps() {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
     return copyProps(keySetsToOmit: consumedPropKeys);
   }
 
+  /// Passes a reference of a component's [UiProps] to be updated with unconsumed props.
```
aaronlademann-wf: We should mention that this should be called as an argument to `modifyProps` - and add an example of doing so:

```suggestion
  /// A prop modifier that passes a reference of a component's `props` to be updated with any unconsumed props.
  ///
  /// Call within `modifyProps` like so:
  ///
  ///     class SomeCompositeComponent extends UiComponent<SomeCompositeComponentProps> {
  ///       @override
  ///       render() {
  ///         return (SomeOtherWidget()..modifyProps(addUnconsumedProps(props)))(
  ///           props.children,
  ///         );
  ///       }
  ///     }
```


I'm also going to update the AC for the codemod (CPLAT-5212) to include:

```diff
- ..addProps(copyUnconsumedProps())
+ ..modifyProps(addUnconsumedProps(props))
```

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,58 @@ Map getPropsToForward(Map props, {
   return propsToForward;
 }
 
+/// Adds unconsumed props to a passed in [Map] reference ([propsToUpdate]).
+///
+/// Based upon configuration, the function will overlook [props] that are not
+/// meant to be passed on, such as non-DOM props or specified values.
+void forwardUnconsumedProps(Map props, {
```
aaronlademann-wf: Should this be advertised as a replacement for `getPropsToForward`?  If so, should `getPropsToForward` be deprecated?

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,58 @@ Map getPropsToForward(Map props, {
...
+        /// If the passed in value of [keySetsToOmit] comes from
+        /// [addUnconsumedProps], there should only be a single index.
+        /// Consequently, this case exists to give the opportunity for the loop
+        /// to continue without initiating another loop (which is less
+        /// performant than `.first.contains()`).
+        if (keySetsToOmit.first.contains(key)) continue;
+
+        if (keySetsToOmit.length > 1) {
+          bool shouldContinue = false;
+          for (Iterable keySet in keySetsToOmit) {
```
aaronlademann-wf: #nit I think `keySet` should be able to be inferred as an `Iterable` by the analyzer since `keySetsToOmit` is typed as `Iterable<Iterable>`.

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,58 @@ Map getPropsToForward(Map props, {
...
+            if (keySet.contains(key)) {
+              shouldContinue = true;
+              continue;
+            }
+          }
+
+          if (shouldContinue) continue;
```
aaronlademann-wf: Is this necessary? 

There is no logic below here which would be "skipped" by continuing to the next iteration in the `for (String key in props.keys)` loop as far as I can tell - which would make this `continue;` superfluous.

Did you mean for it to break out of the loop if `shouldContinue` is false?

```dart
if (!shouldContinue) break;
```

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,58 @@ Map getPropsToForward(Map props, {
...
+      }
+    }
+    return;
+  }
+
+  for (String key in props.keys) {
+    if (keysToOmit != null && keysToOmit.contains(key)) continue;
+
+    if (keySetsToOmit != null) {
+        /// If the passed in value of [keySetsToOmit] comes from
```
aaronlademann-wf: #nit everything inside this `if keySetsToOmit != null` block is indented 4 spaces, should be 2.

lib/src/util/map_util.dart
```diff
@@ -78,6 +130,6 @@ Map<String, dynamic> newStyleFromProps(Map props) {
   return existingStyle == null ? <String, dynamic>{} : new Map.from(existingStyle);
 }
 
-SplayTreeSet _validDomProps = new SplayTreeSet()
+HashSet _validDomProps = new HashSet()
```
aaronlademann-wf: Just want to verify that this is backwards-compatible with `addProps(copyUnconsumedProps())` - since that old pattern also uses this variable.

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -806,6 +806,101 @@ main() {
...
+          component2.addUnconsumedProps(newProps);
+
+          expect(newProps, equals({
+            'prop1': true,
+            'prop2': true,
+          }));
+        });
+      });
+
+      group('copyUnconsumedDomProps()', () {
```
aaronlademann-wf:
```suggestion
      group('addUnconsumedDomProps()', () {
```

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -806,6 +806,101 @@ main() {
       }, timeout: new Timeout(const Duration(milliseconds: 250)));
     });
 
+
+    group('UiComponent2', () {
+      TestComponent2Component component2;
+
+      group('copyUnconsumedProps()', () {
```
aaronlademann-wf:
```suggestion
      group('addUnconsumedProps()', () {
```

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -806,6 +806,101 @@ main() {
       }, timeout: new Timeout(const Duration(milliseconds: 250)));
     });
 
+
+    group('UiComponent2', () {
```
aaronlademann-wf: We should also copy over the tests for the deprecated `copyUnconsumedProps` / `copyUnconsumedDomProps` functions so that they are being tested with a `UiComponent2` instance.

test/over_react/component_declaration/component_base_test.dart
```diff
@@ -996,6 +1091,40 @@ class TestComponentComponent extends UiComponent<TestComponentProps> {
   }
 }
 
+class TestComponent2Props extends UiProps {
```
aaronlademann-wf: Is there a reason we wouldn't just create a new component declaration in a separate file for this stuff instead of manually creating classes that represent what the builder would generate?

Seems brittle to me to be hard-coding class members that are generated IRL.

test/over_react/util/map_util_test.dart
```diff
@@ -122,6 +122,120 @@ main() {
       });
     });
 
+    group('forwardUnconsumedProps() returns a copy of the specified props', () {
+      group('with React props', () {
+        test('omitted out by default', () {
+         Map startingProps = {
```
aaronlademann-wf: #nit this is only indented 1 space

test/over_react/util/map_util_test.dart
```diff
@@ -122,6 +122,120 @@ main() {
...
+            'key': 'my key',
+            'ref': 'my ref',
+            'other prop': 'my other prop'
+          };
+
+         Map actual = {};
```
aaronlademann-wf: #nit this is only indented 1 space

lib/src/component_declaration/component_base/component_base_2.dart
```diff
@@ -72,25 +72,55 @@ abstract class UiComponent2<TProps extends UiProps> extends react.Component2
...
+  /// `4.0.0` release.
   @override
+  @Deprecated('4.0.0')
   Map copyUnconsumedDomProps() {
     var consumedPropKeys = consumedProps?.map((ConsumedProps consumedProps) => consumedProps.keys) ?? const [];
 
     return copyProps(onlyCopyDomProps: true, keySetsToOmit: consumedPropKeys);
   }
 
+  /// Passes a reference of a component's [UiProps] to be updated with the unconsumed DOM props.
```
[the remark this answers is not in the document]
joebingham-wk: Good call! That's much more clear. Wouldn't we want the `modifyProps` line to actually be `..modifyProps(addUnconsumedDomProps)` though, @aaronlademann-wf? `addUnconsumedDomProps` is void because `modifyProps` uses it as a callback and passes in  itself for `props` parameter

lib/src/util/map_util.dart
joebingham-wk: That's a good point @aaronlademann-wf. I tried to look around and see if anyone was using `getPropsToForward` outside of the context of `addProps`, and there were a [few instances](https://github.com/search?p=1&q=getPropsToForward&type=Code).

I feel like the functions are different enough that it could be beneficial to have a more pure version of finding the unconsumed props without mutating a reference directly. So I'd lean towards it not being a replacement, but is that logic sound or am I missing anything?

lib/src/util/map_util.dart
joebingham-wk: Good call!

lib/src/util/map_util.dart
joebingham-wk: My intend was to skip the statements on [lines 117-119](https://github.com/Workiva/over_react/blob/50e497ee5191d5db1289b63657d60d9e81aaaa26/lib/src/util/map_util.dart#L117-L119), where the key is added to the props reference. Since the loop ends on line 120, I believe those lines would be dodged by the continue. Did I miss something and that's not actually the case?

lib/src/util/map_util.dart
joebingham-wk: I can't find any reason why it wouldn't be! It's used pretty limitedly, and the only methods the object needs are `addAll` and `contains`. There are also still unit tests (and manual tests that I had done), and there haven't been any red flags!

test/over_react/component_declaration/component_base_test.dart
joebingham-wk: Yep, good thought!

test/over_react/component_declaration/component_base_test.dart
joebingham-wk: I agree it is less than ideal. @kealjones-wk and I actually asked the same thing originally - we may have had other thoughts that I can't recall, but the big one is that the way it's set up now the test can easily specify the consumed props through the constructor. That way someone can actually write tests with the consumed props changing and you can validate everything works with pretty minimal code. To do the same thing through generated code seemed more complex. 

With all that being the case, would it be better to put the effort in to create the generated componentry?

lib/src/component_declaration/component_base/component_base_2.dart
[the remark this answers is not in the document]
aaronlademann-wf: derp derp, yes.

lib/src/util/map_util.dart
aaronlademann-wf: No, keeping the other one around sounds reasonable.

lib/src/util/map_util.dart
aaronlademann-wf: No - you're right. This is just another example of the pitfalls of reviewing code before 6am ;)

test/over_react/component_declaration/component_base_test.dart
aaronlademann-wf: Looks like we are already doing this in other places, and its not important enough to break pattern for now.

lib/src/util/map_util.dart
joebingham-wk: Hahah yep, that would do it!

joebingham-wk:
Yep! Here's the spreadsheet with all of the tests recorded: 
https://sandbox.wdesk.com/a/QWNjb3VudB81NDg4MTEzOTQ5OTk5MTA0/spreadsheet/c270726ab395458586aae2b90b309fa4/-1/sheet/2a9eefcc19134f91838d8d16d9a8622a

Summery: 
Using `addProps(copyUnconsumedProps)`:
- Average Time: 180ms
- Range: 78ms (with outlier) / 33ms (without outlier)
Using `modifyProps(addUnconsumedProps)`:
- Average Time: 165.76ms
- Range: 14ms

Speed Increase: 8%

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,58 @@ Map getPropsToForward(Map props, {
...
+
+  for (String key in props.keys) {
+    if (keysToOmit != null && keysToOmit.contains(key)) continue;
+
+    if (keySetsToOmit != null) {
+      /// If the passed in value of [keySetsToOmit] comes from
+      /// [addUnconsumedProps], there should only be a single index.
+      /// Consequently, this case exists to give the opportunity for the loop
+      /// to continue without initiating another loop (which is less
+      /// performant than `.first.contains()`).
```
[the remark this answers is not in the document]
greglittlefield-wf: Actually, I wonder if it has to do with the lazy iterable returned by `.map`, which gets its recomputed every time it's called.

Adding a `.toList()` to that (and eventually caching a flattened list of all key sets later on) could potentially speed things up.

test/over_react/util/map_util_test.dart
```diff
@@ -122,6 +122,120 @@ main() {
       });
     });
 
+    group('forwardUnconsumedProps() returns a copy of the specified props', () {
```
greglittlefield-wf: This description needs to be updated; this function does not return anything

lib/src/util/map_util.dart
```diff
@@ -68,6 +68,58 @@ Map getPropsToForward(Map props, {
...
+      if (keySetsToOmit.first.contains(key)) continue;
+
+      if (keySetsToOmit.length > 1) {
+        bool shouldContinue = false;
+        for (final keySet in keySetsToOmit) {
+          if (keySet.contains(key)) {
+            shouldContinue = true;
+            continue;
```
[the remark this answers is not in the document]
joebingham-wk: Yes! Thank you for catching that!

test/over_react/util/map_util_test.dart
joebingham-wk: Yep - should have caught that one. Thank you!

lib/src/util/map_util.dart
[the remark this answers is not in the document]
joebingham-wk: Tested and discovered that in basic cases, `.toList()` slows things down a little. There is, however, more optimization that can be done around the data structures and how exactly the logic loops through the sets. So I added a `todo` for that!

greglittlefield-wf:
+10

PR #390, opened 2019-10-21
CPLAT-7979: Add SafeRenderManager

greglittlefield-wf:
## Motivation
There were some cases where top-level `react_dom.render`/`.unmountComponentAtNode` calls were being done from within the event handlers and lifecycle methods of React components, which encountered timing-related issues in React 16.

## Changes
- Add SafeRenderManager:

> A class that manages the top-level rendering of a `ReactElement` into a given node,
with support for safely rendering/updating via `render` and safely unmounting via `tryUnmount`.
> 
> Content is also unmounted when this object is `dispose`d.
> 
> This is useful in cases where `react_dom.render` or `react_dom.unmountComponentAtNode`
may or may not be called from React events or lifecycle methods, which can have
undesirable/unintended side effects.
> 
> For instance, calling `react_dom.unmountComponentAtNode` can unmount a component
while an event is being propagated through a component, which normally would never happen.
This could result in null errors in the component as the event logic continues.
> 
> SafeRenderManager uses a helper component under the hood to manage the rendering of content
> via Component state changes, ensuring that the content is mounted/unmounted as it
normally would be.

- Add tests
- Adjust two lints

#### Release Notes
  
- Add SafeRenderManager, a class that manages the top-level rendering of a `ReactElement` into a given node, with support for safely rendering, updating, and unmounting content.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        - Testing branches pass when this is pulled in

## Merge Checklist

### DIFF
--- a/lib/src/util/safe_render_manager/safe_render_manager.dart
+++ b/lib/src/util/safe_render_manager/safe_render_manager.dart
@@ -1,0 +1,224 @@
+import 'dart:async';
+import 'dart:html';
+
+import 'package:meta/meta.dart';
+import 'package:over_react/over_react.dart';
+import 'package:over_react/react_dom.dart' as react_dom;
+import 'package:w_common/disposable.dart';
+
+import './safe_render_manager_helper.dart';
+
+/// A class that manages the top-level rendering of a [ReactElement] into a given node,
+/// with support for safely rendering/updating via [render] and safely unmounting via [tryUnmount].
+///
+/// Content is also unmounted when this object is [dispose]d.
+///
+/// This is useful in cases where [react_dom.render] or [react_dom.unmountComponentAtNode]
+/// may or may not be called from React events or lifecycle methods, which can have
+/// undesirable/unintended side effects.
+///
+/// For instance, calling [react_dom.unmountComponentAtNode] can unmount a component
+/// while an event is being propagated through a component, which normally would never happen.
+/// This could result in null errors in the component as the event logic continues.
+///
+/// SafeRenderManager uses a helper component under the hood to manage the rendering of content
+/// via Component state changes, ensuring that the content is mounted/unmounted as it
+/// normally would be.
+class SafeRenderManager extends Disposable {
+  SafeRenderManagerHelperComponent _helper;
+
+  /// Whether to automatically add [mountNode] to the document body when
+  /// rendered, and remove it when unmounted.
+  ///
+  /// Useful when manually managing the mount node isn't necessary.
+  final bool autoAttachMountNode;
+
+  /// The mount node for content rendered by [render].
+  ///
+  /// If not specified, a new div will be used.
+  final Element mountNode;
+
+  /// The ref to the component rendered by [render].
+  ///
+  /// Due to react_dom.render calls not being guaranteed to be synchronous.
+  /// this may not be populated until later than expected.
+  dynamic contentRef;
+
+  _RenderState _state = _RenderState.unmounted;
+
+  /// A list of [render] calls queued up while the component is in the process
+  /// of rendering.
+  List<ReactElement> _renderQueue = [];
+
+  SafeRenderManager({Element mountNode, this.autoAttachMountNode = false})
+      : mountNode = mountNode ?? new DivElement();
+
+  /// Renders [content]into [mountNode], chaining existing callback refs to
+  /// provide access to the rendered component via [contentRef].
+  void render(ReactElement content) {
+    _checkDisposalState();
+
+    switch (_state) {
+      case _RenderState.mounting:
+        _renderQueue.add(content);
+        break;
+      case _RenderState.mountedOrErrored:
+        // Handle if _helper was unmounted due to an uncaught error.
+        if (_helper == null) {
+          _mountContent(content);
+        } else {
+          _helper.renderContent(content);
+        }
+        break;
+      case _RenderState.unmounted:
+        _mountContent(content);
+        break;
+    }
+  }
+
+  void _mountContent(ReactElement content) {
+    try {
+      _state = _RenderState.mounting;
+      // Use document.contains since `.isConnected` isn't supported in IE11.
+      if (autoAttachMountNode && !document.contains(mountNode)) {
+        document.body.append(mountNode);
+      }
+      react_dom.render((SafeRenderManagerHelper()
+        ..ref = _helperRef
+        ..getInitialContent = () {
+          final value = content;
+          // Clear this closure variable out so it isn't retained.
+          content = null;
+          return value;
+        }
+        ..contentRef = _contentCallbackRef
+      )(), mountNode);
+    } catch (_) {
+      _state = _RenderState.unmounted;
+      rethrow;
+    }
+  }
+
+  /// Attempts to unmount the rendered component, calling [onMaybeUnmounted]
+  /// with whether the component was actually unmounted.
+  ///
+  /// Unmounting could fail if a call to [render] is batched in with this
+  /// unmount during the propagation of this event. In that case, something
+  /// other call wanted something rendered and trumped the unmount request.
+  ///
+  /// This behavior allows the same SafeRenderManager instance to be used to
+  /// render/unmount a single content area without calls interfering with each
+  /// other.
+  ///
+  /// If nothing is currently rendered, [onMaybeUnmounted] will be called immediately.
+  void tryUnmount({void onMaybeUnmounted(bool isUnmounted)}) {
+    // Check here since we call _tryUnmountContent in this class's disposal logic.
+    _checkDisposalState();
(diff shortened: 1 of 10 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/util/safe_render_manager/safe_render_manager_helper.dart
```diff
@@ -0,0 +1,68 @@
...
+
+// ignore: uri_has_not_been_generated
+part 'safe_render_manager_helper.over_react.g.dart';
+
+/// A component that allows for safe unmounting of its single child by waiting for state changes
+/// sometimes queued by ReactJS to be applied.
+@Factory()
+UiFactory<SafeRenderManagerHelperProps> SafeRenderManagerHelper =
+// ignore: undefined_identifier
+_$SafeRenderManagerHelper;
```
aaronlademann-wf: #supernit should be continuation indent

lib/src/util/safe_render_manager/safe_render_manager_helper.dart
```diff
@@ -0,0 +1,68 @@
...
+
+@Props()
+class _$SafeRenderManagerHelperProps extends UiProps {
+  @requiredProp
+  ReactElement Function() getInitialContent;
```
aaronlademann-wf: Currently the master branch has to remain Dart 1 and Dart 2 friendly.  IIRC, inline function typedef syntax like this causes problems in Dart 1

lib/src/util/safe_render_manager/safe_render_manager.dart
```diff
@@ -0,0 +1,197 @@
...
+///
+/// This is useful in cases where [react_dom.render] or [react_dom.unmountComponentAtNode]
+/// may or may not be called from React events or lifecycle methods, which can have
+/// undesirable/unintended side effects.
+///
+/// For instance, calling [react_dom.unmountComponentAtNode] can unmount a component
+/// while an event is being propagated through a component, which normally would never happen.
+/// This could result in null errors in the component as the event logic continues.
+///
+/// SafeRenderManager uses a helper  component under the hood to manage the rendering of content
```
joebingham-wk: #supernit
```suggestion
/// SafeRenderManager uses a helper component under the hood to manage the rendering of content
```

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
```diff
@@ -0,0 +1,707 @@
...
+
+      test('when autoAttachMountNode is true', () {
+        renderManager = new SafeRenderManager(autoAttachMountNode: true);
+        expect(document.contains(renderManager.mountNode), isFalse, reason: 'test setup check');
+
+        renderManager.render(Dom.div()());
+        expect(renderManager.mountNode.parent, document.body);
+
+        renderManager.tryUnmount();
+        expect(renderManager.mountNode.parent, isNull);
```
aaronlademann-wf: I realize this is probably kosher - but when I first read this - I thought... _it... null'd `document.body`??!!?!!?_

Not saying you have to change it, but something like this would test the same thing:

```dart
expect(document.body.contains(renderManager.mountNode), isFalse);
```

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
```diff
@@ -0,0 +1,707 @@
...
+      });
+
+      test('unless autoAttachMountNode is false', () {
+        renderManager = new SafeRenderManager(autoAttachMountNode: false);
+        expect(renderManager.mountNode.parent, isNull, reason: 'test setup check');
+
+        renderManager.render(Dom.div()());
+        expect(renderManager.mountNode.parent, isNull);
```
aaronlademann-wf: If `mountNode` is not specified, and `autoAttachMountNode` is false... isn't that an invalid configuration? What would it render into?

I realize this test is not really testing that - I'm just saying that the conditions of this test made me think this probably shouldn't be supported.

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
```diff
@@ -0,0 +1,707 @@
...
+              // happen only when rendering is deferred...
+              try {
+                triggerRenders();
+              } on TestExceptionThrownFromLifecycle catch (_) {
+                // ...but rethrow them if we're not expecting them so they aren't swallowed.
+                if (!isThrowingTest) rethrow;
+              }
+
+              if (isThrowingTest) {
+                expect(mountNode.text, isEmpty, reason: 'test setup check: React should have unmounted throwing component tree');
```
aaronlademann-wf: Will we just remove this logic when we port this change into `2.x` stable?

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
```diff
@@ -0,0 +1,707 @@
...
+                  verifyDeferredRender: true,
+                  setUpAndReturnTriggerRender: (doRenders) {
+                    document.body.append(renderManager.mountNode);
+                    renderManager.render((Wrapper()
+                      ..onClick = (_) {
+                        doRenders();
+                      }
+                    )('setup render'));
+
+                    // Use a real click since simulated clicks don't trigger this async behavior
```
aaronlademann-wf: 💯

lib/src/util/safe_render_manager/safe_render_manager.dart
```diff
@@ -0,0 +1,224 @@
...
+        // Don't do this in a finally since onMaybeUnmounted can get called async.
+        callOnMaybeUnmounted(true);
+        rethrow;
+      }
+    } else {
+      try {
+        _unmountContent();
+      } finally {
+        callOnMaybeUnmounted(true);
+      }
```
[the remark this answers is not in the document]
greglittlefield-wf: No, since there isn't a `onMaybeUnmounted` callback being used here.

lib/src/util/safe_render_manager/safe_render_manager_helper.dart
```diff
@@ -0,0 +1,70 @@
...
+@Component()
+class SafeRenderManagerHelperComponent extends UiStatefulComponent<SafeRenderManagerHelperProps, SafeRenderManagerHelperState> {
+  @override
+  getInitialState() => (newState()..content = props.getInitialContent());
+
+  void renderContent(ReactElement content) {
+    setState(newState()..content = content);
+  }
+
+  void tryUnmountContent({void onMaybeUnmounted(bool isUnmounted)}) {
```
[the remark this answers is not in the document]
greglittlefield-wf: This form, which only works in arguments, actually works great in Dart 1.

The new function syntax, which works only in Dart 2, would be:
```
({void Function(bool isUnmounted) onMaybeUnmounted})
```

lib/src/util/safe_render_manager/safe_render_manager_helper.dart
```diff
@@ -0,0 +1,70 @@
...
+    setState(newState()..content = null, () {
+      onMaybeUnmounted?.call(state.content == null);
```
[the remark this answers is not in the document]
greglittlefield-wf: Yes; this can happen if this setState is batched with another setState that sets it to something else.

lib/src/util/safe_render_manager/safe_render_manager_helper.dart
```diff
@@ -0,0 +1,70 @@
...
+    });
+  }
+
+  bool get hasContent => state.content != null;
+
+  @override
+  render() {
+    final content = state.content;
```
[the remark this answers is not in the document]
greglittlefield-wf: Sorry, this is leftover from this being a react-dart component that used string keys for state. Will update

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
greglittlefield-wf: I think this is good to keep in as a sanity check

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
greglittlefield-wf: I actually had this exact thing at first, but then changed it to what I have now so that the failure messages would be more helpful.

> Expected: null
> Actual: BodyElement

vs

> Expected: false 
> Actual: true

test/over_react/util/safe_render_manager/safe_render_manager_test.dart
greglittlefield-wf: No, it's a valid configuration. The manager defaults to a new, empty div if one isn't specified, and the consumer would be responsible for attaching/detaching it from the DOM.

PR #435, opened 2019-12-04
CPLAT-8600: Unrecoverable ErrorBoundary Fix

kealjones-wk:
## Motivation
  
Even with ErrorBoundaries in place, we were seeing issues with the app still entirely unmounting even with the logic for "unrecoverable" errors. It turns out in React if an ErrorBoundary catches an error, it will have its error lifecycle methods called, if on that render cycle the component throws during the render cycle React will move up the tree to a parent ErrorBoundary instead. And due to all the Previous ErrorBoundaries having the same logic none would ever handle the render cycle error.

## Changes
  
- Updated `ErrorBoundaryComponent` to handle render cycle unrecoverable errors (hopefully this wont happen often.)
- Moved the previous `ErrorBoundary` logic into `RecoverableErrorBoundary` and nested it as the first child of the updated `ErrorBoundary`.
  - This nesting allows us to know that if `RecoverableErrorBoundary` was unable to recover, that the error is "Unrecoverable" and needs to be handled.

#### Release Notes
  
- Updated ErrorBoundary to better handle Unrecoverable Errors that might be thrown during the render cycle.
- Deprecated ErrorBoundaryMixin as it only handles errors that are able to be "recovered" and recommend using ErrorBoundary with your custom props instead.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/component/error_boundary.dart
+++ b/lib/src/component/error_boundary.dart
@@ -57,14 +57,16 @@ class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBounda
   //
   //     [2.1] If we have `props.fallbackUIRenderer` we present that.
   //
-  //     [2.2] Present nothing instead of the broken children.
+  //     [2.2] Due to this ErrorBoundary only catching during "unrecoverable" render cycle errors
+  //           there is never any DOM created that can be used, and because `props.fallbackUIRenderer`
+  //           is not set we have no content to display, so we present an empty div instead.
   // ---------------------------------------------- /\ ----------------------------------------------
 
   ReactErrorInfo _errorInfo;
   dynamic _error;
 
   @override
-  Map get defaultProps => (newProps()
+  get defaultProps => (newProps()
     ..identicalErrorFrequencyTolerance = Duration(seconds: 5)
     ..loggerName = defaultErrorBoundaryLoggerName
     ..shouldLogErrors = true
@@ -110,7 +112,7 @@ class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBounda
         )(); // [2.2]
       }
     }
-    return (RecoverableErrorBoundary()..addProps(props))(props.children); // [1]
+    return (RecoverableErrorBoundary()..modifyProps(addUnconsumedProps))(props.children); // [1]
   }
 
   @override
@@ -120,6 +122,7 @@ class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBounda
     if (state.hasError) {
       final childThatCausedError = typedPropsFactory(prevProps).children.single;
       if (childThatCausedError != props.children.single) {
+        print('children are different!');
         reset();
       }
     }
--- a/lib/src/component/error_boundary.dart
+++ b/lib/src/component/error_boundary.dart
@@ -1,6 +1,7 @@
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart';
+import 'package:over_react/src/component/error_boundary_mixins.dart';
 import 'package:over_react/src/component/error_boundary_recoverable.dart';
 
 part 'error_boundary.over_react.g.dart';
@@ -25,7 +26,7 @@ class _$ErrorBoundaryState extends UiState with ErrorBoundaryStateMixin {}
 
 @Component2(isWrapper: true, isErrorBoundary: true)
 class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBoundaryState>
-    extends UiStatefulComponent2<T, S> {
+    extends UiStatefulComponent2<T, S> with ErrorBoundaryApi<T, S> {
 
   // ---------------------------------------------- \/ ----------------------------------------------
   // How This ErrorBoundary Works:
--- a/lib/src/component/error_boundary.dart
+++ b/lib/src/component/error_boundary.dart
@@ -1,4 +1,6 @@
+import 'package:meta/meta.dart';
 import 'package:over_react/over_react.dart';
+import 'package:over_react/src/component/error_boundary_recoverable.dart';
 
 part 'error_boundary.over_react.g.dart';
 
@@ -22,4 +24,35 @@ class _$ErrorBoundaryState extends UiState with ErrorBoundaryStateMixin {}
 
 @Component2(isWrapper: true, isErrorBoundary: true)
 class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBoundaryState>
-    extends UiStatefulComponent2<T, S> with ErrorBoundaryMixin<T, S> {}
+    extends UiStatefulComponent2<T, S> with ErrorBoundaryMixin<T, S> {
+
+  @override
+  Map getDerivedStateFromError(error) {
+    return (newState()
+      ..hasError = true
+      ..showFallbackUIOnError = true
+    );
+  }
+
+  @override
+  void componentDidCatch(error, ReactErrorInfo info) {
+    if (props.onComponentDidCatch != null) {
+      props.onComponentDidCatch(error, info);
+    }
+
+    if (props.onComponentIsUnrecoverable != null) {
+      props.onComponentIsUnrecoverable(error, info);
+    }
+  }
+
+  @override
+  void componentDidUpdate(Map prevProps, Map prevState, [dynamic snapshot]) {}
+
+  @override
+  render() {
+    if (state.hasError) {
+      return super.render();
+    }
+    return (RecoverableErrorBoundary()..addProps(props))(props.children);
+  }
+}
--- a/lib/src/component/error_boundary.dart
+++ b/lib/src/component/error_boundary.dart
@@ -1,4 +1,4 @@
-import 'package:meta/meta.dart';
+import 'package:logging/logging.dart';
 import 'package:over_react/over_react.dart';
 import 'package:over_react/src/component/error_boundary_recoverable.dart';
 
@@ -24,10 +24,60 @@ class _$ErrorBoundaryState extends UiState with ErrorBoundaryStateMixin {}
 
 @Component2(isWrapper: true, isErrorBoundary: true)
 class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBoundaryState>
-    extends UiStatefulComponent2<T, S> with ErrorBoundaryMixin<T, S> {
+    extends UiStatefulComponent2<T, S> {
+
+  // ---------------------------------------------- \/ ----------------------------------------------
+  // How This ErrorBoundary Works:
+  //
+  //   Background Info:
+  //     React gives each Error Boundary one chance to handle or recover when an error is throws, if
+  //     the Error Boundary's children throw again on its next render after React calls
+  //     `getDerivedStateFromError` and `componentDidCatch`, React will ascend the tree to the next
+  //     Error Boundary above it and rerender offering it a chance to handle the error. If none of
+  //     the parent Error Boundaries have a successful render cycle, React unmounts the entire tree.
+  //       __Note:__ When an Error Boundary remounts its children React clears all child components
+  //       previous state, including child ErrorBoundaries meaning they lose all previous knowledge
+  //       of any errors thrown.
+  //
+  //   Solution:
+  //     To prevent unmounting the entire tree when React cannot find an Error Boundary that is able
+  //     to handle the error we wrap an Error Boundary with another Error Boundary (this one!). The
+  //     child Error Boundary will handle errors that are "recoverable", so if an error gets to this
+  //     Error Boundary we know it is "unrecoverable" and can present a fallback.
+  //
+  // -----------------------------------------------------------------------------------------------
+  // Implementation:
+  //
+  // [1] Renders a child Error Boundary that is able to handle Errors thrown outside of the initial
+  //     render cycle, allowing it a chance to "recover".
+  //
+  // [2] If we catch an error in this Error Boundary that indicates that the child Error Boundary was
+  //     unable to handle or recover from the error, so we know that it was "unrecoverable" and we
+  //     haven't had a successful render.
+  //
+  //     [2.1] If we have `props.fallbackUIRenderer` we present that.
+  //
+  //     [2.2] Present nothing instead of the broken children.
+  // ---------------------------------------------- /\ ----------------------------------------------
+
+  ReactErrorInfo _errorInfo;
+  String _error;
+
+  @override
+  Map get defaultProps => (newProps()
+    ..identicalErrorFrequencyTolerance = Duration(seconds: 5)
+    ..loggerName = defaultErrorBoundaryLoggerName
+    ..shouldLogErrors = true
+  );
+
+  @override
+  get initialState => newState()
+    ..hasError = false
+    ..showFallbackUIOnError = true;
 
   @override
   Map getDerivedStateFromError(error) {
+    _error = error;
     return (newState()
       ..hasError = true
       ..showFallbackUIOnError = true
@@ -36,9 +86,14 @@ class ErrorBoundaryComponent<T extends ErrorBoundaryProps, S extends ErrorBounda
 
   @override
   void componentDidCatch(error, ReactErrorInfo info) {
+    _error = error;
+    _errorInfo = info;
+
     if (props.onComponentDidCatch != null) {
       props.onComponentDidCatch(error, info);
     }
+
+    _logErrorCaughtByErrorBoundary(error, info);
 
     if (props.onComponentIsUnrecoverable != null) {
       props.onComponentIsUnrecoverable(error, info);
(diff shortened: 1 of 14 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/component/error_boundary.dart
```diff
@@ -22,4 +24,130 @@ class _$ErrorBoundaryState extends UiState with ErrorBoundaryStateMixin {}
...
+      if (props.fallbackUIRenderer != null) {
+        return props.fallbackUIRenderer(_error, _errorInfo); // [2.1]
+      } else {
+        return (Dom.div()
+          ..key = 'ohnoes'
+          ..addTestId('ErrorBoundary.unrecoverableErrorInnerHtmlContainerNode')
+        )(); // [2.2]
+      }
+    }
+    return (RecoverableErrorBoundary()..addProps(props))(props.children); // [1]
```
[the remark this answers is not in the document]
kealjones-wk: but i want all the props this component gets to go to the child as well, doesn't `addUnconsumedProps` remove basically all the props that this one uses... meaning that the child would get nothing?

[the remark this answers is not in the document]
aaronlademann-wf: If they share a props namespace (in this case, `ErrorBoundaryPropsMixin.`) - they should get forwarded.

lib/src/component/error_boundary.dart
```diff
@@ -22,4 +24,130 @@ class _$ErrorBoundaryState extends UiState with ErrorBoundaryStateMixin {}
...
+  }
+
+  @override
+  void componentDidUpdate(Map prevProps, Map prevState, [dynamic snapshot]) {
+    // If the child is different, and the error boundary is currently in an error state,
+    // give the children a chance to mount.
+    if (state.hasError) {
+      final childThatCausedError = typedPropsFactory(prevProps).children.single;
+      if (childThatCausedError != props.children.single) {
```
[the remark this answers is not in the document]
kealjones-wk: this reset logic was actually taken from the original error boundary, i thought it was a way that once there was an error, that if the children passed into the errorboundary were different, it trys again, if we didnt have this logic i think it wouldnt re-render due to its state error. ill add some tests for this if there arent already.

lib/src/component/error_boundary.dart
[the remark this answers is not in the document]
kealjones-wk: it totally does :)

lib/src/component/error_boundary.dart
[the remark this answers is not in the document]
kealjones-wk: Double checked this logic and tested this manually by updating other props and not the children. With this logic if its the same children (aka the one that caused the error) it will NOT re-render (thats good its what we expect and want), but if the children are different this logic resets the state and allows the new child to be rendered, this is good.

One thing to consider though is that we wrapped around components in wdesk like so:
```
ErrorBoundary()(
  SidePanel()()
);
```
that means that that child probably won't change even if the props to it do, but if we were to instead wrap the error boundary around the contents of these components `render` method and new content passed into them would reset the error boundary... 

@aaronlademann-wf @greglittlefield-wf @sydneyjodon-wk @joebingham-wk  thoughts?

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -164,16 +184,20 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
           final component = jacket.getDartInstance();
           component.setState(component.newState()
             ..hasError = true
-            ..showFallbackUIOnError = true
           );
           expect(getByTestId(jacket.getInstance(), 'dummyChild'), isNull);
-          expect(getByTestId(jacket.getInstance(), 'fallbackNode'), isNotNull);
+          if (!isWrapper) {
+            // wrapper ErrorBoundary doesn't use `props.fallbackUIRenderer` because `ResolvableErrorBoundary`
+            // will display it on first error. Meaning wrapper ErrorBoundary will never be reached if it is set.
```
kealjones-wk: The child error boundary catching and displaying the fallback renderer is tested here:
https://github.com/Workiva/over_react/pull/435/files#diff-89d97c9289557c366adc6964193a6057R372-R388

lib/src/component/error_boundary.dart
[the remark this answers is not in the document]
aaronlademann-wf: tldr; from offline convo:

This is not a concern for this PR.

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -164,16 +184,20 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
+            expect(getByTestId(jacket.getInstance(), 'fallbackNode'), isNotNull);
+          }
         });
 
         test('when reset() is called', () {
-          jacket.getDartInstance().reset();
-          expect(jacket.getDartInstance().state.hasError, isFalse);
-          expect(jacket.getDartInstance().state.showFallbackUIOnError, isTrue);
+          dynamic component = jacket.getDartInstance();
+          (component as dynamic).reset();
```
aaronlademann-wf: I don't think making this dynamic is necessary.  You just need to update line 15 to:

```dart
TestJacket<ErrorBoundaryMixin> jacket; // ignore: deprecated_member_use_from_same_package
```

and then this can change to:

```dart
final component = jacket.getDartInstance();
component.reset();
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -304,6 +336,72 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
         _setCallbackVarValues();
       }
 
+      if (isWrapper) {
+        group('and the errors are during the render cycle', () {
+          group('and they occurred on every render', () {
+            setUp(() async {
```
aaronlademann-wf:
```suggestion
            setUp(() {
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -304,6 +336,72 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
+              expect(errorSentToComponentDidCatchCallback, errorSentToComponentIsUnrecoverableCallback);
+
+              expect(errorInfoSentToComponentDidCatchCallback, isA<ReactErrorInfo>());
+              expect(errorInfoSentToComponentIsUnrecoverableCallback, isA<ReactErrorInfo>());
+              expect(errorInfoSentToComponentDidCatchCallback, errorInfoSentToComponentIsUnrecoverableCallback);
+            });
+          });
+
+          group('and they occurred on every render and fallbackUIRenderer is set', () {
+            setUp(() async {
```
aaronlademann-wf:
```suggestion
            setUp(() {
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -304,6 +336,72 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
+              expect(jacket.getDartInstance().state.showFallbackUIOnError, isTrue,
+                  reason: 'Fallback UI should be rendered instead of the flawed component tree');
+              expect(jacket.getNode(),
+                  hasAttr(defaultTestIdKey, 'fallbackUIRenderer'));
+              calls.clear();
+            });
+
+            test('the correct callbacks are called with the correct arguments', () {
+              expect(calls.any((call) => call.keys.single == 'onComponentDidCatch'), isTrue,
+                  reason: 'onComponentDidCatch should have been called');
```
aaronlademann-wf: It seems like the primary difference between this _"and they occurred on every render and fallbackUIRenderer is set"_ block - and the previous _"and they occurred on every render"_ block - is that the latter should not ever get to the "unrecoverable" state, correct?  If that is the case - I believe an assertion needs to be made that `props. onComponentIsUnrecoverable` is never called.

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -450,8 +548,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
           test('the components wrapped by the ErrorBoundary get remounted', () {
             expect(getByTestId(jacket.getInstance(), 'flawedComponent'), isNotNull,
                 reason: 'The flawed component should have been remounted');
-            expect(jacket.getDartInstance().state.showFallbackUIOnError, isFalse,
-                reason: 'Fallback UI should be not be rendered');
+            if (!isWrapper) {
+              expect(jacket.getDartInstance().state.showFallbackUIOnError, isFalse,
+                  reason: 'Fallback UI should be not be rendered');
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -558,8 +658,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -626,8 +728,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -659,8 +763,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -698,8 +804,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
           test('the components wrapped by the ErrorBoundary get remounted', () {
             expect(getByTestId(jacket.getInstance(), 'nestedFlawedComponent'), isNotNull,
                 reason: 'The flawed component should have been remounted');
-            expect(jacket.getDartInstance().state.showFallbackUIOnError, isFalse,
-                reason: 'Fallback UI should be not be rendered');
+            if (!isWrapper) {
+              expect(jacket.getDartInstance().state.showFallbackUIOnError, isFalse,
+                  reason: 'Fallback UI should be not be rendered');
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -731,8 +839,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -766,8 +876,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -799,8 +911,10 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
```
aaronlademann-wf:
```suggestion
                  reason: 'Fallback UI should not be rendered');
```

web/src/demos/faulty-on-mount-component.dart
```diff
@@ -0,0 +1,31 @@
...
+class FaultyOnMountProps extends _$FaultyOnMountProps with _$FaultyOnMountPropsAccessorsMixin {
+  // ignore: undefined_identifier, undefined_class, const_initialized_with_non_constant_value
+  static const PropsMeta meta = _$metaForFaultyOnMountProps;
+}
+
+@Component2()
+class FaultyOnMountComponent extends UiComponent2<FaultyOnMountProps> {
+  @override
+  render() {
+    throw Shade();
```
aaronlademann-wf: you win the internet today Keal

test/over_react/component/shared_error_boundary_tests.dart
kealjones-wk: The usage of `latter` is confusing because i am interpreting it as you are referring to the "latter" in the actually testing order not in the sentence you used. which would be the former? lol ugh, either way i think i understood and added an expect that its not there. you let me know if i misunderstood.

test/over_react/component/shared_error_boundary_tests.dart
kealjones-wk: i cannot because ErrorBoundary no longer has anything to do with ErrorBoundaryMixin.

lib/src/component/error_boundary_mixins.dart
```diff
@@ -13,6 +13,8 @@ const String defaultErrorBoundaryLoggerName = 'over_react.ErrorBoundary';
 /// within a custom component.
 ///
 /// > See: [ErrorBoundaryMixin] for a usage example.
+@Deprecated('Building custom error boundaries with this mixin will no longer be supported in version 4.0.0.'
+    'Use ErrorBoundary and its prop API to customize error handling instead.')
```
greglittlefield-wf: NICE

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -226,17 +252,22 @@ void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
...
       final componentDidCatchCallbackArguments = calls[0]['onComponentDidCatch'];
       if (componentDidCatchCallbackArguments != null) {
         errorSentToComponentDidCatchCallback = componentDidCatchCallbackArguments[0];
         errorInfoSentToComponentDidCatchCallback = componentDidCatchCallbackArguments[1];
       }
 
+      // At this point the component is "Unrecoverable"
       if (calls.length > 1) {
-        expect(calls[1].keys.single, 'onComponentIsUnrecoverable', reason: 'test setup sanity check');
+        expect(calls[calls.length - 1].keys.single, 'onComponentIsUnrecoverable', reason: 'test setup sanity check');
```
greglittlefield-wf: #nit could use `.last`
```suggestion
        expect(calls.last.keys.single, 'onComponentIsUnrecoverable', reason: 'test setup sanity check');
```

test/over_react/component/shared_error_boundary_tests.dart
```diff
@@ -9,8 +9,10 @@ import './fixtures/flawed_component_on_mount.dart';
 import './fixtures/flawed_component_that_renders_a_string.dart';
 import './fixtures/flawed_component_that_renders_nothing.dart';
 
-void sharedErrorBoundaryTests(BuilderOnlyUiFactory builder) {
-  TestJacket<ErrorBoundaryMixin> jacket;
+/// `isWrapper` identifies an ErrorBoundary that wraps another Error Boundary in order to handle
```
greglittlefield-wf: #nit this should be a doc comment reference
```suggestion
/// [isWrapper] identifies an ErrorBoundary that wraps another Error Boundary in order to handle
```

test/over_react/component/shared_error_boundary_tests.dart
aaronlademann-wf: k… then lets moved the shared public method in an interface like so:

mixin ErrorBoundaryApi on UiStatefulComponent2 {
  /// Resets the [ErrorBoundary] to a non-error state.
  ///
  /// This can be called manually on the component instance using a `ref` -
  /// or by passing in a new child instance after a child has thrown an error.
  void reset() {
    setState(initialState);
  }
}

The interface should not be exported publicly, but it can then be mixed in to both of the error boundary components.

test/over_react/component/shared_error_boundary_tests.dart
aaronlademann-wf: mm yeah - i can see how that was confusing... i think you got it though.

test/over_react/component/shared_error_boundary_tests.dart
kealjones-wk: well la-di-da mr dart is awesome... back in my day when javascript ruled the world we didnt have the luxury of your fancy smanchy `.last` you had to put in the effort and really think, i need the last item in this array so its length... but its indexed from 0 so i need to subtract 1... thats real dev...

kealjones-wk:
> +1
> 
> Nothin but nits

swish!

evanweible-wf:
Security +1
- Watched term `InnerHtml` was flagged, but it's actually just a part of a test ID.

kealjones-wk:
@Workiva/release-management-p

PR #438, opened 2019-12-10
CPLAT-8775 Flux to Redux Examples

joebingham-wk:
## Motivation
  
Provide examples of what it looks like to go from Flux to Redux.
## Changes
  
- Add simple and advanced examples in `web`
#### Release Notes
  
- Add Flux to Redux transition examples.
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary2-wf:
## Security Insights

## Action Items
---

doc/flux_to_redux.md
```diff
@@ -0,0 +1,134 @@
...
+- Takes extra time as there are ultimately two refactors instead of one (however minimal the second one will be). While ideally the second refactor (from Influx to Redux) should be an easy lift, it's still ultimately unnecessary.
+- Does not provide any performance gains. The Redux `connected` components will update on every store update as long as Flux is in the mix. This mimics the behavior of Flux, but takes away the largest benefit of connecting to Redux.
+- Can add increased complexity and code to a possibly already complex state architecture. There is additional boilerplate and utilities necessary to maintain an Influx architecture that are completely unnecessary to Redux, making Influx more verbose and a little more confusing.
+
+In summary, it never makes sense to add steps and refactoring if you can afford not to. The primary advantage in Influx is that you can break up the project. If that may become important or is enticing, it is an option!
+
+## Example Explained
+This section goes over the example in `/web/flux_to_redux/advanced/` and includes the goal of each phase and the steps to go from the last phase (starting with Flux) to the current phase.
```
greglittlefield-wf: I think it'd be helpful to also clarify that there's also a simpler example and link to it for users that have simpler use cases, so they don't have to read through all of this when they don't need to.

doc/flux_to_redux.md
```diff
@@ -0,0 +1,134 @@
...
+    * [Getting to Influx Architecture](#getting-to-influx-architecture)
+        * [Influx Components](#influx-components)
+        * [Influx Stores](#influx-stores)
+        * [Example Specific Explanations](#example-specific-explanations)
+    * [Getting to Redux Architecture](#getting-to-redux-architecture)
+        * [Redux Stores](#redux-stores)
+        * [Redux Components](#redux-components)
+
+## Goal
+The goal of this document is explain major elements of transitioning to Flux from Redux. This includes the introduction of an Influx architecture (which means your state management system will be... in flux), a simple and advanced transition example, and specific steps to navigate the transition.
```
kealjones-wk: uhh-no
```suggestion
The goal of this document is explain major elements of transitioning from Flux to Redux. This includes the introduction of an Influx architecture (which means your state management system will be... in flux), a simple and advanced transition example, and specific steps to navigate the transition.
```

joebingham-wk: This is actually a doc to get people _to_ Flux, Keal :trollface:

kealjones-wk: then you have a hell of alot MORE issue in other places, Joe :trollface:

joebingham-wk: Ah shoot.. yeah, you're right. I guess I'll go and fix everything to go from redux _to_ flux..

kealjones-wk: lol im sitting here looking at all gregs feedback and the fact that the first sentence had that issue...
see-but-dont-observe
lulz jk @greglittlefield-wf i love you

joebingham-wk: Hahahah this is why multiple reviewers are good

trentgrover-wf:
Sorry to say, but I find the text a bit difficult to follow and am afraid that a developer who is only familiar with their existing Flux architecture / terminology will get lost pretty quickly. I would suggest a structure that follows a pattern something like this:
- Start with the simple example
  - Short description of the simple Flux example (reader should be familiar with this and grok it easily)
  - Explanation of what the same simple example would ideally look like in Redux, introducing the new terms and noting important differences in structure
  - If the reader's Flux usage is pretty simple, ^^^ that might be all they need so they can stop reading and do a wholesale replacement, which would be the best case scenario for both us and them
- Then move to the more advanced example that includes multiple / nested stores
  - Short description of what it looks like in Flux
  - Same advanced example in Redux, noting important differences in structure / approach (e.g. flattened stores, connected components, etc.)
  - At this point the reader hopefully understands how their Flux mental model maps to Redux, though they might still be concerned about the scale of a wholesale migration of pattern
- Then bring in the Influx migration option
  - It's probably ok to jump right to the advanced example at this point
  - Using this is really the ugly option so it should be brought up as a last resort

A couple of other notes:
- The sample code will be easier to grok if all of the related code is colocated logically like it would be in a real project, even though that means there will be repetition in the example code (right now all the stores are in 1 shared file location for multiple examples).
- Some of the steps list names of particular utility components / classes that should be used without providing context as to why or for what purpose (e.g. `Wrap every Flux store in a FluxToReduxAdapterStore.` Why? What does it do?)

Thanks for putting this together. Education is 1 of the most important things we can provide and it takes a ton of effort and practice to do it well.

joebingham-wk:
Thanks for taking the time to review it @trentgrover-wf! I appreciate the feedback and think your points are very valid

trentgrover-wf:
On the terminology end of things, if there's a good quick primer that's available elsewhere to point people to, it's perfectly acceptable to insert 'Go read this first and then we'll continue'

doc/flux_to_redux.md
```diff
@@ -0,0 +1,257 @@
...
+    * [A Basic Flux App](#a-basic-flux-app)
+    * [The Redux Version](#the-redux-version)
+    * [Basic Example: The Main Differences](#basic-example:-the-main-differences)
+    * [Basic Conversion Step by Step](#basic-conversion-step-by-step)
+* __[An Advanced Example](#an-advanced-example)__
+    * [An Advanced Flux App](#an-advanced-flux-app)
+    * [Converted to Redux](#converted-to-redux)
+    * [Advanced Conversion Step by Step](#advanced-conversion-step-by-step)
+    * [What if the Advanced Conversion is too Daunting?](#what-if-the-advanced-conversion-is-too-daunting?)
+* __[Influx Architecture](#influx-architecture)__
```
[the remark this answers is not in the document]
joebingham-wk: I definitely see your point, and I could go either way. Do you feel strongly? I lean towards keeping Influx because I feel like it is actually pretty descriptive to the situation of the code and isn't a lot to comprehend for the few that will need to worry about it.

doc/flux_to_redux.md
```diff
@@ -0,0 +1,257 @@
...
+
+> NOTE: This section does not cover the complexities of having multiple stores, which presents more challenges than a single store. For a more complex example [see the advanced example below](#an-advanced-example).
+
+### A Basic Flux App
+For our example, here are our different items of interest:
+- __The Flux Store Class:__ we have a Flux store called `RandomColorStore` that is responsible for diplaying a random background color. The only state field is `backgroundColor`. `RandomColorStore` also has a `_changeBackgroundColor` function, and when called it sets `backgroundColor` to a new random color. 
+- __The Actions Class:__ There is an Actions class that has a single action: `RandomColorActions`. It only holds one action (`changeBackgroundColor`), which triggers `_changeBackgroundColor` in the store. 
+- __The Action Object:__ we create an object `randomColorActions` that is an intantiated instance of `RandomColorActions`.
```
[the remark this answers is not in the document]
joebingham-wk: Good call!

doc/flux_to_redux.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Basic Example: The Main Differences
+- __We replace the Flux store with both a state class and a reducer.__ In Redux, we do not define a class that is responsible for handling state updates. Rather, we create the model to represent the state and a function to describe updates. Then, use them in the instantiation of a Redux `Store` object.
+- __Actions are not held in an overarching wrapper class.__ Typically in Redux, actions "stand alone" from each other. Our reducer should have cases for all the actions that exist, but unlike Flux there is no need to knit our store and actions together by actually providing an actions instance. Consequently, actions are broken away, pulled directly into the component, and triggered using `dispatch`.
+- __In Redux, actions and state are not passed directly as props from a parent component.__ In actuality, that statement is more nuanced and has some hidden complexity. On the surface though, Redux UI uses context and the `connect` function to associate state directly with props without needing to actively pass the store from parent to child all the way through a component tree.
```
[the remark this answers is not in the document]
joebingham-wk: Sounds good to me!

lib/src/over_react_redux/over_react_flux.dart
```diff
@@ -28,9 +28,9 @@ class FluxToReduxAdapterStore<S extends flux.Store> extends redux.Store<S> {
 
   StreamSubscription _storeListener;
 
-  FluxToReduxAdapterStore(this.store, dynamic actions) : super(_noopReducer, initialState: store, distinct: false) {
+  FluxToReduxAdapterStore(this.store, dynamic actions, {Function reducer}) : super(reducer ?? _noopReducer, initialState: store, distinct: false) {
```
kealjones-wk: We should add a middleware option, as one of the biggest benefits of moving to redux allowing them to use the `overReactReduxDevToolsMiddleware` while transitioning seems like an added benefit, also, we should type `reducer` a bit more strictly than just `Function`
```suggestion
  FluxToReduxAdapterStore(this.store, dynamic actions, {{List<redux.Middleware<S>>middleware, redux.Reducer<S> reducer}) : super(reducer ?? _noopReducer, initialState: store, middleware: middleware, distinct: false) {
```

web/flux_to_redux/simple/examples/influx_implementation/components/connect_flux_big_block.dart
```diff
@@ -0,0 +1,40 @@
...
+
+import '../store.dart';
+
+part 'connect_flux_big_block.over_react.g.dart';
+
+UiFactory<ConnectFluxBigBlockProps> ConnectedConnectFluxBigBlock =
+    connectFlux<FluxStore, RandomColorActions, ConnectFluxBigBlockProps>(
+  mapStateToProps: (state) => (ConnectFluxBigBlock()..backgroundColor = state?.backgroundColor),
+  mapActionsToProps: (actions) =>
+      (ConnectFluxBigBlock()..changeBackgroundColor = () => actions.changeBackgroundColor()),
```
kealjones-wk:
```suggestion
      (ConnectFluxBigBlock()..changeBackgroundColor = actions.changeBackgroundColor),
```

when you do it as an inline function it will be a "new" instance every time and thus trigger a re-render even if props are identical.

kealjones-wk: With Inline Function:
Screen Shot 2020-01-02 at 2 32 54 PM

Without inline function:
Screen Shot 2020-01-02 at 2 32 18 PM

doc/flux_to_redux.md
```diff
@@ -0,0 +1,257 @@
...
+
+### Disadvantages
+The biggest disadvantages are that it:
+- __Takes extra time as there are ultimately two refactors instead of one (however minimal the second one will be).__ While ideally the second refactor (from Influx to Redux) should be an easy lift, it's still ultimately unnecessary.
+- __Does not provide any performance gains until the transition is complete.__ The Redux `connected` components will update on every store update as long as Flux is in the mix. This mimics the behavior of Flux, but takes away the largest benefit of connecting to Redux.
+- __Can add increased complexity and code to a possibly already complex state architecture.__ There is additional boilerplate and utilities necessary to maintain an Influx architecture that are completely unnecessary to Redux, making Influx more verbose and a little more confusing.
+
+In summary, it never makes sense to add steps and refactoring if you can afford not to. The primary advantage in Influx is that you can break up the project. If that may become important or is enticing, it is an option!
```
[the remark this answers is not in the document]
joebingham-wk: Sounds good to me!

lib/src/over_react_redux/over_react_flux.dart
joebingham-wk: Good catch!!

web/flux_to_redux/simple/examples/influx_implementation/components/redux_big_block.dart
```diff
@@ -0,0 +1,39 @@
...
+
+import '../store.dart';
+
+part 'redux_big_block.over_react.g.dart';
+
+UiFactory<ReduxBigBlockProps> ConnectedReduxBigBlock = connect<FluxStore, ReduxBigBlockProps>(
+  mapStateToProps: (state) => (ReduxBigBlock()..backgroundColor = state.backgroundColor),
+  mapDispatchToProps: (dispatch) =>
+      (ReduxBigBlock()..changeBackgroundColor = () => dispatch(UpdateBackgroundColorAction())),
+  pure: false,
```
kealjones-wk: @greglittlefield-wf @joebingham-wk 
So after a little testing, if `pure: false` is used the redux component will always re-render on any state change (similar to flux) BUT if you remove this arg and instead use `areStatesEqual: (_, __) => false,` it will not re-render on state changes that do not affect its props:

influx_redux_pure

doc/flux_to_redux.md
```diff
@@ -0,0 +1,257 @@
...
+
+> NOTE: This section does not cover the complexities of having multiple stores, which presents more challenges than a single store. For a more complex example [see the advanced example below](#an-advanced-example).
+
+### A Basic Flux App
+For our example, here are our different items of interest:
+- __The Flux Store Class:__ we have a Flux store called `RandomColorStore` that is responsible for diplaying a random background color. The only state field is `backgroundColor`. `RandomColorStore` also has a `_changeBackgroundColor` function, and when called it sets `backgroundColor` to a new random color. 
+- __The Actions Class:__ There is an Actions class that has a single action: `RandomColorActions`. It only holds one action (`changeBackgroundColor`), which triggers `_changeBackgroundColor` in the store. 
+- __The Action Object:__ we create an object `randomColorActions` that is an intantiated instance of `RandomColorActions`.
+- __The Store Object:__ Our store, called `randomColorStore`, is instantiated with `randomColorActions` being passed in.
```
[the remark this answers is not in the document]
joebingham-wk: Yep, I like it and see what you mean. I originally thought that if they wanted to see code, they could just hop over to the example to see it, but I agree in retrospect that having the isolated snippets as examples here makes it a lot more clear!

lib/src/over_react_redux/over_react_flux.dart
kealjones-wk: forgot to add ` = const []` woops should be `List<redux.Middleware<S>>middleware = const []`

joebingham-wk: Hahaha don't worry, I ran into it and it was pretty easy to see what was going on

web/flux_to_redux/simple/examples/influx_implementation/components/redux_big_block.dart
joebingham-wk: I must be missing some code somewhere - would you mind pushing that change up to a branch? When I make the parameter change and update the component code to look like yours, my profiler still shows the component rendering

kealjones-wk: oh i also removed the dispatch to props... because similar to what i mentioned for the flux action to props stuff, same thing applied here... it seems problematic to me and im not sure how to go about even using dispatchToProps too use the same instance.

joebingham-wk: I removed that as well and am still not seeing what you are :/

kealjones-wk: I added an additional prop to the middle component that toggles a state bool field and did not include that in the mapStateToProps for the redux component.

joebingham-wk: Ah yep, that would do it!

kealjones-wk: @greglittlefield-wf

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+    // After
+    class SetBackgroundColorAction {
+        final String backgroundColor;
+        SetBackgroundColorAction(this.backgroundColor);
+    }
+    ```
+
+1. __Pull state mutation logic out of the store and into a reducer.__ Typically within a Flux store you have a `triggerOnActionV2` call that identifies an action and a function used to respond to that action. In many cases, that same logic should be perfect for a reducer.
+    ```dart
+    // Before transition
```
aaronlademann-wf: I think we should show the necessary imports for clarity as to what the `flux` namespace represents.

```suggestion
    import 'package:w_flux/w_flux.dart' as flux;

    // Before transition
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+            final color = '#' + (Random().nextDouble() * 16777215).floor().toRadixString(16);
+            return ExampleState.update(color: color);
+        }
+
+        return oldState;
+    }
+    ```
+1. __Create the state model.__ After pulling out state mutation logic, this should be as simple as renaming the class, not inheriting from the Flux store class, and cleaning up anything unrelated to the actual state fields. 
+1. __Instantiate the store class, using the state model and reducer.__
+1. __Add a `ReduxProvider` around the component tree.__ This will take in the store instantiated in step 4 as the component's `store` prop.
```
aaronlademann-wf: Out of curiosity - why do none of these examples show any example code for the UI bits? (e.g. showing the actual wrapping of the component)

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+    - props.littlebig_Store.state
+
+    Our `BigBlock` component renders three `SmallBlock` components. Each one connects to a different store's state, with the remaining background color being tied to `BigBlock` itself. 
+    
+    `BigBlock` also has four buttons - one connected to each action that changes the background color property of one of the stores (and thereby updating the component background color).
+
+    A Flux specific part of the `BigBlock` component is that we need to set the `redrawOne` to listen to the story at every level.
+
+```dart
+// A nested store
```
aaronlademann-wf: I think we should show the necessary imports for clarity as to what the `flux` namespace represents.

```suggestion
import 'package:w_flux/w_flux.dart' as flux;
    
// A nested store
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+    }
+
+    _changeBackgroundColor(String _) {
+        _backgroundColor = '#' + (Random().nextDouble() * 16777215).floor().toRadixString(16);
+    }
+}
+```
+
+```dart
+// A shared action class for all the stores, with an action to update each background color
```
aaronlademann-wf: I think we should show the necessary imports for clarity as to what the `flux` namespace represents.

```suggestion
import 'package:w_flux/w_flux.dart' as flux;
    
// A shared action class for all the stores, with an action to update each background color
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+
+    return ReduxState.update(oldState, blockThreeBackgroundColor: color);
+  }
+
+  return oldState;
+}
+```
+
+```dart
+// Instantiate a store using the reducer and state class
```
aaronlademann-wf: I think we should show the necessary imports for clarity as to what the `redux` namespace represents.

```suggestion
import 'package:redux/redux.dart' as redux;
    
// Instantiate a store using the reducer and state class
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+
+    Despite there being four stores, all of them have the same job: handle a single random color property. Therefore, they are all nearly identical objects. Naturally in the real world they would likely be drastically different, but for simplicity sake and keeping the focus on the big picture they are all left the same.
+- __The Store Objects:__ `bigStore` and `littleStore`
+- __The Actions Class__, `RandomColorActions`
+- __The Actions Object__, `randomColorActions`
+- __The UI:__ The only complex component is a `BigBlock` component that receives `randomColorActions`, `bigStore`, and `littleStore` as props. Updates are triggered by using the action prop. Naturally now state is accessed via:
+    - props.bigStore.state
+    - props.bigStore.midLevelStore.state
+    - props.bigStore.midLevelStore.lowLevelStore.state
+    - props.littlebig_Store.state
```
aaronlademann-wf: This seems like a typo

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+1. __For each group, refactor all the stores.__ By doing the stores and reducers at the same time, you can completely invalidate a single store at a time. In other words, you can move all the necessary state logic into the new state class and move the state mutation logic into the reducer. Then, the entire Flux store can be deleted.
+
+    If you have already created a state class and a reducer for an earlier store group, continue adding to the same reducer and state class. Worth noting is that these app examples use a very simplistic reducer approach, and understanding the options with reducers may be worthwhile. Because there are multiple ways to implement reducers, it is likely one will make more sense than others.
+1. __Create any new action classes that are necessary.__ After refactoring a Flux store, it should be clear which actions need to be converted to Redux actions.
+1. __Refactor components into connected components.__ Update the component to `UiComponent2`, removing all Flux boilerplate, and wrapping the already declared factory with the `connect` call.
```
aaronlademann-wf: Again - it seems odd that the UI portion of these refactors lacks any code examples of any kind...

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+
+### What if the Advanced Conversion is too Daunting?
+The advanced conversion might sound like a huge initiative with numerous unknowns and complexities. If that's the case the first question is whether Redux will offer the benefits to justify a large initiative. Our [OverReact Redux documentation](./over_react_redux_documentation.md) includes some of the benefits, and there are many [articles online](https://www.google.com/search?q=the+benefits+of+redux&oq=the+benefits+of+redux) that provide more examples. 
+
+If Redux makes sense but the conversion project seems extremely challenging, we have a middleground: introducing __Influx__.
+
+## Influx Architecture
+### What is Influx?
+Influx is just the term we're giving to an architecture that is both Redux and Flux at the same time. To aid in the transition from Flux to Redux, tools have been built to allow Flux and Connected (Redux) components to all talk to the same store. It's a transitional architecture that allows the library to be non-commital to the state management system, ultimately allowing the transition from Flux to Redux to be a much more incremental process.
```
aaronlademann-wf:
```suggestion
Influx is just the term we're giving to an architecture that is both Redux and Flux at the same time. To aid in the transition from Flux to Redux, we have built some utilities to allow Flux and Connected (Redux) components to all talk to the same store. It's a transitional architecture that allows the library to be non-commital to the state management system, ultimately allowing the transition from Flux to Redux to be a much more incremental process.
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+__tl;dr__
+
+__If you assess your architecture and are confident you can go straight from Flux to Redux, you should not use Influx. If you are not confident, you should consider it.__
+
+The Influx architecture is not a required part of the process, and may make your life more difficult by adding extra steps. There are both advantages and disadvantages, and assessing how large the refactor is will likely provide the largest indication of whether or not it makes sense.
+
+### Advantages
+- __You can split the effort into very tangible subtasks.__ Stores can be updated to Redux without necessarily updating the Flux components. Components reliant on multiple stores can talk to Flux stores and Redux stores at the same time. Ultimately, this means a library can pick and choose the areas that can be updated and do them one at a time.
+- __Inability to update entirely to `Component2` is not a blocker.__ Since `connected` components need to be `Component2`, the Redux refactor may be blocked by those efforts. This option allows the library to update the state system to a Redux friendly architecture while updating components to `UiComponent2` and `connect` at the same time.
```
aaronlademann-wf:
```suggestion
- __Inability to update entirely to `Component2` is not a blocker.__ Since `connected` components need to be `Component2`, the Redux refactor may be blocked by those efforts. This option allows the library to update the state system to a Redux friendly architecture while [updating components to `UiComponent2`](./ui_component2_transition.md) and `connect` at the same time.
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+- __Does not provide any performance gains until the transition is complete.__ The Redux `connected` components will update on every store update as long as Flux is in the mix. This mimics the behavior of Flux, but takes away the largest benefit of connecting to Redux.
+- __Can add complexity and code to a possibly already complex state architecture.__ There are additional boilerplate and utilities necessary to maintain an Influx architecture that are completely unnecessary to Redux, making Influx more verbose and a little more confusing.
+
+In summary, Influx adds steps to the transition process, but can make the work easier to break up.
+
+### Important Terms
+If this architecture is appealing, there are a few new classes and utilities it will be beneficial to be aware of.
+- __Adapted Influx Store:__ The instance returned from wrapping an Influx store with a `FluxToReduxAdapterStore`.
+- __`composeHocs`:__ If a component takes in multiple stores, it needs to be connected to all of them. This function that allows you to combine multiple `connect` or `connectFlux` calls using a flat list, as opposed to nesting them inside each other.
```
aaronlademann-wf:
```suggestion
- __`composeHocs`:__ If a component takes in multiple stores, it needs to be connected to all of them. This function allows you to combine multiple `connect` or `connectFlux` calls using a flat list, as opposed to nesting them inside each other.
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+- __`connectFlux`:__ Like [`connect`](https://github.com/Workiva/over_react/blob/master/doc/over_react_redux_documentation.md#connect), but for Flux stores instead of Redux stores. This is useful because it is one step closer to a Redux connected component without being Redux. If, for any reason, implementing the Redux side of Influx is presenting challenges, `connectFlux` provides a good middle ground.
```
aaronlademann-wf:
```suggestion
- __`connectFlux`:__ Like [`connect`](./over_react_redux_documentation.md#connect), but for Flux stores instead of Redux stores. This is useful because it is one step closer to a Redux connected component without being Redux. If, for any reason, implementing the Redux side of Influx is presenting challenges, `connectFlux` provides a good middle ground.
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+- __`FluxToReduxAdapterStore`:__ This is a class that wraps an Influx store and makes it look like a Redux store. It is the cornerstone of Influx because pure Flux components will stay connected to the original store instance, but Redux components and connected Flux components (using `connectFlux`) will connect to the instatiated `FluxToReduxAdapaterStore` object. This works by passing in an Influx store instance and a Flux actions instance.
```
aaronlademann-wf:
```suggestion
- __`FluxToReduxAdapterStore`:__ This is a class that wraps an Influx store and makes it look like a Redux store. It is the cornerstone of Influx because pure Flux components will stay connected to the original store instance, but Redux components and connected Flux components (using `connectFlux`) will connect to the instatiated `FluxToReduxAdapaterStore` object. This works by passing in an Influx store instance and a Flux `Actions` instance.
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+- __Influx store:__ A Flux store that has implemented the `InfluxStoreMixin` and converted its internal logic (constructor, getters, overrides) to match the Influx pattern.
+- __`InfluxStoreMixin`:__ A mixin that attaches to a Flux store in order to add necessary Influx utilities. Namely, the method `influxReducer` must be accessible on the Flux store class, and `influxReducer` expects the class to have a `reduxReducer` method and a `state` field. The `state` field is essentially a proxy for a Redux state class, and makes it so that state class can be built slowly over time without any need to refactor it again.
```
aaronlademann-wf:
```suggestion
- __`InfluxStoreMixin`:__ A mixin that attaches to a Flux store in order to add necessary Influx utilities. Namely, the method `influxReducer` must be accessible on the Flux `Store` class, and `influxReducer` expects the class to have a `reduxReducer` method and a `state` field. The `state` field is essentially a proxy for a Redux state class, and makes it so that state class can be built slowly over time without any need to refactor it again.
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+        1. __Override the `reduxReducer` getter on the class.__ The `reduxReducer` getter should point to the Redux reducer function created earlier.
+        1. __Update the store fields.__ The end result of this step is that the Influx store (previously the Flux store) should have a getter for every Redux state class field. The getters should point to corresponding values on the `state` field. 
+        
+        Updating getters isn't necessary as the correct fields should be accessible via `influxStoreInstance.state.reduxProperty`, but refactoring the getters means that any Flux components pointing to this store do not need to be refactored. If it is preferable just to remove getters in favor of accessing the `state` fields directly, any original getters or fields that correspond to a value moved to the Redux state class should be removed.
+
+        ```dart
+        // An Influx Store
```
aaronlademann-wf: I think we should show the necessary imports for clarity as to what the `flux` namespace represents, and where they can access stuff like `InfluxStoreMixin`.

```suggestion
        import 'package:over_react/over_react_redux.dart';
        import 'package:w_flux/w_flux.dart' as flux;

        // An Influx Store
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+        class ExampleFluxActions {
+            final flux.Action<String> exampleFluxAction = flux.Action();
+        }
+        ```
+1. __Add adapter stores where necessary.__ Each store that will be Influx should be wrapped with a `FluxToReduxAdapterStore` to give Flux and Connected Flux access as well. In other words, if the store was updated to Redux directly or will only talk to Redux components, this step can be skipped.
+    
+    `FluxToReduxAdapterStore` takes in an Influx store and an actions instance. It should also receive the Influx store class and the Redux state class as generics.
+    ``` dart
+    FluxToReduxAdapterStore influxAdapterStore =
+        FluxToReduxAdapterStore<ExampleInfluxStore, ReduxStateClass>(exampleInfluxStoreInstance, exampleFluxActionsInstance);
```
aaronlademann-wf:
```dart
final influxAdapterStore = FluxToReduxAdapterStore<ExampleInfluxStore, ReduxStateClass>(
    exampleInfluxStoreInstance, exampleFluxActionsInstance);
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+    ```
+1. __Create `Context` instances for the necessary stores.__ If a component will use multiple stores, a `Context` instance is required. The [OverReact Redux docs](./over_react_redux_documentation.md) have examples of having multiple stores, which describes the fundamentals of why this is necessary. In the end, it is most likely that the majority of `FluxToReduxAdapterStore` instances will need a `Context` instance. 
+1. __Wrap the component tree in a either a `ReduxProvider` or a `ReduxMultiProvider`.__ If there are multiple stores, a `ReduxMultiProvider` is more elegant and is encouraged. The store instances passed in should be the `FluxToReduxAdapterStore` instances, not the normal Flux store instances.
+1. __Refactor components.__ See the specific component type below for a reminder on which store instance is correct and any "gotchas" in the refactor. In general, remember that if a component talks to multiple stores, `composeHocs` can be used to simplify the connected factory declarations.
```
aaronlademann-wf: I think we need at least a couple of examples of a flux component being converted to a redux one.

aaronlademann-wf: I'll leave it up to you whether that gets put in its own dedicated section - and then just linked to from all the different places where we mention doing things like wrapping a component in a provider, etc. - or whether you add relevant examples in each place where mentioned.

doc/flux_to_redux.md
```diff
@@ -0,0 +1,579 @@
...
+
+If, through this process, the document or examples miss any edge cases you encounter, please create an issue or reach out so it can be added.
+
+## Do I have to transition to Redux?
+No. OverReact Redux is meant to help to provide a recommended state management method for OverReact components, as well as provide benefits over w_flux and built_redux. Those benefits may provide enough reason to make a refactor worthwhile. If, after evaluating the benefits and effort, the juice doesn't seem worth the squeeze, then no need to worry about it!
+
+To evaluate if the refactor is worth it, the details of OverReact Redux can be found in the [OverReact Redux doc](./over_react_redux_documentation.md). That document paired with this guide will illustrate the full scope of costs and benefits.
+
+## A Simple Example
+To illustrate the basic changes that will occur, this section will go through a basic Flux architecture and then show that same system with Redux instead. This section will also have a step by step list of instructions for an update as straightforward as this one. An actual working example can be found in the [Flux to Redux Example](../web/flux_to_redux/simple/readme.md).
```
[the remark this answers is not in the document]
joebingham-wk: This was originally what I wanted to do, but it seems like because that path doesn't have a specific endpoint, trying to use the link doesn't do anything. I actually created the readme just to add that endpoint so people could click on it. Should that path navigate me to the directory? I may be overlooking something

doc/flux_to_redux.md
joebingham-wk: Yep, good catch!

doc/flux_to_redux.md
joebingham-wk: All of the questions / points on adding UI examples are really good ones - my main reason was just that UI snippets were bulky and all of the code could be seen in the example. 

Adding the `ReduxProvider` wouldn't really be bulky though, so I added those examples!

For general "Refactoring UI Components", I just referred them to the examples in the web directory. I also added more comments in the Influx and Redux components to show what the refactor included. @aaronlademann-wf does that cover everything or do you think it'd be better in a markdown doc with the before and after examples together?

joebingham-wk:
@aviary-wf

doc/flux_to_redux.md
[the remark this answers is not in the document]
aaronlademann-wf: @joebingham-wk the link you've got now doesn't work at all (files on github are case sensitive).

Changing it to `../web/flux_to_redux/simple` navigates me correctly to https://github.com/Workiva/over_react/tree/0731dbb71e883a13670efb54297681d3e60230a1/web/flux_to_redux/simple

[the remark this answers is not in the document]
joebingham-wk: Ohh that makes more sense, @aaronlademann-wf. I had always used the IDE markdown previewer to view markdown docs (and do my link testing), so I wasn't even thinking of Github 🤦‍♂ Thank you for pointing that out - I updated references to point to the simple / advanced directory instead of the readme files

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+class RandomColorState {
+    // The state values the app needs
+    String backgroundColor;
+
+    // A constructor used to establish the default state
+    RandomColorState.defaultState() : this.backgroundColor = 'gray';
+
+    // A convenience constructor used by the reducer to return new instances of the class
+    RandomColorState.update(RandomColorState oldState, {String backgroundColor})
+        : this.backgroundColor = backgroundColor ?? oldState.backgroundColor;
```
trentgrover-wf: why isn't this just something like:
```
RandomColorState.withColor(String backgroundColor) : this.backgroundColor = backgroundColor;
```

I understand that if the internal logic somehow depended on the oldState, you'd want to include that, but this doesn't use it.

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+```
+
+```dart
+// A simple reducer that has a case for every action (in this case, a single one)
+RandomColorState reducer(RandomColorState oldState, dynamic action) {
+  if (action is UpdateBackgroundColorAction) {
+    var color = '#' + (Random().nextDouble() * 16777215).floor().toRadixString(16);
+
+    // Return a new state class instance, changing the background color
+    return RandomColorState.update(oldState, backgroundColor: color);
```
trentgrover-wf: why isn't this just something like:
```
return RandomColorState.withColor(color);
```

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+    - props.bigStore.state
+    - props.bigStore.midLevelStore.state
+    - props.bigStore.midLevelStore.lowLevelStore.state
+    - props.littleStore.state
+
+    Our `BigBlock` component renders three `SmallBlock` components. Each one connects to a different store's state, with the remaining background color being tied to `BigBlock` itself. 
+    
+    `BigBlock` also has four buttons - one connected to each action that changes the background color property of one of the stores (and thereby updating the component background color).
+
+    A Flux specific part of the `BigBlock` component is that we need to set the `redrawOne` to listen to the story at every level.
```
trentgrover-wf: `redrawOne` -> `redrawOn`
`story` -> `store`

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+
+    // A constructor for creating the default app state
+    ReduxState.defaultState()
+        : this.mainBackgroundColor = 'gray',
+        this.blockOneBackgroundColor = 'red',
+        this.blockTwoBackgroundColor = 'orange',
+        this.blockThreeBackgroundColor = 'blue';
+
+    // A convenience constructor for creating new instances of the state class
+    ReduxState.update(ReduxState oldState,
```
trentgrover-wf: the `oldState` fallback here makes perfect sense for a multivariate store

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+
+class UpdateBlockTwoBackgroundColorAction {}
+
+class UpdateBlockThreeBackgroundColorAction {}
+```
+
+```dart
+// A simple reducer that has a case for every action.
+// Because the example is relatively contrived to keep things simple, there are other
+// ways this could be done given that all the actions are pretty much the same,
```
trentgrover-wf: if it's easy, you could include both this as the verbose option and an equivalent shortened version

trentgrover-wf: actually if you're just talking about bubbling the `var color` line outside of the if statements or breaking it out into a separate utility method (my preference), I don't think it's any harder to grok with that change so you could probably stick with the short version

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+1. __Decide which stores will be Redux and which stores will be Influx.__ An Influx implementation can have both Redux stores and Influx stores. The rules are:
+    - A Redux component can talk to both a Redux store and an adapted Influx store at the same time.
+    - A Flux component can talk to an Influx store and a Flux store at the same time.
+    - A `connectFlux` component can talk to an adapted Influx store.
+
+    The result of these rules is that a store can be converted directly to Redux [iff](https://en.wikipedia.org/wiki/If_and_only_if) it will be used exclusively by Redux components, but a component can be converted to a Redux component if it will pull data from either an adapted Influx store or a traditional Flux store.
+
+    If the stores are not overly complex and all of the related UI can be updated to Redux connected components (extending from `UiComponent2`), then going straight to Redux may be a good option.
```
trentgrover-wf: some of the verbage around which piece to convert when and how to decide reads as a bit circuitous - could some of the prose be eliminated by using a simple flowchart / decision tree?

doc/flux_to_redux.md
```diff
@@ -0,0 +1,641 @@
...
+
+    For code examples on what this could look like, compare the different components in the [advanced web example](../web/flux_to_redux/advanced).
+    - __A Flux component__ will operate exactly the same way. 
+        - Pass in the Influx store instance as a prop (which should already be done). Not the `FluxToReduxAdapterStore` instance, but _the same Influx store instance_ used to istantiate the `FluxToReduxAdapterStore` object. 
+        - The actions prop should also be _the same action class instance_ passed into the `FluxToReduxAdapterStore` constructor. 
+    - __A Redux component__ will use the `connect` function. 
+        - The `pure` parameter on `connect` should be set to `false`, or else it will not receive regular updates. 
+        - If the component tree was wrapped in a `ReduxMultiProvider`, the `context` parameter should also be set to the relevant store context.
+    - __A Connected Flux component__ is essentially the same as a Redux component.
+        - The only exception is thate `mapDispatchToProps` (Redux `connect` parameter) is `mapActionsToProps` for a Connected Flux component.
```
trentgrover-wf: `thate` -> `that`

doc/flux_to_redux.md
joebingham-wk: Good point. I was leaning on using the standard way of doing things even when stuff was verbose / unused so readers would consistently see the same patterns, but I believe you're correct that it makes more sense to have the example be practical

doc/flux_to_redux.md
joebingham-wk: I was actually thinking that the actions and reducer together could be refactored to really only require one action - just an `UpdateColorAction` that takes in all the colors, including one that's changed, and passes them all in to the `update` constructor. That was really unclear though, so I updated the wording to be a little more specific as to what 'simplified' meant.

I had thought about breaking color generation into a different method though, and was just on the fence originally. So I updated that as well!

doc/flux_to_redux.md
joebingham-wk: Yep, good call! I added that and reworded it a little bit

web/flux_to_redux/advanced/examples/influx_implementation/components/redux_big_block.dart
```diff
@@ -0,0 +1,149 @@
...
+// [4] Make a `composeHocs` call, passing in a `List`.
+// [5] Make a `connect` call for each store the component needs access to, with each call being an element in the `List`.
+// [6] Tie `connect` to the correct context by using the `context` parameter.
+// [7] Use `mapStateToProps` and `mapDispatchToProps` to point props to the correct state values.
+// [8] Set the `pure` constructor on `connect` to be `false`.
+
+/*[4]*/ UiFactory<ReduxBigBlockProps> ConnectedReduxBigBlock = composeHocs([
+  /*[5]*/ connect<RandomColorStore, ReduxBigBlockProps>(
+    /*[6]*/ context: randomColorStoreContext,
+    /*[7]*/ mapStateToProps: (state) => (ReduxBigBlock()
```
greglittlefield-wf: The way these comments affect indentation makes it hard to see the visual hierarchy/nesting of things. Thoughts on putting comments on their own lines or at the end of the lines, like so? https://gist.github.com/greglittlefield-wf/deb3f9eeca051aa46284080058f13d10

joebingham-wk: Yep, I like that a lot better!! I'm happy to go through and do that for other instances if it's easier, or happy to accept any PRs to fix it. Just let me know! :)

greglittlefield-wf: I'll make a PR, thanks!

doc/flux_to_redux.md
```diff
@@ -37,7 +37,7 @@ No. OverReact Redux is meant to help to provide a recommended state management m
 To evaluate if the refactor is worth it, the details of OverReact Redux can be found in the [OverReact Redux documentation](./over_react_redux_documentation.md). That document paired with this guide will illustrate the full scope of costs and benefits.
 
 ## A Simple Example
-To illustrate the basic changes that will occur, this section will go through a basic Flux architecture and then show that same system with Redux instead. This section will also have a step by step list of instructions for an update as straightforward as this one. An actual working example can be found in the [Flux to Redux Example](../web/flux_to_redux/simple/readme.md).
+To illustrate the basic changes that will occur, this section will go through a basic Flux architecture and then show that same system with Redux instead. This section will also have a step by step list of instructions for an update as straightforward as this one. An actual working example can be found in the [Flux to Redux Example](../web/flux_to_redux/simple).
```
joebingham-wk: This change was actually a topic of conversation:
https://github.com/Workiva/over_react/pull/438#discussion_r369290885

joebingham-wk: I'm happy to go either way, just FYI! :)

joebingham-wk:
+1 to Greg's commits

greglittlefield-wf:
Oh wait, I could have misinterpreted that that; it says "Reflux", not "Influx"? Is it supposed to be "Reflux"? Was that term introduced anywhere?

joebingham-wk:
Nope, that's just suuuuuuper old. "Reflux" was the precursor to "Influx". Sooo that's crazy outdated haha. I'll take a look!

PR #478, opened 2020-03-05
New boilerplate parser confidence refactor

greglittlefield-wf:
### Major Changes:
- Be more strict 
- Move version confidence logic out of member classes and into member detection. This:
    - allows for computation before we create member instances, so we don't have to filter them out after the fact.
    - unifies the "is this a boilerplate member" logic with the "which version is it" logic, since the two have a lot of overlap and resulted in duplicate code
- Remove `BoilerplateVersion.noGenerate`; instead, use a confidence threshold to determine whether something should be generated. For declarations, this is the mean confidence of all members.
    - This allows us to easily determine whether leftover boilerplate members should result in warning/errors (above threshold) or be ignored (below threshold)
    - This removes the multiple ways to say something shouldn't be generated (low confidence in boilerplates vs high confidence in `noGenerate`)

- Create classes for version confidence instead of using a Map, which was more susceptible to missing keys

### Minor changes / fixes
- Find components' props by the UiProps parameter if the name doesn't match
- Rename `BoilerplateVersion` -> `Version` since it was too long lol
- Fix inverted `classish.abstractKeyword` null-check
- Clean up `BoilerplateMemberDetector` / decouple it from BoilerplateMembers

### DIFF
--- a/lib/src/builder/generation/parsing/declarations.dart
+++ b/lib/src/builder/generation/parsing/declarations.dart
@@ -252,10 +252,12 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
     if (resolveVersion([stateClass]) == BoilerplateVersion.noGenerate) continue;
     errorCollector.addError('State class is missing factory and/or component.', errorCollector.spanFor(stateClass.node));
   }
-  for (var componentClass in components) {
-    if (resolveVersion([componentClass]) == BoilerplateVersion.noGenerate) continue;
-    errorCollector.addError('componentClass class is missing factory and/or props.', errorCollector.spanFor(componentClass.node));
-  }
+  // ignore components since they don't require any generation if not accompanied by a props class and factory
+  // todo don't ignore components with a sufficiently high confidence
+//  for (var componentClass in components) {
+//    if (resolveVersion([componentClass]) == BoilerplateVersion.noGenerate) continue;
+//    errorCollector.addError('componentClass class is missing factory and/or props.', errorCollector.spanFor(componentClass.node));
+//  }
 }
 
 class FactoryGroup {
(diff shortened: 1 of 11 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/builder/generation/parsing/declarations.dart
```diff
@@ -49,14 +49,6 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
   final states = members.states;
   final stateMixins = members.stateMixins;
 
-  // -----------------------------------------------------------------------------------------------
-  //
-  // Legacy map views
-  //
-  // Remove them so they don't get grouped with anything else and throw things off.
-  //
-  props.removeWhere((p) => p.isLegacyMapView);
```
greglittlefield-wf: I can't remember why this isn't needed anymore... 😬 I might want to put it back?

lib/src/builder/generation/parsing/version.dart
```diff
@@ -5,58 +5,114 @@ import 'members.dart';
...
+    final confidenceMap = {
+      'v2_legacyBackwardsCompat': v2_legacyBackwardsCompat,
+      'v3_legacyDart2Only': v3_legacyDart2Only,
+      'v4_mixinBased': v4_mixinBased,
+    };
+
+    return '${runtimeType.toString()}; confidence:$confidenceMap';
+  }
+}
+
```
joebingham-wk: #nit extra line
```suggestion
```

lib/src/builder/generation/parsing/declarations.dart
greglittlefield-wf: Good call, okay, I'll reinstate it and add a todo to figure out if we can remove it

PR #529, opened 2020-06-23
Unit tests for boilerplate_utils

sydneyjodon-wk:
## Motivation
  
We need to write tests for all the utilities in `boilerplate_util.dart`.
## Changes
  
- Wrote tests for utilities.
- Updated `boilerplate_util.dart` as needed based on findings in tests.
## Review

  

Please review: @greglittlefield-wf @joebingham-wk @aaronlademann-wf @evanweible-wf @luisgomez-wk @corwinsheahan-wf 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

tools/analyzer_plugin/lib/src/util/boilerplate_utils.dart
```diff
@@ -47,6 +47,14 @@ void addOverReactGeneratedPartDirective(DartFileEditBuilder builder, Compilation
     lastDirective = unit.directives.last;
   }
 
+  if (lastDirective is PartDirective && overReactGeneratedPartDirectiveIsValid(lastDirective, fileUri)) return;
+
+  final overReactPartDirective = getOverReactGeneratedPartDirective(unit);
+  if(overReactPartDirective != null && !overReactGeneratedPartDirectiveIsValid(overReactPartDirective, fileUri)) {
+    fixOverReactGeneratedPartDirective(builder, unit, fileUri);
```
sydneyjodon-wk, line 54: Previously, if an over_react part directive existed, but didn't match the file name, this utility would add a valid over_react part directive without removing the old one. It seemed better to replace the old one.

tools/analyzer_plugin/test/test_util.dart
```diff
@@ -115,7 +125,7 @@ Future<Map<String, ResolvedUnitResult>> parseAndGetResolvedUnits(Map<String, Str
   for (final path in dartSourcesByPath.keys) {
     final context = collection.contextFor(transformPath(path));
     final result = await context.currentSession.getResolvedUnit(transformPath(path));
-    if (result.errors.isNotEmpty) {
+    if (shouldThrowErrors && result.errors.isNotEmpty) {
```
sydneyjodon-wk, line 128: I don't know if ignoring the parse errors has bad side effects. I needed the resolved ast to get the uri information, but it didn't matter if there were static analysis errors in the testcase. Is there a better way I should be doing this?

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
```diff
@@ -0,0 +1,282 @@
...
+        expect(editList.first.replacement, 'part \'different_file_name.over_react.g.dart\';');
+      });
+    });
+  });
+}
+
+Future<void> overReactGeneratedPartDirectiveIsValidTestHelper({
+  bool isPartOverReact,
+  String path,
+  bool expectedResult,
```
joebingham-wk, line 260: #nit since these are optional, it may be nice to add defaults or the @required annotation. For the bool values, I personally think it's can be nice to use defaults 🤷‍♂️

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
```diff
@@ -0,0 +1,282 @@
...
+}) async {
+  final result = await parseAndGetResolvedUnit(
+    isPartOverReact ? sourceWithOverReactPart : sourceWithNonOverReactPart,
+    path: path,
+  );
+  final part = result.unit.directives.whereType<PartDirective>()?.first;
+  expect(part, isNotNull);
```
joebingham-wk, line 267: I could definitely be wrong, but I don't believe this test will ever fail directly. I don't believe `whereType` would ever return `null`, instead it returns an empty list. Then `.first` would throw a state error. This might fine since it still fails the test, but `firstOrNull` or `firstWhere` with a type check would have the behavior I'd expect 267 to be checking for

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
```diff
@@ -0,0 +1,282 @@
...
+        expect(editList.first.replacement, 'part \'different_file_name.over_react.g.dart\';');
+      });
+    });
+  });
+}
+
+Future<void> overReactGeneratedPartDirectiveIsValidTestHelper({
+  bool isPartOverReact,
+  String path,
+  bool expectedResult,
```
joebingham-wk, line 260: #superdupernit Since there are three different `expects`, I think this name could be more descriptive to the actual `part` validation (`expectedPartValidity`, `partIsValid`, etc). Though, I know the scope of this utility is super limited so ignore me if I'm being too nit picky 😄

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
```diff
@@ -0,0 +1,282 @@
...
+          final sourceFileEdits = await getSourceFileEdits(
+            sourceWithNonOverReactPart,
+            (builder, result) => removeOverReactGeneratedPartDirective(builder, result.unit),
+            path: 'foo.dart',
+          );
+          expect(sourceFileEdits, isEmpty);
+        });
+      });
+
+      test('removes over_react part directive if it exists', () async {
```
joebingham-wk, line 191: Since we've tested how the utilities react when adding a part when one already exists, it may be worth testing how the utilities react when removing the over_react part when another one exists

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
sydneyjodon-wk: Good point! I'll update those!

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
sydneyjodon-wk: Good catch! I'll update it to `firstOrNull`

tools/analyzer_plugin/test/unit/util/boilerplate_utils_test.dart
sydneyjodon-wk: Good idea! I'll add that test case.

tools/analyzer_plugin/lib/src/util/boilerplate_utils.dart
luisgomez-wk: Good call

luisgomez-wk:
@Workiva/release-management-p

PR #530, opened 2020-06-23
Do not lint comments on declarations in a file with a MapView

aaronlademann-wf:
Props mixin map views look a lot like a `BoilerplateFactory`, but they are different. Doc comments found on their analogous props mixin should not be linted as being in the wrong place.

This PR fixes that, and updates the test case in the playground to account for this.


aviary3-wk:
## Security Insights

## Action Items
---

PR #533, opened 2020-06-24
Enable diagnostics to run on all files in project instead of just open ones

greglittlefield-wf:
AnalysisDriver doesn't fully resolve files that are added via `addFile`; they need to be either explicitly requested via `getResult`/etc, or added to `priorityFiles`.

Since by default, all files just get added via `addFile`, we weren't getting resolved units for those files, so our Diagnostics weren't getting run.

This PR adds logic to update `priorityFiles` on the driver to include:

- Any files prioritized by the analysis server via `handleAnalysisSetPriorityFiles`
- All other files the driver has been told to analyze via addFile (in `ServerPlugin.handleAnalysisSetContextRoots`)

As a result, [processDiagnosticsForResult] will get called with resolved units, and thus __all of our diagnostics
will get run on all files in the repo__ instead of only the currently open/edited ones!

### DIFF
--- a/tools/analyzer_plugin/lib/src/plugin.dart
+++ b/tools/analyzer_plugin/lib/src/plugin.dart
@@ -121,7 +121,7 @@ class OverReactAnalyzerPlugin extends ServerPlugin
     super.driverForPath(path).addFile(path);
   }
 
-//  @override
+  //  @override
 //  List<OutlineContributor> getOutlineContributors(String path) {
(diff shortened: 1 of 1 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

aaronlademann-wf:
+10

Definitely analyzed everything 😅 

image

PR #551, opened 2020-06-24
Preserve namespace when converting to createRef

### LINKED ISSUE #544: [BUG] `createRef` edit strips namespace on component type
_(e.g., `foo.MyComponent` becomes `MyComponent`)_.

aaronlademann-wf:
## Motivation
When creating the `createRef<typeName>()` string as part of the quick fix, the namespace of components that are imported within a namespace is not being preserved within the _"typeName"_ portion of the string.

## Changes
1. Add a new `componentClassTypeName` getter to `FluentComponentUsage` that returns the class name for a component, along with the namespace _(using the new `FluentComponentUsage.namespaceName` getter)_.
1. Update the test case file to include a component that is imported within a namespace.

Fixes #544


aviary2-wf:
## Security Insights

## Action Items
---

tools/analyzer_plugin/lib/src/component_usage.dart
```diff
@@ -40,6 +42,29 @@ class FluentComponentUsage {
     return null;
   }
 
+  /// Returns the name of the `UiComponent` class instance for the usage, complete with any
+  /// [namespaceName] if one exists.
+  String get componentClassTypeName {
```
greglittlefield-wf, line 47: These implementations aren't valid for DOM components, and should either be updated to handle them, or documented as such

tools/analyzer_plugin/lib/src/component_usage.dart
```diff
@@ -40,6 +42,29 @@ class FluentComponentUsage {
     return null;
   }
 
+  /// Returns the name of the `UiComponent` class instance for the usage, complete with any
```
greglittlefield-wf, line 45:
```suggestion
  /// Returns the inferred name of the `UiComponent` class instance for the usage, complete with any
```
Since we're not actually looking up the component class

tools/analyzer_plugin/lib/src/component_usage.dart
```diff
@@ -40,6 +42,29 @@ class FluentComponentUsage {
...
+    if (namespaceName == null) return '${componentName}Component';
+
+    return '$namespaceName.${componentName}Component';
+  }
+
+  /// The name of the imported usage namespace.
+  String get namespaceName {
+    if (builder is MethodInvocation) {
+      return (builder as MethodInvocation).target.tryCast<Identifier>()?.name;
+    } else if (builder is FunctionExpressionInvocation) {
```
greglittlefield-wf, line 58: In which case would this be a `FunctionExpressionInvocation`?

aaronlademann-wf: That is what it showed up as for an invoked factory I think. If you put a breakpoint in here - I think the MethodInvocation only triggers for Dom components maybe?

PR #552, opened 2020-06-24
Fix num type check that only works when implicit casts are enabled

### LINKED ISSUE #539: [BUG] Numbers are treated as invalid children when implicit casts are disabled

greglittlefield-wf:
Fixes #539

## Motivation
The existing type-check for whether a child was num was backwards, and thus only worked with implicit casts are enabled.

The code, 
```dart
typeSystem.isAssignableTo(typeProvider.numType, type)
```
translates to: "could a value of type `num` be assigned to `type`"?

For the case of `int`, "could a value of type `num` be assigned to `int`", the answer is:
- with implicit casts enabled: true
- with implicit casts disabled: false (because you can't assign `num` to `int` without casting it)

## Changes
Fix ordering of arguments an use more explicit `isSubTypeOf` check


aviary3-wk:
## Security Insights

## Action Items
---

PR #537, opened 2020-06-24
Warn for all except last duplicate prop

corwinsheahan-wf:
## Motivation
  
We should retain the last duplicate prop in a prop cascade
## Changes
  
Flag every duplicate prop except for the last instance.
## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 

### QA Checklist
- [ ] Tests were updated and provide good coverage of the changeset and other affected code
- [ ] Manual testing was performed if needed
    - [ ] Steps from PR author: 
        
    - [ ] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [ ] A Client Platform member has reviewed these changes
- [ ] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/over_react/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/over_react/blob/master/CONTRIBUTING.md#manual-testing-criteria

PR #554, opened 2020-06-24
Update dupe prop fix to remove all dupes except the last one

### LINKED ISSUE #547: Update the quick fix for the dupe props lint to remove *all* dupes
_(preserving the last one in the cascade)_ instead of having to do it one at a time.

Also, revert the changes from #537 so that all usages of the prop are linted. Then, when the quick fix is applied - remove all except the last usage.

Side note: Figure out what the `appliedTogetherMessage` is supposed to do / how that relates to a fix being applied more than once versus making the fix be applied recursively by default.

aaronlademann-wf:
## Motivation
1. #537 made it so that the last of 1 or more duplicate props didn't show as a lint.
2. Fixing duplicates had to be done on individual props, and with the changes from #537 reverted, the last of the dupes could be removed - which would change the runtime behavior of the component.

## Changes
1. Revert #537 so that all dupes are linted.
1. Update the fix so that all dupes except the last one is removed - preserving the runtime behavior of the component.

Fixes #547


aviary3-wk:
## Security Insights

## Action Items
---

PR #555, opened 2020-06-24
Increase severity of / narrow scope of arrow function in cascade lint

### LINKED ISSUE #541: Arrow function lints are suuuper noisy
Since people tend to just put them at the end of the cascade. 

We should consider turning the lint off by default, or only linting as an error when it isn't  the last item in the cascade.

aaronlademann-wf:
## Motivation
1. The arrow functions in prop cascades lint is very noisy - especially since most consumers just move them to be the last setter in the cascade to avoid the other core dart errors caused by placing it in the middle of the cascade.

## Changes
1. Upgrade the lint to an error
1. Only lint when the unwrapped function expression is in the middle of a cascade with more than one setter.

Closes #541

### DIFF
--- a/tools/analyzer_plugin/lib/src/diagnostic/arrow_function_prop.dart
+++ b/tools/analyzer_plugin/lib/src/diagnostic/arrow_function_prop.dart
@@ -6,13 +6,14 @@ import 'package:over_react_analyzer_plugin/src/fluent_interface_util.dart';
 
 class ArrowFunctionPropCascadeDiagnostic extends ComponentUsageDiagnosticContributor {
   static final code = DiagnosticCode(
-      'over_react_cascaded_arrow_functions',
-      'Unparenthesized arrow function values prevent subsequent cascades',
-      AnalysisErrorSeverity.WARNING,
-      AnalysisErrorType.STATIC_WARNING);
+    'over_react_cascaded_arrow_functions',
+    'Never place un-parenthesized arrow functions in the middle of prop setter cascades.',
+    AnalysisErrorSeverity.ERROR,
+    AnalysisErrorType.SYNTACTIC_ERROR,
+    correction: 'Wrap arrow functions in parentheses when placed in the middle of prop setter cascades.',
+  );
 
-  static final fixKind = FixKind(code.name, 200, 'Wrap arrow function in parentheses',
-      appliedTogetherMessage: 'Wrap arrow functions in parentheses');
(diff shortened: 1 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

tools/analyzer_plugin/lib/src/diagnostic/arrow_function_prop.dart
```diff
@@ -21,6 +23,10 @@ class ArrowFunctionPropCascadeDiagnostic extends ComponentUsageDiagnosticContrib
       if (rhs is FunctionExpression && rhs.body is ExpressionFunctionBody) {
         final body = rhs.body as ExpressionFunctionBody;
 
+        // If a cascade expression is not found in the body, it is not an un-parenthesized
+        // function expression in the middle of another cascade... do not lint.
+        if (allDescendantsOfType<CascadeExpression>(body).isEmpty) continue;
```
corwinsheahan-wf, line 28: Could this incorrectly avoid adding this lint to something like this snippet below? There's probably not a straightforward way to differentiate this case from cases where we want to lint, though.
```
        ..foo = () => FooObject()..someMethod()
```

aaronlademann-wf: @corwinsheahan-wf in offline discussions - we decided that linting this is a good idea anyway - because one formatted - cascades like that without parentheses wrapping around the entire expression... makes stuff extremely difficult to read.

For example... I've seen quite a bit of stuff like this - which makes me want to 🤮 

```dart
@override
render() {
  return (Foo()
    ..id = 'foo'
    ..className = 'dont-mess-with-foo'
    ..barRenderer = () => (Bar()
      ..id = 'bar'
      ..aria.label = 'Yeah I put my feet on Foos couch!'
    )(
      'I am a bar',
    )
  )(); 
}
```

Once wrapped - the nested function is much easier to reason about, IMO:

```dart
@override
render() {
  return (Foo()
    ..id = 'foo'
    ..className = 'dont-mess-with-foo'
    ..barRenderer = (() =>
        (Bar()
          ..id = 'bar'
          ..aria.label = 'Yeah I put my feet on Foos couch!'
        )(
          'I am a bar',
        )
    )
  )();
}
```

corwinsheahan-wf: I like it

PR #556, opened 2020-06-24
Fix boilerplate generated part directive lint false positives

### LINKED ISSUE #528: [BUG] Address boilerplate validation false positives that appear as a result of collecting errors within the `for (final member in allMembers)` loop.
Since those are just *potential* members - false positives can occur. 

Example: 

```dart
mixin WsDraggableItemIndexingApi<T extends WsDraggableItemProps, S extends WsDraggableItemState>
    on UiStatefulComponent2<T, S> {
```

is getting picked up as a potential component - so the "missing an `.over_react.g.dart`" lint error is showing up when it should not.

Another example (abstract component):

```dart
abstract class AbstractToolbarItemComponent
```

We should only iterate over members that are part of declaarations:
```diff
+final allMembersPartOfADeclaration = declarations.expand((d) => d.members).toSet()
+for (final member in allMembersPartOfADeclaration) {
-for (final member in members) {
  ...
```

aaronlademann-wf:
## Motivation
1. The lints that appear when a generated `.over_react.g.dart` `PartDirective` is missing - are appearing when they should not.  This is happening because `members` are really just things that *might* get generated.

## Changes
1. Bail out before producing the lints if the version confidence of the builder is not `likely` or `certain`.

Fixes #528

### DIFF
--- a/tools/analyzer_plugin/lib/src/diagnostic/bool_prop_name_readability.dart
+++ b/tools/analyzer_plugin/lib/src/diagnostic/bool_prop_name_readability.dart
@@ -28,9 +28,12 @@ class BoolPropNameReadabilityDiagnostic extends DiagnosticContributor {
         if (field.type != typeProvider.boolType) continue;
         if (propName == null) continue; // just in case
 
+        final fieldDecl = propsClass.getField(propName);
+        if (fieldDecl == null) continue;
+
         final readability = checkBoolPropReadability(propName);
         if (!readability.isReadable) {
-          collector.addError(code, result.locationFor(propsClass.getField(field.name)),
+          collector.addError(code, result.locationFor(fieldDecl),
               errorMessageArgs: [propsClass.name, propName, allowedPrefixesForBoolProp.join(', ')]);
         }
       }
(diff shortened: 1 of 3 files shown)


tools/analyzer_plugin/lib/src/diagnostic/bool_prop_name_readability.dart
```diff
@@ -28,9 +28,12 @@ class BoolPropNameReadabilityDiagnostic extends DiagnosticContributor {
         if (field.type != typeProvider.boolType) continue;
         if (propName == null) continue; // just in case
 
+        final fieldDecl = propsClass.getField(propName);
+        if (fieldDecl == null) continue;
+
         final readability = checkBoolPropReadability(propName);
         if (!readability.isReadable) {
-          collector.addError(code, result.locationFor(propsClass.getField(field.name)),
+          collector.addError(code, result.locationFor(fieldDecl),
```
aaronlademann-wf, line 36: This was an unrelated runtime exception I encountered while testing this out in WSD.

aviary3-wk:
## Security Insights

## Action Items
---

Issue #514, opened 2020-06-23
Establish Analyzer Plugin Testing


aaronlademann-wf:

aaronlademann-wf:
PR: https://github.com/Workiva/over_react_analyzer_plugin/pull/22

PR #559, opened 2020-06-24
Integration test setups and base classes

evanweible-wf:
## Motivation
Provide a way to test individual assist and diagnostic contributors using a real `AnalysisDriver` with source files that are fully resolved with a real `package:over_react/over_react.dart` import. Should be able to apply `SourceChange`s produced by these contributors to verify the resulting source contents.

## Changes
- Added several testing base classes that handle the construction of an `AnalysisDriver` and a `TestPlugin` that utilizes a single contributor that is being tested. Some utilities for getting the assists or errors with fixes from the contributor are also provided.
- Two bare minimum examples of how to test an assist and diagnostic contributor.


aviary3-wk:
## Security Insights

## Action Items
---

tools/analyzer_plugin/test/integration/assists/add_props_test.dart
```diff
@@ -0,0 +1,67 @@
...
+@reflectiveTest
+class AddPropsAssistTest extends AssistTestBase {
+  static int expectedPriority = AddPropsAssistContributor.addPropsKind.priority;
+
+  @override
+  AsyncAssistContributor getContributorUnderTest() =>
+      AddPropsAssistContributor();
+
+  Future<void> test_noPropsAndNoParens() async {
+    var source = newSource('test.dart', '''
```
evanweible-wf, line 25: > A helper to load a render method from a path would be really nice, especially for harder cases

_From Luis on the previous PR_

tools/analyzer_plugin/test/integration/test_bases/analysis_driver_test_base.dart
```diff
@@ -0,0 +1,115 @@
...
+import 'package:analyzer/src/dart/analysis/driver.dart'
+    show AnalysisDriver, AnalysisDriverScheduler;
+import 'package:analyzer/src/dart/analysis/driver.dart';
+import 'package:analyzer/src/dart/analysis/file_state.dart';
+import 'package:analyzer/src/dart/analysis/performance_logger.dart';
+import 'package:analyzer/src/generated/engine.dart';
+import 'package:analyzer/src/generated/source.dart';
+import 'package:analyzer/src/generated/source_io.dart';
+import 'package:analyzer/src/source/package_map_resolver.dart';
+import 'package:analyzer/src/test_utilities/mock_sdk.dart';
```
evanweible-wf, line 18: > Are all of these src imports necessary, or could we potentially eliminate some?

_From @greglittlefield-wf on the previous PR_

tools/analyzer_plugin/test/integration/test_bases/analysis_driver_test_base.dart
```diff
@@ -0,0 +1,115 @@
...
+        packageName: [await _loadRealPackage(packageName, resourceProvider)],
+    };
+
+    // Setup a testing directory. All calls to [newSource] will create files
+    // within this directory.
+    _testPath = resourceProvider.newFolder('/test').path;
+
+    final logger = PerformanceLog(StringBuffer());
+    final analysisScheduler = AnalysisDriverScheduler(logger)..start();
+    _analysisDriver = AnalysisDriver(
```
evanweible-wf, line 65: > I noticed that we also create a driver in plugin.dart; could we reuse any of that code here?

_From @greglittlefield-wf on the previous PR_

tools/analyzer_plugin/test/integration/test_bases/server_plugin_contributor_test_base.dart
```diff
@@ -0,0 +1,79 @@
...
+import '../mocks.dart';
+import 'analysis_driver_test_base.dart';
+
+abstract class ServerPluginContributorTestBase extends AnalysisDriverTestBase {
+  PluginForTest get testPlugin => _plugin;
+
+  Source applySourceChange(SourceChange change, Source expectedSource) {
+    final path = p.normalize(expectedSource.uri.toFilePath());
+    final applicableFileEdits = change.edits.where((edit) => edit.file == path);
+    expect(applicableFileEdits.map((edit) => edit.file), contains(path),
```
evanweible-wf, line 20: > Something about expects in the AssistTestBase class (which isn't really testing anything) rubs me the wrong way. Would you be opposed to replacing these with asserts?

_From @luisgomez-wk on the previous PR_

tools/analyzer_plugin/test/integration/test_bases/analysis_driver_test_base.dart
greglittlefield-wf: Not a big deal if not

tools/analyzer_plugin/test/integration/test_bases/server_plugin_contributor_test_base.dart
evanweible-wf: Yeah, I guess I don't have a strong preference here. My thought process here was that in tests, you'd call `applySourceChange()` right before verifying the resulting source contents, so adding another expectation within this method that verifies the given change includes edits for the expected source would be helpful. ¯\_(ツ)_/¯

tools/analyzer_plugin/test/integration/matchers.dart
```diff
@@ -0,0 +1,47 @@
+import 'package:analyzer_plugin/protocol/protocol_common.dart';
+import 'package:over_react_analyzer_plugin/src/diagnostic_contributor.dart';
+import 'package:test/test.dart';
+
+import 'mocks.dart';
+
+TypeMatcher<AnalysisError> isAnalysisErrorForDiagnostic(
+    DiagnosticCode diagnosticCode,
+    {SourceSelection atSelection,
+    bool hasFix}) {
```
greglittlefield-wf, line 10: This should be required or have a default

tools/analyzer_plugin/test/integration/assists/add_props_test.dart
```diff
@@ -0,0 +1,67 @@
...
+import 'package:over_react/over_react.dart';
+var foo = Dom.div()('hello');
+''');
+    var selection = createSelection(source, '#Dom.div#');
+    final assists = await getAssists(selection);
```
greglittlefield-wf, line 30: Do `createSelection` and `getAssists` support a selection of length 0? e.g., could you do the following?
```dart
    var selection = createSelection(source, '##Dom.div');
    final assists = await getAssists(selection);
```

tools/analyzer_plugin/test/integration/test_bases/server_plugin_contributor_test_base.dart
```diff
@@ -0,0 +1,79 @@
...
+        reason: 'Expected SourceChange to include edits for "$path".');
+    final updated = SourceEdit.applySequence(
+      resourceProvider.getFile(path).readAsStringSync(),
+      [for (final fileEdit in applicableFileEdits) ...fileEdit.edits],
+    );
+    final file = resourceProvider.updateFile(path, updated);
+    return file.createSource();
+  }
+
+  SourceSelection createSelection(Source source, String target, {int offset}) {
```
greglittlefield-wf, line 30: Some simple doc comments on `applySourceChange`/`createSelection` would be helpful for someone trying to understand what's going on in tests

tools/analyzer_plugin/test/integration/matchers.dart
```diff
@@ -0,0 +1,47 @@
...
+  }
+  if (atSelection != null) {
+    matcher = matcher.havingLocation(atSelection.offset, atSelection.length);
+  }
+  return hasFix ? matcher.thatHasFix() : matcher.thatHasNoFix();
+}
+
+extension AnalysisErrorHavingUtils on TypeMatcher<AnalysisError> {
+  TypeMatcher<AnalysisError> havingCode(String code) =>
+      having((e) => e.code, 'code', code);
```
greglittlefield-wf, line 26: TIL about `having` 😮

tools/analyzer_plugin/test/integration/test_bases/server_plugin_contributor_test_base.dart
evanweible-wf: Will do!

tools/analyzer_plugin/test/integration/assists/add_props_test.dart
evanweible-wf: They do! I'll add some additional test cases and docs to clarify this.

tools/analyzer_plugin/test/integration/assists/add_props_test.dart
evanweible-wf: Actually I'm not sure I follow, @luisgomez-wk do you have an example of what you mean?

tools/analyzer_plugin/test/integration/test_bases/server_plugin_contributor_test_base.dart
luisgomez-wk: I think checking the edit is applied to the expected source would be helpful!

My worry is that if the expect in this function fails, the test writer will see their test failing, and the stack trace dart gives them will not be in this base class instead of their test

I think an assert would make it clearer that the failure is a side effect of their edits applying on the wrong source

luisgomez-wk: TLDR; I think it's fine for now for the sake of merging! And if it does result in confusing failures it'll be easy to change later!

tools/analyzer_plugin/test/integration/assists/add_props_test.dart
luisgomez-wk: I was thinking something like `newSourceFromPath('../playground/foo.dart')`

Thinking about it again though, it seems unnecessary. Especially since I learned you can use a comment to get syntax highlighting in JetBrains, e.g.
```dart
newSource('test.dart', /*language=dart*/ r'''
''');
```

tools/analyzer_plugin/lib/src/diagnostic/arrow_function_prop.dart
```diff
@@ -51,9 +51,6 @@ class ArrowFunctionPropCascadeDiagnostic extends ComponentUsageDiagnosticContrib
 
   @override
   computeErrorsForUsage(result, collector, usage) async {
-    // If there is only one cascaded prop, do not lint
-    if (usage.cascadedProps.length == 1) return;
-
```
evanweible-wf, line 56: This is effectively already handled by the `allDescendantsOfType<CascadeExpression>(body).isEmpty` check below

tools/analyzer_plugin/test/integration/test_bases/assist_test_base.dart
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  /// Returns all assists produced at [selection].
+  Future<List<PrioritizedSourceChange>> _getAllAssists(SourceSelection selection) async {
+    final parameters = EditGetAssistsParams(
+        resourceProvider.convertPath(selection.source.uri.path), selection.offset, selection.length);
+    return (await testPlugin.handleEditGetAssists(parameters))
+        .assists
+        .where((psc) => psc.change.id == assistKindUnderTest.id)
+        .toList();
```
evanweible-wf, line 63: fyi @greglittlefield-wf I ended up filtering this to exclude other assists because it seemed like there was going to be a lot of overlap for assists on component usages and it seems a little brittle that adding a new assist in that area would mean having to update unrelated tests to ignore the new assist. It also seems like less of an issue for assists since they're not actually errors and it doesn't hurt the UX (IMO) for multiple to show up in the code actions IDE overlay. Thoughts?

greglittlefield-wf: Good call, makes sense to me!

I do think there are some places where UX can be affected if the priorities aren't right, but that's something that's probably better suited for manual testing or other tests.

tools/analyzer_plugin/test/integration/matchers.dart
```diff
@@ -0,0 +1,149 @@
...
+import 'package:test/test.dart';
+
+import 'test_bases/server_plugin_contributor_test_base.dart';
+
+/// Returns a [Matcher] that verifies that the item is a source change with a
+/// priority and ID matching [assistKind].
+///
+/// If [locatedAt] is non-null, the analysis error's location (offset and
+/// length) matches it
+TypeMatcher<PrioritizedSourceChange> isAssist(AssistKind assistKind, {SourceSelection locatedAt}) {
```
evanweible-wf, line 15: Just realized it doesn't make sense for these `isAssist()` and `isFix()` matchers to support the `locatedAt` option because `PrioritizedSourceChange`s don't target a single selection range like an error does.. they can have multiple file edits and a resulting selection position that the cursor should be placed at. Going to remove `locatedAt` for these two and instead tests should just verify that applying the assist/fix results in the expected source contents.

tools/analyzer_plugin/test/integration/assists/add_props_test.dart
evanweible-wf: Gotcha. Yeah given that the source contents needed for each test suite will differ, too, I think it'd be tough to write a utility for that in the test base class.

tools/analyzer_plugin/test/integration/test_bases/analysis_driver_test_base.dart
evanweible-wf: I was able to remove one, but otherwise I think we're stuck with them for now.

greglittlefield-wf:
Fixes #514

PR #580, opened 2020-06-26
Add tests for toggle stateful assist

joebingham-wk:
Just added the tests specifically for the ToggleStatefulAssist!

### DIFF
--- a/tools/analyzer_plugin/test/integration/assists/boilerplate_assist_utils.dart
+++ b/tools/analyzer_plugin/test/integration/assists/boilerplate_assist_utils.dart
@@ -1,0 +1,64 @@
+mixin BoilerplateAssistTestStrings {
+  static const fileName = 'test.dart';
+
+  static const prefix = 'Foo';
+
+  static const stateMixin = '''
+
+mixin ${prefix}State on UiState {}
+''';
+
+  static const getInitialState = '''
+
+  @override
+  get initialState => (newState());
+''';
+
+  String componentName({bool withSelection = false}) => withSelection ? '#${prefix}Component#' : '${prefix}Component';
+
+  String simpleComponentSuperclass({bool withState = false}) =>
+      withState ? 'UiStatefulComponent2<${prefix}Props, ${prefix}State>' : 'UiComponent2<${prefix}Props>';
+
+  String fluxComponentSuperclass({bool withState = false}) =>
+      withState ? 'FluxUiStatefulComponent2<${prefix}Props, ${prefix}State>' : 'FluxUiComponent2<${prefix}Props>';
+
+  String simpleUiComponentSource({bool isStateful = false, bool shouldIncludeSelection = false}) {
+    return '''
+import 'package:over_react/over_react.dart';
+part 'test.over_react.g.dart';
+
+UiFactory<${prefix}Props> $prefix = _\$$prefix; // ignore: undefined_identifier
+
+mixin ${prefix}Props on UiProps {}
+${isStateful ? stateMixin : ''}
+class ${componentName(withSelection: shouldIncludeSelection)} extends ${simpleComponentSuperclass(withState: isStateful)} {
+  @override
+  get defaultProps => (newProps());
+${isStateful ? getInitialState : ''}
+  @override
+  render() {}
+}
+''';
+  }
+
+  String fluxUiComponentSource({bool isStateful = false, bool shouldIncludeSelection = false}) {
+    return '''
+import 'package:over_react/over_react.dart';
+part 'test.over_react.g.dart';
+
+UiFactory<${prefix}Props> $prefix = _\$$prefix; // ignore: undefined_identifier
+
+mixin ${prefix}Props on UiProps {}
+
+class ${prefix}Props = UiProps with FluxUiPropsMixin, ${prefix}PropsMixin;
+${isStateful ? stateMixin : ''}
+class ${componentName(withSelection: shouldIncludeSelection)} extends ${fluxComponentSuperclass(withState: isStateful)} {
+  @override
+  get defaultProps => (newProps());
+${isStateful ? getInitialState : ''}
+  @override
+  render() {}
+}
+''';
+  }
+}
--- a/tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
+++ b/tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
@@ -1,132 +1,151 @@
 import 'dart:async';
 
 import 'package:analyzer_plugin/utilities/assist/assist.dart';
-import 'package:over_react_analyzer_plugin/src/assist/add_props.dart';
 import 'package:over_react_analyzer_plugin/src/assist/toggle_stateful.dart';
-import 'package:over_react_analyzer_plugin/src/async_plugin_apis/assist.dart';
 import 'package:test/test.dart';
 import 'package:test_reflective_loader/test_reflective_loader.dart';
 
 import '../test_bases/assist_test_base.dart';
+import 'boilerplate_assist_utils.dart';
 
 void main() {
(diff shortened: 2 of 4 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
```diff
@@ -0,0 +1,151 @@
...
+
+  @override
+  AssistKind get assistKindUnderTest => ToggleComponentStatefulness.makeStateful;
+
+  Future<void> test_noAssistOnFactory() async {
+    final generatedSource = simpleUiComponentSource();
+
+    var source = newSource(BoilerplateAssistTestStrings.fileName, generatedSource);
+    var selection = createSelection(source,
+        generatedSource.replaceFirst(r'UiFactory<FooProps> Foo = _$Foo;', r'UiFactory<FooProps> #Foo# = _$Foo;'));
```
greglittlefield-wf, line 30: Why the replacement and not just the following?
```suggestion
    var selection = createSelection(source, r'UiFactory<FooProps> #Foo# = _$Foo;');
```
The whole source isn't needed to make a selection; the context before/after can be as small/large as you want

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
```diff
@@ -0,0 +1,151 @@
...
+class AddStatefulnessAssist extends AssistTestBase with BoilerplateAssistTestStrings {
+  static int expectedPriority = ToggleComponentStatefulness.statefulAssistPriority;
+
+  @override
+  AssistKind get assistKindUnderTest => ToggleComponentStatefulness.makeStateful;
+
+  Future<void> test_noAssistOnFactory() async {
+    final generatedSource = simpleUiComponentSource();
+
+    var source = newSource(BoilerplateAssistTestStrings.fileName, generatedSource);
```
greglittlefield-wf, line 28: #nit is having this long constant necessary? I would say just put in a dummy name, or make `fileName` an instance member so you don't have to prefix it
```suggestion
    var source = newSource(fileName, generatedSource);
```

greglittlefield-wf: We could also just make newSource's fileName arg be an optional/named arg, since a lot of the time you don't beed it
```suggestion
    var source = newSource(generatedSource);
```

tools/analyzer_plugin/test/integration/assists/boilerplate_assist_utils.dart
```diff
@@ -0,0 +1,64 @@
...
+
+  String componentName({bool withSelection = false}) => withSelection ? '#${prefix}Component#' : '${prefix}Component';
+
+  String simpleComponentSuperclass({bool withState = false}) =>
+      withState ? 'UiStatefulComponent2<${prefix}Props, ${prefix}State>' : 'UiComponent2<${prefix}Props>';
+
+  String fluxComponentSuperclass({bool withState = false}) =>
+      withState ? 'FluxUiStatefulComponent2<${prefix}Props, ${prefix}State>' : 'FluxUiComponent2<${prefix}Props>';
+
+  String simpleUiComponentSource({bool isStateful = false, bool shouldIncludeSelection = false}) {
```
greglittlefield-wf, line 25: >shouldIncludeSelection

See my note below about `#`

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
```diff
@@ -0,0 +1,151 @@
...
+    await expectNoAssist(selection);
+  }
+
+  Future<void> test_addsStatefulness() async {
+    var retrievedSource = newSource(BoilerplateAssistTestStrings.fileName, simpleUiComponentSource());
+    var selection = createSelection(retrievedSource, simpleUiComponentSource(shouldIncludeSelection: true));
+    final change = await expectAndGetSingleAssist(selection);
+    retrievedSource = applySourceChange(change, retrievedSource);
+    expect(retrievedSource.contents.data, simpleUiComponentSource(isStateful: true));
+  }
```
greglittlefield-wf, line 72: Since there's logic around the presence of `defaultProps`, can we add a test case for when it's not present?

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
```diff
@@ -0,0 +1,151 @@
...
+
+  Future<void> test_removesStatefulness() async {
+    var retrievedSource = newSource(BoilerplateAssistTestStrings.fileName, simpleUiComponentSource(isStateful: true));
+    var selection =
+        createSelection(retrievedSource, simpleUiComponentSource(shouldIncludeSelection: true, isStateful: true));
+    final change = await expectAndGetSingleAssist(selection);
+    retrievedSource = applySourceChange(change, retrievedSource);
+    expect(retrievedSource.contents.data, simpleUiComponentSource());
+  }
```
greglittlefield-wf, line 141: Since there's logic to conditionally remove `initialState`, can we add a test case for when it's not present?

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
```diff
@@ -0,0 +1,151 @@
...
+class ${withSelection ? '#TestComponent#' : 'TestComponent'} extends UiComponent<TestProps> {
+  @override
+  render() {}
+}
+''';
+
+    var source = newSource(BoilerplateAssistTestStrings.fileName, oldBoilerplate());
+    var selection = createSelection(source, oldBoilerplate(withSelection: true));
+    await expectNoAssist(selection);
+  }
```
greglittlefield-wf, line 64: Can we add a couple test cases for edge cases?

- non-supported component base class
- running this assist in a file with no component

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
joebingham-wk: Good call, I can shorten that! We could make it optional, but I think that's a larger question than this PR since the method is in use elsewhere (and those spots would have to be updated too)

greglittlefield-wf: Oh yeah for sure, we could definitely address that separately; just thought it'd be worth mentioning that these helper methods are still fairly new and we should feel free to change em to better suit our use cases!

joebingham-wk: True true, that's a good point!

tools/analyzer_plugin/test/integration/assists/toggle_statefulness_test.dart
joebingham-wk: For sure! How would the non-supported base class differ from https://github.com/Workiva/over_react/pull/580/files#diff-a479f45e42c16ab64749a5016a13af57R43? Is there another base class we should test against?

But added a test with a file with no component!

greglittlefield-wf: My thought process was to test `_getNewBase ` and the related code path: https://github.com/Workiva/over_react/blob/master/tools/analyzer_plugin/lib/src/assist/toggle_stateful.dart#L49-L52

since for old boilerplate, the assist will short-circuit by returning false in `initializeAssistApi`, since `ComponentDeclarationAssistApi` doesn't support legacy boilerplate

joebingham-wk: Ohh, duhh, haha. Good call, adding that in the other branch!

PR #583, opened 2020-06-28
Fix unintended removal of keys in granchildren, add removeProp util

### LINKED ISSUE #545: Remove cascade parens when removing last prop in assists/fixes
For fixes/assists that remove props, instead of:
```diff
-(Dom.div()..key = 'foo')()
+(Dom.div())()
```

it would be nice to remove the extra parentheses:
```diff
-(Dom.div()..key = 'foo')()
+Dom.div()()
```

This could be implemented as a shared change builder, similar to `addProp`.

greglittlefield-wf:
## Motivation
- `convertUsageListLiteralToVariadicChildren` was iterating through children recursively, and removing keys on non-children descendants
- I noticed when testing it out that it was pretty annoying that the  parentheses were left in place after the fixes were applied, so I spiked out a solution to #545 and ended up tacking it on since it was so straightforward. 
    We previously considered this post-MVP, but I think it does a lot to improve the UX of our plugin, so it seemed important enough to include.

## Changes
- Only remove keys from children
- Add removeProp utility that removes parens when the last cascade is removed (fixes #545)
    - Update all other prop removals to use this

### DIFF
--- a/tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
+++ b/tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
@@ -44,10 +44,9 @@ class DuplicatePropCascadeDiagnostic extends ComponentUsageDiagnosticContributor
           fixKind: fixKind,
           computeFix: () => buildFileEdit(result, (builder) {
             for (var i = 0; i < propUsages.length - 1; i++) {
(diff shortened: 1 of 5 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

greglittlefield-wf:
QA +1

PR #606, opened 2020-07-08
CPLAT-3873 Function Components

sydneyjodon-wk:
## Motivation
  
Now that we've migrated to the new boilerplate, we need to implement function components so they can be used in `over_react`.
## Changes
  
- Add function component supporting APIs (`uiFunction`, `PropsFactory`) (implemented in [Greg's spike branch](https://github.com/Workiva/over_react/blob/new_boilerplate_spike/cleanup/example/builder/function_component.dart))
- Update parsing logic to validate function components.
- Update codegen for functional components.
    - Generate props config for each factory.
- Write tests
   - component integration tests
   - parsing tests (validate that the correct errors are raised by the builder when there is something wrong with the boilerplate)
- Update documentation
   - Show how to declare in README, and `uiFunction`
- Add snippet
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] Run `pbr serve example` and verify that function components work as expected on http://localhost:8080/builder/
      - [ ] Add [code snippets](https://github.com/Workiva/over_react/blob/3a2daeb4f015d85cc1f112fd31cb7061677f9fca/snippets/README.md) to IDE and verify that they work as expected.

## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/builder/parsing/declarations_from_members.dart
```diff
@@ -209,6 +210,12 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
     ..sort((a, b) => b.versionConfidences.maxConfidence.confidence
         .compareTo(a.versionConfidences.maxConfidence.confidence));
   for (final factory in factoriesMostToLeastConfidence) {
+    if(factory.versionConfidences.maxConfidence.version == Version.v5_functionComponent) {
```
greglittlefield-wf, line 213: Do we need a v5? We already have logic in place that yields a `PropsMapViewOrFunctionComponentDeclaration`; could we just use that and build in some factory grouping logic beforehand?

sydneyjodon-wk: That's true, I didn't think about how I could just check if `uiFunctionComponent` was called here instead of when the members are detected. That would make it a lot simpler!

And I made a `FunctionComponentDeclaration` class separately from `PropsMapViewOrFunctionComponentDeclaration` because I wanted `factory` to be a list of factories, but I guess I could just change the props map view logic to look for a single factory in the list.

Thanks for the feedback!

greglittlefield-wf: Sweet, sounds good. Np!

doc/new_boilerplate_migration.md
```diff
@@ -727,19 +727,19 @@ import 'package:over_react/over_react.dart';
 
 part 'foo.over_react.g.dart';
 
-UiFactory<FooProps> Foo = uiFunctionComponent(
+UiFactory<FooProps> Foo = uiFunction(
```
sydneyjodon-wk, line 730: The [RFD](https://sandbox.wdesk.com/a/QWNjb3VudB82NzE5NTMwMjcyOTQ4MjI0/doc/e601ef61809a432eb36f1b358c20a06d/r/-1/v/1/sec/e601ef61809a432eb36f1b358c20a06d_196) used both names for this function. `uiFunction` was what the ticket said to name it, but if anyone thinks it should be `uiFunctionComponent` I can update it.

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+///
+/// __Example__:
+/// ```dart
+/// UiFactory<FooProps> Foo = uiFunction((props) {
+///     // Set default props using null-aware operators.
+///     final isDisabled = props.isDisabled ?? false;
+///     final items = props.items ?? [];
+///
+///     // Return the rendered component contents here.
+///     return Dom.div()(items);
```
aaronlademann-wf, line 43: Should we demonstrate prop forwarding at all?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+///
+/// __OR__ Optionally pass in an existing [PropsFactory] in place of a props [config].
+///
+/// ```dart
+/// UiFactory<FooProps> Bar = uiFunction((props) {
+///     return Dom.div()(props.items);
+///   },
+///   null,
+///   propsFactory: PropsFactory.fromUiFactory(Foo),
+///   displayName: 'Bar',
```
aaronlademann-wf, line 63: Is the `displayName` argument necessary?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+/// UiFactory<UiProps> Baz = uiFunction((props) {
+///     return Dom.div()('prop id: ${props.id}');
+///   },
+///   null,
+///   displayName: 'Baz',
+/// );
+/// ```
+///
+/// Learn more: <https://reactjs.org/docs/components-and-props.html#function-and-class-components>.
+// FIXME right now only top level factory declarations will generate props configs.
```
aaronlademann-wf, line 79: Is this still valid?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+UiFactory<TProps> uiFunction<TProps extends UiProps>(
+  UiFunctionComponent<TProps> functionComponent,
+  FunctionComponentConfig<TProps> config, {
```
aaronlademann-wf, line 82: Since these are generated, it feels like `FunctionComponentConfig` should either be private, or `$$` prefixed or something to indicate its not intended for public consumption _(and subsequently not exported)_

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+/// UiFactory<FooProps> Foo = uiFunction((props) {
+///     // Set default props using null-aware operators.
+///     final isDisabled = props.isDisabled ?? false;
+///     final items = props.items ?? [];
+///
+///     // Return the rendered component contents here.
+///     return Dom.div()(items);
+///   },
+///   // The generated props config will match the factory name.
+///   $FooPropsConfig,
```
aaronlademann-wf, line 46: We should add `displayName` if that is the only way to make it so that it shows up as `Foo` in the dev tools.

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+  if (config != null) {
+    if (propsFactory != null) {
+      throw ArgumentError('propsFactory cannot be used along with config');
+    }
+    propsFactory = config.propsFactory;
+    displayName ??= config.componentName;
+  }
+
+  // Get the display name from the inner function if possible so it doesn't become `_uiFunctionWrapper`
+  displayName ??= _getFunctionName(functionComponent);
```
aaronlademann-wf, line 95: What would this look like? All of our examples show an anonymous function closure for the `functionComponent` argument... Does this mean the var name that `uiFunction` "wrapper" is assigned to is not what displayName will be by default?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+
+  dynamic _uiFunctionWrapper(Map props) {
+    return functionComponent(propsFactory.jsMap(props as JsBackedMap));
```
aaronlademann-wf, line 98: Is there any reason to not type `props` as `JsBackedMap` instead of casting?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  ReactDartFunctionComponentFactoryProxy factory =
```
aaronlademann-wf, line 101: #nit this could be final

```suggestion
  final factory =
```

Will probably result in nicer formatting as well.

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+      react.registerFunctionComponent(
+    _uiFunctionWrapper,
+    displayName: displayName,
+  );
+
+  if (propsFactory == null) {
+    if (TProps != UiProps && TProps != _GenericUiProps) {
+      throw ArgumentError(
+          'config.propsFactory must be provided when using custom props classes');
```
aaronlademann-wf, line 110:
```suggestion
          'config.propsFactory must be provided when using a custom props mixin');
```

lib/src/component_declaration/function_component.dart
aaronlademann-wf: And if not, is this a safe cast to make if the map being passed in is not already a `JsBackedMap` instance?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  ReactDartFunctionComponentFactoryProxy factory =
+      react.registerFunctionComponent(
+    _uiFunctionWrapper,
+    displayName: displayName,
+  );
+
+  if (propsFactory == null) {
+    if (TProps != UiProps && TProps != _GenericUiProps) {
```
aaronlademann-wf, line 108: Are these not type comparisons? E.g. should this be `is!` instead of `!=`?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+  @override
+  bool get $isClassGenerated => true;
+}
+
+/// Helper class used to keep track of generated information for [uiFunction].
+@protected
+class FunctionComponentConfig<TProps extends UiProps> {
+  @protected
+  final PropsFactory<TProps> propsFactory;
+  final String componentName;
```
aaronlademann-wf, line 158: How is this different from `displayName`? Or can the builder not always generate it?

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+
+/// Helper class to keep track of props factories used by [uiFunction],
+class PropsFactory<TProps extends UiProps> {
+  /// Returns a typed props object backed by the specified [Map].
+  final TProps Function(Map props) map;
+
+  /// Returns a typed props object backed by the specified [JsBackedMap].
+  final TProps Function(JsBackedMap props) jsMap;
+
+  PropsFactory({
```
aaronlademann-wf, line 172: Should this constructor be public? Is there a use case for an instance of this that is not created using the `fromUiFactory` factory constructor?

pubspec.yaml
```diff
@@ -47,3 +47,9 @@ dev_dependencies:
   pedantic: ^1.8.0
   test: ^1.9.1
   yaml: ^2.2.1
+
+dependency_overrides:
+  react:
+    git:
+      url: https://github.com/cleandart/react-dart.git
+      ref: 5.4.0-wip
```
aaronlademann-wf, line 55: This should be updated to 5.5.0-wip when you have a chance.

lib/src/builder/parsing/declarations_from_members.dart
```diff
@@ -209,6 +212,12 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
     ..sort((a, b) => b.versionConfidences.maxConfidence.confidence
         .compareTo(a.versionConfidences.maxConfidence.confidence));
   for (final factory in factoriesMostToLeastConfidence) {
+    if (factory.isFunctionComponentFactory) {
+      _functionComponentFactories.add(factory);
+      // will be validated below the for-loop.
```
greglittlefield-wf, line 217: Is there a way we could combine the below for-loop into this one? It would be nice if all of the grouping logic was in one place, and make it easier to tell what grouping logic takes precedence.

I had some old logic that grouped factories together: https://github.com/Workiva/over_react/commit/f62af3fb005531b3576d9fd707db52ff9e52a7c8

So maybe parts of that could be used here. M Perhaps something like

```diff    

-  final factoriesMostToLeastConfidence = List.of(members.factories)
+  final factoriesMostToLeastConfidence = factoryGroups.map((group) => group.bestFactory).toList()
     ..sort((a, b) => b.versionConfidences.maxConfidence.confidence
         .compareTo(a.versionConfidences.maxConfidence.confidence));
   for (final factory in factoriesMostToLeastConfidence) {
...

     final component = getComponentFor(factory, members.components);
     if (component != null && hasBeenConsumed(component)) {
       continue;
     }

+    if (factory.isFunctionComponentFactory) {
+      final factories = factoryGroups.firstWhere((group) => group.bestFactory == factory);
+      yield PropsMapViewOrFunctionComponentDeclaration(factories: factories)
+    }
```

lib/src/builder/parsing/members/factory.dart
```diff
@@ -45,6 +45,10 @@ class BoilerplateFactory extends BoilerplateMember {
 
   bool get hasFactoryAnnotation => node.hasAnnotationWithName('Factory');
 
+  bool get isFunctionComponentFactory =>
+      node.variables.firstInitializer is MethodInvocation &&
+      (node.variables.firstInitializer as MethodInvocation).methodName.name == 'uiFunction';
```
greglittlefield-wf, line 50: FYI `tryCast` can be used to avoid extra casting these checks a little more compact
```suggestion
      node.variables.firstInitializer?.tryCast<MethodInvocation>()?.methodName?.name == 'uiFunction';
```

greglittlefield-wf, line 50: See note below about nested uiFunction calls; this will need to be updated

lib/src/builder/parsing/declarations_from_members.dart
```diff
@@ -360,6 +402,25 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
...
+// Returns whether or not the props config argument of `uiFunction` is `null`.
+bool _hasConfigArg(BoilerplateFactory factory) {
+  final rightHandSide = factory.node.variables.firstInitializer;
+  final args = (rightHandSide as MethodInvocation).argumentList.arguments;
+  if (args == null || args.length < 2) return false;
+  return args[1] is! NullLiteral;
+}
+
+// Returns the name of the props for [factory] based on typing arguments in the `uiFunction` declaration.
+String _getPropsNameFromFunctionComponent(BoilerplateFactory factory) {
```
greglittlefield-wf, line 414: This should be a getter, similar to `BoilerplateComponent.propsGenericArg`

lib/src/builder/parsing/members_from_ast.dart
```diff
@@ -168,6 +168,18 @@ class _BoilerplateMemberDetector {
       return;
     }
 
+    final rightHandSide = node.variables.firstInitializer;
+    if (rightHandSide is MethodInvocation && rightHandSide.methodName.name == 'uiFunction') {
```
greglittlefield-wf, line 172: Also, I think we should look for nested calls to this, to support wrapping these in HOCs inline. For example:

```dart
UiFactory<FooProps> Foo = connect(
 ...
)(uiFunction({
});
```

We might want to swap this out for 
```dart
anyDescendantIdentifiers(rightHandSide, (identifier) => identifier.name == 'uiFunction');
```

Can we also add a test case for that?

lib/src/component_declaration/function_component.dart
greglittlefield-wf, line 79: Is this being addressed in the PR? If not, this should be downgraded to a TODO

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+
+export 'component_type_checking.dart'
+    show isComponentOfType, isValidElementOfType;
+
+/// A React component declared using a function that takes in [props] and returns rendered output.
+///
+/// See <https://facebook.github.io/react/docs/components-and-props.html#functional-and-class-components>.
+typedef dynamic UiFunctionComponent<T extends UiProps>(T props);
+
+/// Returns a `UiFactory<TProps>` for [functionComponent].
```
greglittlefield-wf, line 33: We don't need to mention what the return type is here (`UiFactory<TProps>`) https://dart.dev/guides/language/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context

I think we could include a little more context on what the purpose of this function is, too, for people seeing it for the first time:
```suggestion
/// Declares a function component and returns a factory that can be used to render it.
```

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+///
+/// __Example__:
+/// ```dart
+/// UiFactory<FooProps> Foo = uiFunction((props) {
```
greglittlefield-wf, line 37: Formatting
```suggestion
/// UiFactory<FooProps> Foo = uiFunction(
///   (props) {
```

lib/src/component_declaration/function_component.dart
greglittlefield-wf, line 46: This should have an ignore comment

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+
+  // Get the display name from the inner function if possible so it doesn't become `_uiFunctionWrapper`
+  displayName ??= _getFunctionName(functionComponent);
+
+  dynamic _uiFunctionWrapper(Map props) {
+    return functionComponent(propsFactory.jsMap(props as JsBackedMap));
+  }
+
+  ReactDartFunctionComponentFactoryProxy factory =
+      react.registerFunctionComponent(
```
greglittlefield-wf, line 102: #nit prefer omitting LHS typing for local variables
```suggestion
  final factory = react.registerFunctionComponent(
```

lib/src/builder/parsing/declarations_from_members.dart
```diff
@@ -280,13 +289,46 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
...
+      members.propsMixins,
+    ].expand((i) => i).whereNot(hasBeenConsumed);
+
+    final standAloneFactories = _functionComponentFactories.where((factory) =>
+        _getPropsNameFromFunctionComponent(factory) == 'UiProps' || !_hasConfigArg(factory));
+    for (final factory in standAloneFactories) {
+      consume(factory);
+      yield PropsMapViewOrFunctionComponentDeclaration(
+        factories: [factory],
+        props: null,
```
greglittlefield-wf, line 311: If we don't need to generate anything for these cases, can we instead not `yield` them? That's what we do for other declarations that don't need generation.

Also, for other declarations, we use a confidence of `neutral` so that the line
```dart
if (!resolveVersion(nonNullFactoryPropsOrComponents).shouldGenerate) continue;
```
below bails out of generation, as opposed to explicitly consuming them.

Can we update the logic around whether a factory is standalone into its initial confidence computation within `members_from_ast.dart`, so that it's consistent with how other members are handled?

test/vm_tests/builder/declaration_parsing_test.dart
```diff
@@ -1538,7 +1538,8 @@ main() {
                 ]));
                 final decl = declarations.firstWhereType<PropsMapViewOrFunctionComponentDeclaration>();
 
-                expect(decl.factory?.name?.name, name);
+                expect(decl.factories.length, 1);
```
greglittlefield-wf, line 1541: Prefer using `hasLength`, which gives better failure messages
```suggestion
                expect(decl.factories, hasLength(1));
```

test/vm_tests/builder/declaration_parsing_test.dart
```diff
@@ -1550,7 +1551,8 @@ main() {
                 ''');
                 final decl = expectSingleOfType<PropsMapViewOrFunctionComponentDeclaration>(declarations);
 
-                expect(decl.factory?.name?.name, name);
+                expect(decl.factories.length, 1);
```
greglittlefield-wf, line 1554: Prefer using `hasLength`, which gives better failure messages
```suggestion
                expect(decl.factories, hasLength(1));
```

lib/src/component_declaration/function_component.dart
greglittlefield-wf: These are comparisons of two types, not a type and an instance, so you can't use  `is`/`is!`

lib/src/builder/parsing/declarations.dart
```diff
@@ -271,15 +271,15 @@ class ClassComponentDeclaration extends BoilerplateDeclaration
...
+  final List<BoilerplateFactory> factories;
 
   /// The related props instance.
   ///
   /// Can be either [BoilerplateProps] or [BoilerplatePropsMixin], but not both.
   final Union<BoilerplateProps, BoilerplatePropsMixin> props;
 
   @override
-  get _members => [factory, props.either];
+  get _members => props == null ? factories : [...factories, props.either];
```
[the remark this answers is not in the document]
greglittlefield-wf: I believe it's for cases  where it shouldn't be generated. I commented that we can just not `yield` anthing in that case here https://github.com/Workiva/over_react/pull/606#discussion_r458997663, which means we can probably go back to making props non-nullable

lib/src/builder/parsing/declarations_from_members.dart
```diff
@@ -12,6 +12,8 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
+import 'package:analyzer/dart/ast/ast.dart';
```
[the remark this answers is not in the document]
greglittlefield-wf: I don't really think that's necessary.
1. anything trying to use fully resolved AST just won't work, since we're not dealing with fully resolved AST
2. we don't do that anywhere else where we import this

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+import 'package:react/react_client.dart';
+import 'package:react/react_client/js_backed_map.dart';
+
+export 'component_type_checking.dart'
+    show isComponentOfType, isValidElementOfType;
+
+/// A React component declared using a function that takes in [props] and returns rendered output.
+///
+/// See <https://facebook.github.io/react/docs/components-and-props.html#functional-and-class-components>.
+typedef dynamic UiFunctionComponent<T extends UiProps>(T props);
```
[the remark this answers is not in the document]
greglittlefield-wf: Mm, yeah, I feel like we should just inline this since it's currently only used in one place, and I don't think there are a lot of use-cases fo rit

lib/src/component_declaration/function_component.dart
greglittlefield-wf: It's not; `$FooPropsConfig` contains that information

lib/src/component_declaration/function_component.dart
greglittlefield-wf: This is due to the typing of the arg to `registerFunctionComponent`, which is `dynamic Function(Map props)`. Passing in a `dynamic Function(JsBackedMap props)` would be a type error.

Hmmm yeah maybe we should update that typing in react-dart after all...

lib/src/component_declaration/function_component.dart
greglittlefield-wf: This could be a props class or a props mixin; I think "props classes" includes both

lib/src/component_declaration/function_component.dart
greglittlefield-wf: It's not different; we should probably just rename it to `displayName`

lib/src/component_declaration/function_component.dart
greglittlefield-wf: This has applications for making over_react JS components wrappers

lib/src/component_declaration/function_component.dart
greglittlefield-wf: I'm not sure I agree. I think this is similar to `PropsMeta`; most of the time it's generated, but there's no reason to prevent people from consuming it.

aaronlademann-wf: How would someone create their own?

doc/new_boilerplate_migration.md
```diff
@@ -803,14 +803,14 @@ UiFactory<FooProps> createFooHoc(UiFactory otherFactory) {
   Object closureVariable; 
   // ... 
 
-  final FooHoc = uiFunctionComponent<FooProps>(
+  final FooHoc = uiFunction<FooProps>(
```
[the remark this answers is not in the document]
greglittlefield-wf: We basically decided:
- it was better to keep it consistent with class component factories
- it avoids triggering the type_annotate_public_apis lint

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,181 @@
...
+/// ```
+///
+/// __OR__ Optionally pass in an existing [PropsFactory] in place of a props [config].
+///
+/// ```dart
+/// UiFactory<FooProps> Bar = uiFunction((props) {
+///     return Dom.div()(props.items);
+///   },
+///   null,
+///   propsFactory: PropsFactory.fromUiFactory(Foo),
```
greglittlefield-wf, line 62: I think we should also have an example of using an existing generated PropsConfig and a different display name, to show what declaring multiple function components with the same props looks like.

lib/src/component_declaration/function_component.dart
greglittlefield-wf: Yes. Without it, it would have nowhere to get a name from

lib/src/component_declaration/function_component.dart
greglittlefield-wf: Not sure I follow; what would prevent them from creating their own?

doc/new_boilerplate_migration.md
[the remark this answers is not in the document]
sydneyjodon-wk: I didn't add any examples of it, but declaring a top level function component without LHS typing does work. So should I change it so that they can't be declared without LHS typing?

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: Yes, when there's no generated props config, `displayName` has to be set here

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: I can add an example of multiple function components with the same props. I made it so a PropsConfig is generated for each function component so that the display name would be automatically set correctly:
```
mixin FooProps on UiProps {}

UiFactory<FooProps> Foo = uiFunction(
  (props) => Dom.div()(),
  $FooPropsConfig,
);

UiFactory<FooProps> Bar = uiFunction(
  (props) => Dom.div()(),
  $BarPropsConfig,
);
```
Does that make sense, or do you think I should change it so that one config is generated for each props mixin and then the display name has to be set on multiple function components? Or I could change the config names to `$FooConfig` and `$BarConfig` to associate them more with the factory than the props?

doc/new_boilerplate_migration.md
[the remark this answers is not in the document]
greglittlefield-wf: Oh nice! Hmmm I guess if there's not a reason to disallow it, we shouldn't disallow it.

Are there tests for declaring them without LHS typing?

lib/src/component_declaration/function_component.dart
greglittlefield-wf: Ooh nice, no, I like what you have a lot!

doc/new_boilerplate_migration.md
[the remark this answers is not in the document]
sydneyjodon-wk: The parsing and codegen tests have with/without LHS typing declarations sprinkled throughout, but I'll add one to the integration tests as well

example/builder/src/function_component.dart
```diff
@@ -0,0 +1,79 @@
...
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:over_react/over_react.dart';
+import 'package:react/react_client.dart';
+import 'package:react/react_client/react_interop.dart';
```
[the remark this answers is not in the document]
sydneyjodon-wk: Nope! They're not necessary - I'll remove them

lib/src/builder/codegen/typed_map_impl_generator.dart
```diff
@@ -46,7 +46,7 @@ abstract class TypedMapImplGenerator extends BoilerplateDeclarationGenerator {
 
   TypedMapNames get names;
   bool get isComponent2;
-  FactoryNames get factoryNames;
+  List<FactoryNames> get factoryNames;
```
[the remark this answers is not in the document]
sydneyjodon-wk: We need to generate a props config for every function component factory (even if they share the same props mixin) so that the display name is set up correctly. This has to be a list in order to get all the factory names from `PropsMapViewOrFunctionComponentDeclaration.factories`. In all other cases `factoryNames` would have a single element -- I just added an assert for that in `_generateFactory`.

lib/src/builder/parsing/declarations_from_members.dart
sydneyjodon-wk: Oh that makes sense! I'll update that!

lib/src/component_declaration/function_component.dart
aaronlademann-wf: How / why would you create an instance of a config like this when it is something that our builder generates?

greglittlefield-wf: How: just instantiate it 
Why: If you want to use a `uiFunction` that doesn't have a generated builder

It's definitely an edge-case that I can't think of any concrete use-cases for, but I also see no reason to prohibit it.

On a related note: we have this config, and then two arguments that map to the same fields as the config. Perhaps we should just combine those?

Existing usages could be replaced like so:

```diff
-uiFunction(() ..., null, displayName: 'Foo');
+uiFunction(() ..., GenericUiPropsConfig(displayName: 'Foo'));
```
```diff
-uiFunction(
-  () ..., 
-  null, 
-  propsFactory: PropsFactory.fromUiFactory(Bar),
-  displayName: 'Foo',
);
+uiFunction(
+  () ..., 
+  FunctionComponentConfig(
+    propsFactory: PropsFactory.fromUiFactory(Bar),
+    displayName: 'Foo',
+  ),
+);
```

Then, we would have a legitimate use for people doing this.

Thoughts, @sydneyjodon-wk  / @aaronlademann-wf ?

lib/src/component_declaration/function_component.dart
greglittlefield-wf: Oh wait, I misread that last part.

> Or I could change the config names to `$FooConfig` and `$BarConfig` to associate them more with the factory than the props?

I think that makes sense, so that the naming aligns more with `FunctionComponentConfig` (it's not `FunctionComponentPropsConfig`) and doesn't get falsely associated with a props class of a given name (e.g., `BarProps` in `$BarPropsConfig`).

sydneyjodon-wk: Cool I'll update that!

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: Renamed this in cec79b9

README.md
```diff
@@ -846,6 +846,30 @@ that you get for free from OverReact, you're ready to start building your own cu
...
+  UiFactory<FooProps> Foo = uiFunction((props) {
+    // Set default props using null-aware operators.
+    final isDisabled = props.isDisabled ?? false;
+    final items = props.items ?? [];
+  
+    // Return the rendered component contents here.
+    // The `props` variable is typed; no need for string keys!
+  },
+  // The generated props config will match the factory name.
+  $FooPropsConfig);
```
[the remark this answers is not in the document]
sydneyjodon-wk: Added ignore comments and updated formatting in cec79b9

lib/src/builder/parsing/declarations_from_members.dart
sydneyjodon-wk: That's so much nicer!

lib/src/component_declaration/function_component.dart
aaronlademann-wf: Not a fan of requiring them to use the `FunctionComponentConfig` to set a `displayName`.

sydneyjodon-wk: I agree with Aaron and the `UiProps` case seems a little over complicated.

But, now that I think about it, we could generate a config for every function component so that the display name will be set automatically and just not set `propsFactory` in the config if `propsFactory` is already set on `uiFunction` or if `UiProps` is used because those are taken care of inside `uiFunction`. Thoughts?

Then we could change the name to `$$FunctionComponentConfig` or something like Aaron was saying before since the config would always be generated.

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: No, this is not being addressed. Currently only non-top level factory declarations only work if no codegen is needed. I'll update this to a todo

lib/src/component_declaration/function_component.dart
greglittlefield-wf: > Not a fan of requiring them to use the FunctionComponentConfig to set a displayName.
That's fair, but keep in mind this is a bit of an edge case already.

> I agree with Aaron and the UiProps case seems a little over complicated.
That's fair; my thought was that it'd a little more straightforward than passing in `null`.

> we could generate a config for every function component 
Don't we already do that though for generated function components?

This case is for non-generated function components, many of which may be local variables, and I don't think we should generate for those.

lib/src/component_declaration/function_component.dart
greglittlefield-wf: > Does this mean the var name that uiFunction "wrapper" is assigned to is not what displayName will be by default?
No, that is the default; it's pulled from `config`, which is generated by the default

This is for a case where someone has a uiFunction (no generated config) and doesn't specify the displayName
```dart
// displayName will be something like: withFoo_closure
withFoo(...) {
  final factoryWithFoo = uiFunction<UiProps>(() {
    ...
  }, null);
}
```
```dart
// displayName will be: withFooFunctionComponent
withFoo(...) {
  withFooFunctionComponent() {
    ...
  }
  final factoryWithFoo = uiFunction<UiProps>(withFooFunctionComponent, null);
}
```

lib/src/builder/parsing/declarations.dart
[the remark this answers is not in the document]
sydneyjodon-wk: Yep! I removed that now that these cases aren't yielded anymore

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: Yes, function components that aren't top level declarations will only work if no props config needs to be generated (i.e. for `UiProps` or custom `propsFactory`). I changed this to a todo

lib/src/builder/parsing/declarations_from_members.dart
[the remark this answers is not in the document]
sydneyjodon-wk: Because of other changes, this import is no longer needed

lib/src/builder/parsing/declarations_from_members.dart
sydneyjodon-wk: Updated it! That factory grouping is really nice and it fixed some of the other parsing problems I was having too!

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: So should I update this to have propsFactory and displayName set in a `FunctionComponentConfig` or just leave `FunctionComponentConfig` as an optional way to set them?

greglittlefield-wf: I would lean toward FunctionComponentConfig and get rid of the args, but I'd be curious to hear other people's opinions. Maybe we should ask in Slack?

sydneyjodon-wk: I'll ask in slack

greglittlefield-wf: Thank you!

lib/src/builder/parsing/ast_util.dart
```diff
@@ -41,6 +41,22 @@ extension InitializerHelperTopLevel on TopLevelVariableDeclaration {
 
   /// The first variable in this list.
   VariableDeclaration get firstVariable => variables.firstVariable;
+
+  /// Returns whether or not the config argument of `uiFunction` is `null`.
+  ///
+  /// Returns `false` if [firstInitializer] is not a method invocation.
```
greglittlefield-wf, line 47: This line of the doc comment isn't accurate and could use be removed

lib/src/builder/codegen.dart
```diff
@@ -35,9 +35,10 @@ class ImplGenerator {
   void generate(BoilerplateDeclaration declaration) {
     switch (declaration.type) {
       case DeclarationType.propsMapViewOrFunctionComponentDeclaration:
-        _generatePropsMapViewOrFunctionComponent(declaration);
-        break;
-      case DeclarationType.genericFunctionComponentDeclaration:
+        // Generated code only needed for props config.
+        if ((declaration as PropsMapViewOrFunctionComponentDeclaration).props != null) {
```
greglittlefield-wf, line 39: We should be able to get rid of this check now that we're not yielding functions that don't need generation, right?

lib/src/builder/parsing/declarations_from_members.dart
```diff
@@ -248,6 +243,27 @@ Iterable<BoilerplateDeclaration> getBoilerplateDeclarations(
...
     }
 
+    if (factory.isFunctionComponentFactory) {
+      final factories = factoryGroups.firstWhere((group) => group.bestFactory == factory);
+      final generatedFactories = factories.factories.where((factory) =>
+          factory.node.hasConfigArg &&
+          factory.isFunctionComponentFactory &&
+          factory.versionConfidences.maxConfidence.shouldGenerate);
+      final associatedProps =
+          getPropsForFunctionComponent(members.props, members.propsMixins, factory);
```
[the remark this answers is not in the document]
sydneyjodon-wk: I had to make `getPropsForFunctionComponent` because `getPropsFor` normalizes the naming so if the names don't match it doesn't find the props class

lib/src/builder/parsing/members/factory.dart
```diff
@@ -38,6 +38,12 @@ class BoilerplateFactory extends BoilerplateMember {
       }
     }
 
+    if (isFunctionComponentFactory) {
+      final rightHandSide = node.variables.firstInitializer;
+      final typeArgs = rightHandSide.tryCast<MethodInvocation>()?.typeArguments?.arguments?.first;
```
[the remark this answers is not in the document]
sydneyjodon-wk: oh you are so right. I'll update that!

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: I updated this!

lib/src/component_declaration/function_component.dart
```diff
@@ -74,32 +81,28 @@ export 'component_type_checking.dart'
...
-  if (config != null) {
-    if (propsFactory != null) {
-      throw ArgumentError('propsFactory cannot be used along with config');
-    }
-    propsFactory = config.propsFactory;
-    displayName ??= config.displayName;
+  FunctionComponentConfig<TProps> config) {
+  if (config == null) {
+    throw ArgumentError('config must not be null');
   }
```
greglittlefield-wf: #nit There's a `.notNull` constructor that should be used
```dart
  if (config == null) {
    throw ArgumentError.notNull('config');
  }
```
There's also a convenience method for it!
```suggestion
  ArgumentError.checkNotNull(config, 'config');
```

lib/src/component_declaration/function_component.dart
```diff
@@ -0,0 +1,188 @@
...
+
+  @override
+  String get propKeyNamespace => '';
+
+  @override
+  bool get $isClassGenerated => true;
+}
+
+/// Helper class used to keep track of generated information for [uiFunction].
+@protected
```
greglittlefield-wf, line 161: Now that we accept it as an argument to uiFunction, this class and its members shouldn't be `@protected` anymore

example/builder/src/function_component.dart
```diff
@@ -0,0 +1,112 @@
...
+mixin BasicProps on UiProps {
+  Ref forwardedRef;
+  String basicProp;
+  String basic1;
+  String basic2;
+  String basic3;
+  String basic4;
+}
+
+UiFactory<BasicProps> Basic = forwardRef<BasicProps>((props, ref) {
```
greglittlefield-wf, line 29: Can we add a comment saying that this will change in the future?

doc/new_boilerplate_migration.md
```diff
@@ -803,20 +816,52 @@ UiFactory<FooProps> createFooHoc(UiFactory otherFactory) {
...
+    return Fragment()(
+      Dom.div()('Some text.'),
+      (Dom.button()
+        ..ref = props.forwardedRef 
+        ..onClick = props.doSomething
+      )('Click me!'),
+    );
+  },
+  $_FooConfig, // ignore: undefined_identifier
+);
```
greglittlefield-wf, line 862: Can we update this to utilize the new syntax that will be added in https://jira.atl.workiva.net/browse/CPLAT-11722 ?
```suggestion
UiFactory<FooProps> Foo = uiForwardRef(
  (props, ref) {
    return Fragment()(
      Dom.div()('Some text.'),
      (Dom.button()
        ..ref = ref
        ..onClick = props.doSomething
      )('Click me!'),
    );
  },
  $_FooConfig, // ignore: undefined_identifier
);
```

lib/src/component_declaration/function_component.dart
sydneyjodon-wk: ooh that's really nice!

PR #607, opened 2020-07-15
[Analyzer Plugin] DuplicatePropCascadeDiagnostic should not flag props from different mixins as dupes

aaronlademann-wf:
## Motivation
Props with the same name from different mixins were being linted as duplicates.

## Changes
1. Use the prop namespace as part of the equality comparison unless the prop comes from `DomPropsMixin`.
2. Add a reproducing case to the playground.

#### Release Notes
Fix false negative over_react analyzer plugin lint when two props share the same names but come from different mixins.

## Review

  

FYI @brettmeyer-wk @dustyholmes-wf 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

tools/analyzer_plugin/lib/src/fluent_interface_util/cascade_read.dart
```diff
@@ -48,6 +48,9 @@ class PropAssignment {
   /// ```
   Identifier get targetName => leftHandSide.target?.tryCast<PropertyAccess>()?.propertyName;
 
+  /// Returns the name of the mixin the prop lives within that serves as its "namespace".
+  String get namespace => leftHandSide.propertyName.staticElement.enclosingElement.name;
```
greglittlefield-wf, line 52: This name be misleading, since the mixin may have a custom namespace. I struggled coming up with a good name for it that's generalizable, and it's not used anywhere else, so I feel like we should just inline it

Also, this should have null-awares after `staticElement` since it may not be resolved

tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
```diff
@@ -31,6 +31,10 @@ class DuplicatePropCascadeDiagnostic extends ComponentUsageDiagnosticContributor
         return '${prop.targetName.name}.${prop.name.name}';
       }
 
+      if (prop.namespace != 'DomPropsMixin') {
+        return '${prop.namespace}.${prop.name.name}';
+      }
+
       return prop.name.name;
     });
```
greglittlefield-wf, line 39: I think a more straightforward solution would be to just always prefix with the type; then we don't have to worry about whether it's in a nested `.dom`.

```dart
final propUsagesByName = groupBy<PropAssignment, String>(usage.cascadedProps, (prop) {
  return '${prop.namespace}.${prop.name.name}';
});

```

aaronlademann-wf:
@greglittlefield-wf I addressed your feedback, but the commit is not showing up here for some reason: https://github.com/Workiva/over_react/commit/4ae5cb8bf1aa39a7d8d3402bb7aec84b0afcee04

tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
```diff
@@ -27,11 +27,13 @@ class DuplicatePropCascadeDiagnostic extends ComponentUsageDiagnosticContributor
   @override
   computeErrorsForUsage(result, collector, usage) async {
     final propUsagesByName = groupBy<PropAssignment, String>(usage.cascadedProps, (prop) {
+      final enclosingElement = prop.name.staticElement?.enclosingElement?.name ?? '';
+
       if (prop.targetName != null && prop.targetName.name != 'dom') {
-        return '${prop.targetName.name}.${prop.name.name}';
+        return '$enclosingElement.${prop.targetName.name}.${prop.name.name}';
       }
```
greglittlefield-wf, line 34: Do we need this check now, or can we remove this whole if statement?

tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
```diff
@@ -27,11 +27,13 @@ class DuplicatePropCascadeDiagnostic extends ComponentUsageDiagnosticContributor
   @override
   computeErrorsForUsage(result, collector, usage) async {
     final propUsagesByName = groupBy<PropAssignment, String>(usage.cascadedProps, (prop) {
+      final enclosingElement = prop.name.staticElement?.enclosingElement?.name ?? '';
```
greglittlefield-wf, line 30: Sorry, I just realized this could use a comment:

Also #nit on naming
```suggestion
      // Prefix with the class the prop was declared in so that things like `.dom.type` don't conflict with an unrelated `type` prop.
      final enclosingElementName = prop.name.staticElement?.enclosingElement?.name ?? '';
```

tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
aaronlademann-wf: I'm pretty sure we still need it - the `targetName` is different than the `enclosingElement`.

Without this check - the following props would be flagged as dupe since they both have an `enclosingElement` name of `AriaPropsMixin`:

```dart
(Dom.div()
  ..hidden = true
  ..aria.hidden = true
)()
```

tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
```diff
@@ -27,11 +27,14 @@ class DuplicatePropCascadeDiagnostic extends ComponentUsageDiagnosticContributor
   @override
   computeErrorsForUsage(result, collector, usage) async {
     final propUsagesByName = groupBy<PropAssignment, String>(usage.cascadedProps, (prop) {
+      // Prefix with the class the prop was declared in so that things like `.dom.type` don't conflict with an unrelated `type` prop.
+      final enclosingElementName = prop.name.staticElement?.enclosingElement?.name ?? '';
+
       if (prop.targetName != null && prop.targetName.name != 'dom') {
-        return '${prop.targetName.name}.${prop.name.name}';
+        return '$enclosingElement.${prop.targetName.name}.${prop.name.name}';
```
evanweible-wf, line 34: I think this should be `enclosingElementName`?

tools/analyzer_plugin/lib/src/diagnostic/duplicate_prop_cascade.dart
greglittlefield-wf: > Without this check - the following props would be flagged as dupe since they both have an enclosingElement name of AriaPropsMixin

Wouldn't we _want_ that case flagged as a dupe though?

aaronlademann-wf: No - they are different.

One results in

```html
<div hidden></div>
```

The other results in 

```html
<div aria-hidden="true"></div>
```

aaronlademann-wf: Oh wait... I think I see what you're saying now - hmmm... maybe we don't need it?

aaronlademann-wf:
CI should start passing once https://github.com/Workiva/over_react/pull/747 merges.

tools/analyzer_plugin/test/integration/diagnostics/duplicate_prop_cascade_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  ..hidden = false
+  ..aria.hidden = false
+)();
+''');
+    final allErrors = await getAllErrors(source);
+    expect(allErrors, hasLength(2));
+
+    for (final selection in [createSelection(source, "#..id# = '1'"), createSelection(source, "#..dom.id# = 'foo'")]) {
+      expect(allErrors, contains(isAnErrorUnderTest(locatedAt: selection, hasFix: true)));
+    }
```
greglittlefield-wf, line 67: #nit this could be simplified
```suggestion
    expect(allErrors, unnorderedEquals([
      isAnErrorUnderTest(locatedAt: createSelection(source, "#..id# = '1'"), hasFix: true)),
      isAnErrorUnderTest(locatedAt: createSelection(source, "#..dom.id# = 'foo'"), hasFix: true)),
    ]));
```

tools/analyzer_plugin/test/integration/diagnostics/duplicate_prop_cascade_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  },
+  _$CustomConfig, // ignore: undefined_identifier
+);
+
+final customUsage = (Custom()
+  ..size = 2
+  ..dom.size = 2
+  ..dom.hidden = false
+  ..aria.hidden = false
+  ..hidden = false
```
greglittlefield-wf, line 47: This test covers a lot of cases, but it might not be clear to future maintainers what this is testing and why. I think it'd be good to have a comment or a better test name that documents all of the cases we want to hit (e.g., props from same class but with a prefix, props that have the same name but aren't from the same class, etc.).

Having them as separate tests might be nice, but I'm also totally fine having them be all in one so long as the intent of the test is clear.

Same comment for the test below.

Also, I noticed that there isn't a test for the same prop being provided twice, without any prefix, which is the much more common case we want to make sure is handled. E.g.,
```dart
final foo = (Dom.div()
  ..id = '1'
  ..id = 'foo'
)();
```
Could we add a test case for that, and also one for more than two usages?

tools/analyzer_plugin/test/integration/diagnostics/duplicate_prop_cascade_test.dart
```diff
@@ -0,0 +1,69 @@
...
+  ..aria.hidden = false
+)();
+''');
+    final allErrors = await getAllErrors(source);
+    expect(allErrors, hasLength(2));
+
+    for (final selection in [createSelection(source, "#..id# = '1'"), createSelection(source, "#..dom.id# = 'foo'")]) {
+      expect(allErrors, contains(isAnErrorUnderTest(locatedAt: selection, hasFix: true)));
+    }
+  }
```
greglittlefield-wf, line 68: We should also add tests for the fix, but we can do that in a follow-up PR

greglittlefield-wf (approved):
+10

Feedback can be addressed in a follow-up PR

PR #612, opened 2020-07-29
CPLAT-11722 Add uiForwardRef

joebingham-wk:
## Motivation
  
When planning function components, we missed a couple cases:

- when you want to forward a ref to a function component
- when you want to forward a ref using something than the existing component's props (way less common / important to solve)
Example: https://reactjs.org/docs/forwarding-refs.html#forwarding-refs-to-dom-components
## Changes
  
API Changes:
- Add `uiForwardRef`
- Update ref example to include different syntax examples
- Refactor forwardRef tests to be able to share test coverage with `uiForwardRef`
- Add `uiForwardRef` tests

Builder Changes:
- Rename `FunctionComponentConfig` to `UiFactoryConfig`
- Rename / refactor `isFunctionComponentFactory` to `shouldGenerateConfig`
- Create a getter on `Identifier` that can be used to check if the name is a factory type
- Switch out a couple checks on the identifier name with checks for the presence of a generated `UiFactoryConfig`

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] QA that the refs print correctly in the example. Note that the components do have code around logging, but that's to keep some consistency between React's docs and our examples. Because there aren't any props updated, there isn't any re-rendering and no reason for the components to log.

## Merge Checklist


aviary2-wf:
## Security Insights
The items listed below may not capture all security relevant changes. Before providing a security review, be sure to review the entire PR for security impact.
<details>
<summary>(1) Security relevant changes were detected</summary>
<li>Watched keyword <code>innerHtml</code> in <code>web/component2/src/demos/forward_ref.dart</code> line(s) ['182', '203'] added</li>
</details>

## Action Items
* Obtain a [security review](https://wiki.atl.workiva.net/display/SECURITY/Development+Security+Review+Guidelines); reviewer should pay special attention to insights listed above
---

test/vm_tests/builder/codegen_test.dart
```diff
@@ -707,7 +707,7 @@ main() {
               (props) {
                 return Dom.div()();
               }, 
-              $BarPropsMixinConfig, // ignore: undefined_identifier
```
joebingham-wk, line 710: This _seems_ like it should have been `$BarConfig` from the get go - my parser change use the name to detect that it's a generated config and therefore enforces the naming pattern `${FactoryName}Config`. Let me know if this doesn't work, though!

test/vm_tests/builder/declaration_parsing_test.dart
```diff
@@ -1665,7 +1665,7 @@ main() {
                   (props) {
                     return Dom.div()();
                   },
-                  \$FooConfig, // ignore: undefined_identifier
```
joebingham-wk, line 1668: This _seems_ like it should have been `$BarConfig` from the get go - my parser change use the name to detect that it's a generated config and therefore enforces the naming pattern `${FactoryName}Config`. Let me know if this doesn't work, though!

tools/analyzer_plugin/lib/src/diagnostic/boilerplate_validator.dart
```diff
@@ -253,16 +253,6 @@ extension on SourceSpan {
   SourceRange asRangeOrEmpty() => this == null ? SourceRange.EMPTY : asRange();
 }
 
-extension<E> on Iterable<E> {
-  void forEachIndexed(void Function(E, int) callback) {
```
joebingham-wk, line 257: The new analyzer version made improvements to the analysis of extensions and caught this was unused

tools/analyzer_plugin/lib/src/diagnostic/missing_cascade_parens.dart
```diff
@@ -1,6 +1,6 @@
 // ignore: deprecated_member_use
 import 'package:analyzer/analyzer.dart'
-    show CompileTimeErrorCode, NodeLocator, StaticTypeWarningCode, StaticWarningCode;
```
joebingham-wk, line 3: The codes below were moved to `CompileTimeErrorCode`: https://github.com/dart-lang/sdk/commit/8de8a1c226bf227c485f0e024e225fbd248139c5#diff-ff9cb85838db7a802ac5aa9c55f26d86

lib/src/builder/parsing/members/factory.dart
```diff
@@ -56,7 +56,7 @@ class BoilerplateFactory extends BoilerplateMember {
   bool get isFunctionComponentFactory =>
       node.variables.firstInitializer != null &&
       anyDescendantIdentifiers(
-          node.variables.firstInitializer, (identifier) => identifier.name == 'uiFunction');
+          node.variables.firstInitializer, (identifier) => identifier.isFunctionType);
```
greglittlefield-wf, line 59: I don't think we should base whether it's a function component off of a set of known function names, but rather use the presence of a generated config to tell instead.

This getter is also used to determine whether to generate a config, and for the case of `uiJsComponent`, the name `isFunctionComponentFactory` isn't accurate, since it's not a function component.

So, I think this should be something like:
```dart
  bool get shouldGenerateConfig => node.referencesGeneratedConfig;
```

lib/src/component_declaration/component_base.dart
```diff
@@ -91,6 +93,10 @@ ReactDartComponentFactoryProxy registerAbstractComponent(Type abstractComponentC
 /// via a fluent-style builder interface.
 typedef TProps UiFactory<TProps extends UiProps>([Map backingProps]);
 
+extension UiFactoryHelpers on UiFactory {
+  UiFactoryConfig<TProps> asForwardRefConfig<TProps extends bh.UiProps>({String displayName}) => UiFactoryConfig(propsFactory: PropsFactory.fromUiFactory(this), displayName: displayName);
+}
```
greglittlefield-wf, line 98: The generic parameter can actually be specifiedd on the extension itself, making it so you don't have to duplicate the typing!
```suggestion
extension UiFactoryHelpers<TProps extends bh.UiProps> on UiFactory<TProps> {
  UiFactoryConfig<TProps> asForwardRefConfig({String displayName}) => UiFactoryConfig(propsFactory: PropsFactory.fromUiFactory(this), displayName: displayName);
}
```

Usage before:
```dart
Dom.button.asForwardRefConfig<DomProps>(displayName: 'FancyButton'),
```
Usage after
```dart
Dom.button.asForwardRefConfig(displayName: 'FancyButton'),
```

lib/src/component/ref_util.dart
```diff
@@ -176,11 +177,11 @@ Ref<T> createRef<T>() {
 ///
 /// Learn more: <https://reactjs.org/docs/forwarding-refs.html>.
 UiFactory<TProps> Function(UiFactory<TProps>) forwardRef<TProps extends UiProps>(
```
greglittlefield-wf, line 179: This should be deprecated and suggest using `uiForwardRef` instead

lib/src/component/ref_util.dart
```diff
@@ -212,3 +214,282 @@ UiFactory<TProps> Function(UiFactory<TProps>) forwardRef<TProps extends UiProps>
...
+
+  dynamic _uiFunctionWrapper(dynamic props, dynamic ref) {
+    return functionComponent(propsFactory.jsMap(props), ref);
+  }
+
+  ReactJsComponentFactoryProxy factory;
+
+  // If a consumer uses `asForwardRefConfig` to generate the function component
+  // config, displayName could be `null`.
+  if (displayName != null && displayName.trim().isNotEmpty) {
```
greglittlefield-wf, line 467: Why `displayName.trim().isNotEmpty`? 

This seems unnecessary

lib/src/component_declaration/component_base.dart
```diff
@@ -91,6 +93,10 @@ ReactDartComponentFactoryProxy registerAbstractComponent(Type abstractComponentC
 /// via a fluent-style builder interface.
 typedef TProps UiFactory<TProps extends UiProps>([Map backingProps]);
 
+extension UiFactoryHelpers on UiFactory {
+  UiFactoryConfig<TProps> asForwardRefConfig<TProps extends bh.UiProps>({String displayName}) => UiFactoryConfig(propsFactory: PropsFactory.fromUiFactory(this), displayName: displayName);
```
greglittlefield-wf, line 97: This should have a doc comment (maybe just a simple description and just point to `uiForwardRef`)

tools/analyzer_plugin/lib/src/diagnostic/missing_cascade_parens.dart
greglittlefield-wf: These don't exist in the lowest allowed version of the `analyzer` package; that dependency needs to be raised in order to use these.

test/over_react/component/ref_util_test.dart
```diff
@@ -0,0 +1,332 @@
...
+    }
+  }
+
+  group('- commonRefForwardingTests -', () {
+    group('on a component with a dom component child', () {
+      testForwardRefWith(Dom.span, verifyRefValue: (ref) {
+        expect(ref, TypeMatcher<SpanElement>());
+      }, useUiForwardRef: useUiForwardRef);
+
+      test('- using DomProps', () {
```
greglittlefield-wf, line 187: This description doesn't really indicate what's being tested

test/over_react/component/ref_util_test.dart
```diff
@@ -0,0 +1,332 @@
...
+    // component props are accessed differently depending on if it is a dom component
+    // or a dart component
+    String idValue;
+    if (refObject.current is Element) {
+      idValue = refObject.current.id;
+    } else {
+      idValue = refObject.current.props['id'];
+    }
+
+    expect(idValue, equals('test'), reason: 'child component should have access to parent props');
```
greglittlefield-wf, line 300: This reason is a little bit misleading; what about something like "child component receives props passed to it"

test/vm_tests/builder/codegen_test.dart
```diff
@@ -658,8 +658,8 @@ main() {
 
       group('and generates props config for function components', () {
```
greglittlefield-wf, line 659: Can we add some test cases for `uiForwardRef` (and other arbitrary functions, assuming we support them as per my comments above), since only `uiFunction` is currently being tested?

tools/analyzer_plugin/lib/src/diagnostic/missing_cascade_parens.dart
joebingham-wk: Ohh yep, thanks for catching that!

lib/src/builder/parsing/ast_util.dart
```diff
@@ -42,16 +42,30 @@ extension InitializerHelperTopLevel on TopLevelVariableDeclaration {
...
+            } else if (args[1] is PrefixedIdentifier) {
+              return args[1]
+                  .childEntities
+                  .where((child) {
+                    return child is SimpleIdentifier &&
+                        child.name == '\$${uiFactoryDeclaration.name.name}Config';
+                  })
+                  .toList()
+                  .isNotEmpty;
+            }
```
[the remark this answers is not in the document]
joebingham-wk: That's so concise and elegant. Thank you!!

lib/src/builder/parsing/members/factory.dart
joebingham-wk: Yep, good call. During my first pass I went back and forth here but thought going off the name was still fair (though should definitely have updated `isFunctionComponentFactory` - sorry about that). With fresh eyes though, you're definitely right!

lib/src/component_declaration/component_base.dart
joebingham-wk: Niice, much cleaner!

lib/src/component/ref_util.dart
joebingham-wk: 🤦 Oh, you mean like it says in the AC? Sorry about that!

greglittlefield-wf: No worries!

lib/src/component/ref_util.dart
```diff
@@ -212,3 +214,282 @@ UiFactory<TProps> Function(UiFactory<TProps>) forwardRef<TProps extends UiProps>
 
   return wrapWithForwardRef;
 }
+
+/// Automatically passes a [Ref] through a component to one of its children.
```
greglittlefield-wf, line 218: Some notes on this doc comment:
- The first sentence in the description could be updated to state a little more clearly what it does
- I think it'd be helpful to link the React JS docs up front, since it contains more information and can help introduce consumers to this concept
- We should show LHS typing in these, since that's our recommendation when declaring top-level function components.
- I think the examples could be adjusted to more closely resemble real-world use-cases, and that we could list the more common ones up front.
    
    For example, I think a good portion of the time, `uiForwardRef` forwarding to a DOM component is going to use a custom props class, and not just directly use `DomProps`. I think we should lead with recommending custom props classes over wrapping them with `asForwardRefConfig`.

- I think the class component HOC example (which I know was copied from `forwardRef`) is  a little convoluted, and could be improved.
- I think all of the imports and react_dom.render parts of the boilerplate can be omitted, since they add clutter/length to the examples

With that in mind, these are my suggested changes which address these points:
```dart
/// Creates a function component capable of forwarding its ref to
/// a component it renders.
///
/// Learn more: <https://reactjs.org/docs/forwarding-refs.html>.
///
/// ### Example 1: Updating a function component to forward a ref:
///
/// _This example mirrors the JS example in the link above._
///
/// Consider a `FancyButton` function component that renders the native button DOM element:
/// ```dart
/// mixin FancyButtonProps on UiProps {}
///
/// UiFactory<FancyButtonProps> FancyButton = uiFunction(
///   (props) {
///     return (Dom.button()
///       ..className = 'FancyButton'
///     )(props.children);
///   },
///   $FancyButtonConfig, // ignore: undefined_identifier
/// );
/// ```
///
/// Normally, you can't set a ref on function components, but if you change
/// [uiFunction] to `uiForwardRef`, you'll get a new `ref` argument that will be
/// populate with any ref set on `FancyButton`. We This ref can then be
/// "forwarded" to some other component.
///
/// ```dart
/// mixin FancyButtonProps on UiProps {}
///
/// UiFactory<FancyButtonProps> FancyButton = uiForwardRef(
///   (props, ref) {
///     return (Dom.button()
///       ..ref = ref
///       ..className = 'FancyButton'
///     )(props.children);
///   },
///   $FancyButtonConfig, // ignore: undefined_identifier
/// );
///
/// usageExample() {
///   // You can now get a ref directly to the DOM button:
///   final ref = createRef<ButtonElement>();
///   return (FancyButton()..ref = ref)();
/// }
/// ```
///
/// ### Example 2: Creating a [higher-order component](https://reactjs.org/docs/higher-order-components.html)
/// that forwards its ref to the wrapped component.
///
/// ```dart
/// /// Wraps a component ([factoryToWrap]) in a new component that logs when rendered.
/// UiFactory<TProps> withLogging<TProps extends UiProps>(UiFactory<TProps> factoryToWrap) {
///   return uiForwardRef(
///     (props, ref) {
///       useEffect(() => 'the component rendered!');
///
///       return (factoryToWrap()
///         ..addAll(props)
///         ..ref = ref
///       )(props.children);
///     },
///     factoryToBeWrapped.asForwardRefConfig(
///       displayName: 'withLogging(${getDisplayName(factoryToWrap)})',
///     ),
///   );
/// }
///
/// UiFactory<FancyButtonProps> FancyButton = ...;
///
/// /// This can be used just like FancyButton, and setting a ref on it will work the same!
/// UiFactory<FancyButtonProps> FancyButtonWithLogging = withLogging(FancyButton);
/// ```
///
/// ### Example 3: Exposing inner refs in class components
///
/// While it's usually easier to pass `uiForwardRef`'s `ref` argument directly
/// to the desired component, sometimes it's nested within a class component
/// that can't forward refs itself, and it's non-trivial to update that class
/// component to a function component.
///
/// In this case, you can pass the ref to the class component as a custom prop,
/// and have the class component forward it to the right location.
///
/// Here, we'll make our class component factory private and only expose a
/// version wrapped in `uiForwardRef`:
///
/// ```dart
/// UiFactory<LogPropsProps> _Foo = _$_Foo; // ignore: undefined_identifier
/// mixin FooProps on UiProps {
///   // Private since we only use this to pass along the ref provided in
///   // uiForwardRef.
///   //
///   // Consumers will effectively be setting this when they set `ref` on the public factory.
///   dynamic _inputRef;
/// }
/// class FooComponent extends UiComponent2<FooProps> {
///   @override
///   render() {
///     return Dom.div()(
///       (Dom.input()
///         ..type = 'text'
///         ..ref = _inputRef
///       )();
///     )
///   }
/// }
///
/// // ---------- Wrapping a Class Component ----------
/// // Here you have two options:
/// //   - Option 1: Use the class component's factory as the UiFactoryConfig arg.
/// //   This needs to be done because the builder recognizes
/// //   `FooProps` as already consumed (by the class component).
/// //
/// //   - Option 2: Create a new props class. This just works around the issue
/// //   described for Option 1 because it is creating a new props class, but it
/// //   only needs to mix in the props mixins that the original props class used.
/// //
/// // Choosing between the options is likely circumstantial or preferential.
/// // The advantage to Option 1 is that if the class component has numerous mixins,
/// // it is much more concise to create the function component config.
/// // Option 2 has the benefit that it matches the declaration of standard
/// // function components (which `uiForwardRef` returns). Additionally, Option 2
/// // illustrates how one could add additional props to the wrapped function component.
///
/// //
/// // Option 1 Example
/// UiFactory<FooProps> Foo = uiForwardRef((props, ref) {
///     return (_Foo()
///       ..addProps(props)
///       .._inputRef = ref
///     )();
///   },
///   _Foo.asForwardRefConfig(),
/// );
///
/// //
/// // Option 2 Example:
///
/// // This is not necessary but is just in place to illustrate that more props
/// // can be specified and consumed.
/// mixin AnotherPropsMixin on UiProps {
///   String anExampleAdditionalProp;
/// }
///
/// class Foo2Props = UiProps with AnotherPropsMixin, FooProps;
/// final Foo2 = uiForwardRef<LogsPropsComponent2Props>((props, ref) {
///     print(props.anExampleAdditionalProp);
///
///     return (_Foo()
///       ..addProps(props)
///       .._inputRef = ref
///     )();
///   },
///   $Foo2Config, // ignore: undefined_identifier
/// );
```

lib/src/component_declaration/component_base.dart
joebingham-wk: Hmmm when I use that syntax, it doesn't pick up the underlying factory's generics value and has a return type of `UiFactoryConfig<UiProps>` when the generic should be a specific class (which yields this error: react-dom.development.js:3843 Uncaught Error: Type 'UiFactoryConfig<UiProps>' should be 'UiFactoryConfig<LogPropsProps>' to implement expected type 'UiFactoryConfig<LogPropsProps>'.). I poked around but didn't see a way to resolve that - does one come to your mind, @greglittlefield-wf?

greglittlefield-wf: Hmm I'm seeing the return type work for me locally:
Screen Shot 2020-08-05 at 3 10 06 PM

I'm about to step away for a quick break, but I can try running an example with it after!

lib/src/component/ref_util.dart
joebingham-wk: Right?? It's weird. This may have just been covering up the symptom instead of the actual problem, so let me know if I should dig deeper. I added this though because if a consumer doesn't add a `displayName`, somehow it ends up ' '. I tried just `displayName.isEmpty`, and that didn't work. The result in the React devtools is (note the empty line that just says `forwardRef` - in pure ReactJS that would be 'Anonymous'):
Screen Shot 2020-08-05 at 3 08 36 PM

So yeah, it's weird and I spent a little time investigating before just adding this line.

lib/src/component_declaration/component_base.dart
joebingham-wk: Ahh, never mind @greglittlefield-wf. I found the issue and it works just fine! 🙂

greglittlefield-wf: Noice!

lib/src/component/ref_util.dart
joebingham-wk: (I should definitely make the code comment more descriptive, at very least though)

greglittlefield-wf: I would be curious how `' '` gets into displayName in the first place; I feel like that shouldn't happen

joebingham-wk: Cool cool, I can look into that!

lib/src/component/ref_util.dart
joebingham-wk: Nice, I really like it. Thanks for typing that up!

lib/src/component/ref_util.dart
joebingham-wk: Fixed here: d66db3f!

joebingham-wk:
Woops, last commit broke tests. Fixing :loading:

lib/src/component/ref_util.dart
joebingham-wk: I reverted that change - the issue was because anon functions in JS do have a name property, but it's an empty string. It seemed like an innocent change but to just have that `getFunctionName` function return `null` if it's empty, but that caused some breakages. react-dart also has a similar method that's identical, so it feels like we've built up a little bit of dependence on the behavior that null doesn't get returned. That makes me think this is still the simplest solution, but what do you think @greglittlefield-wf?

smaifullerton-wk:
QA in progress

lib/src/builder/parsing/ast_util.dart
```diff
@@ -42,17 +42,16 @@ extension InitializerHelperTopLevel on TopLevelVariableDeclaration {
...
-            final args =
-                identifier.thisOrAncestorOfType<MethodInvocation>()?.argumentList?.arguments;
-            if (args == null || args.length < 2) return false;
-            return args[1].toString().startsWith(RegExp(r'\$'));
-          }
-          return false;
+          final uiFactoryDeclaration = identifier.thisOrAncestorOfType<VariableDeclaration>();
+
+          return firstInitializer != null &&
+              anyDescendantIdentifiers(firstInitializer, (identifier) {
```
greglittlefield-wf, line 52: This is already inside a `firstInitializer != null` and `anyDescendantIdentifiers` check

lib/src/builder/parsing/ast_util.dart
```diff
@@ -42,17 +42,16 @@ extension InitializerHelperTopLevel on TopLevelVariableDeclaration {
...
     return firstInitializer != null &&
         anyDescendantIdentifiers(firstInitializer, (identifier) {
-          if (identifier.name == 'uiFunction') {
-            final args =
-                identifier.thisOrAncestorOfType<MethodInvocation>()?.argumentList?.arguments;
-            if (args == null || args.length < 2) return false;
-            return args[1].toString().startsWith(RegExp(r'\$'));
-          }
-          return false;
+          final uiFactoryDeclaration = identifier.thisOrAncestorOfType<VariableDeclaration>();
```
greglittlefield-wf, line 49: You can just use `firstVariable` here instead

web/component2/src/demos/ref.dart
```diff
@@ -0,0 +1,338 @@
+// ignore_for_file: deprecated_member_use_from_same_package
+import 'dart:html';
+
+import 'package:over_react/over_react.dart';
+
+// ignore: uri_has_not_been_generated
+part 'ref.over_react.g.dart';
+
+// ------------ `uiForwardRef` with a function component (simple) ------------
```
greglittlefield-wf, line 9: Sorry, I forgot to take a look at this file when I last reviewed.

Even the "simple" versions of these components are a bit complicated due to how many layers of components there are.

Would it be possible to update these demos to match the simplified versions in the `uiForwardRef` doc comment?

joebingham-wk: Oh no problem! I was kinda surprised there wasn't a comment here ;) Yep, I can work on simplifying!

greglittlefield-wf: Thanks! ❤️

lib/src/component/ref_util.dart
greglittlefield-wf: That sounds good to me!

smaifullerton-wk:
talked to Joe about some puzzling things I found while QAing and he's going to take a look -

web/component2/src/demos/ref.dart
```diff
@@ -147,149 +146,90 @@ class LogPropsComponent extends UiComponent2<LogPropsProps> {
...
-            ..onClick = (_) {
-              printButtonOuterHtml(buttonNodeRefForComplexComponent);
-              lastClickedRef.set(buttonNodeRefForComplexComponent);
-            })(),
+            ..ref = fancyButtonWithLoggingReg
+            ..onClick = (_) => printButtonOuterHtml(fancyButtonUiForwardRef))(),
         ),
+        (RefDemoHoc()..demoTitle = '`uiForwardRef` wrapping a class (option 1)')((Foo()
+          ..ref = fooInputRef
+          ..onChange = (e) => print('Foo Input Ref: ${e.target.value}'))()),
```
greglittlefield-wf, line 212: Formatting is off in this file; might wanna use use over_react_format

web/component2/src/demos/ref.dart
```diff
@@ -147,149 +146,90 @@ class LogPropsComponent extends UiComponent2<LogPropsProps> {
...
+mixin BazProps on UiProps {
+  BuilderOnlyUiFactory<UiProps> builder;
 
-      prevPropsRef.current = props;
-    });
+  // Private since we only use this to pass along the value of `ref` to
+  // the return value of forwardRef.
+  //
+  // Consumers can set this private field value using the public `ref` setter.
+  Ref _forwardedRef;
```
greglittlefield-wf, line 168: This shouldn't be typed as `Ref` since it prevents callback refs from being passed in, which is a valid use-case.

smaifullerton-wk:
QA +1

tested functionality in `demos/ref.dart`:

- [x] button ref forwarded to function component HTML is printed when button is clicked
- [x] button ref forwarded to higher order component HTML is printed when button is clicked
- [x] input element ref from class wrapped w/ `uiForwardRef` using `asForwardRefConfig` input value is printed 
- [x] input element ref from class wrapped w/ `uiForwardRef` and generated config using custom props input value & custom prop are printed

smaifullerton-wk:
QA +1 still stands - latest commit is a wording change only

PR #621, opened 2020-08-25
CPLAT-12205 Add WithTransition wrapper component to enable "controlled" transitions using props

aaronlademann-wf:
## Motivation
Currently, `AbstractTransitionComponent`s can only be programmatically shown/hidden using the `show` / `hide` / `toggle` API methods on the instance.

With React 16 componentry - this is not ideal since calling the methods from a parent component requires the use of a `ref`, and the child component is no longer guaranteed to have been mounted when the parent component is. This makes it very difficult to hide/show things within data-driven applications without having to work around runtime race conditions.

## Changes
1. Add the `WithTransition` wrapper component. 
    This component is designed to have identical behavior to `AbstractTransitionComponent`, but allows the consumer to control the behavior using the value of `props.isShown`.
2. Add tests.

#### Release Notes
Add `WithTransition` wrapper component to enable "controlled" transitions using props

### QA Checklist

## Merge Checklist

### DIFF
--- a/lib/src/component/with_transition.dart
+++ b/lib/src/component/with_transition.dart
@@ -280,7 +280,7 @@ class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, 
     assert(_hasSingleValidChild(props));
 
     final childElement = props.children.single as ReactElement;
-    final childProps = domProps({...getProps(childElement)});
+    final childProps = domProps(getProps(childElement));
     final phaseProps = props.childPropsByPhase[state.$transitionPhase];
     final phaseClasses = ClassNameBuilder.fromProps(childProps)
       ..addFromProps(phaseProps);
--- a/lib/src/component/with_transition.dart
+++ b/lib/src/component/with_transition.dart
@@ -281,7 +281,7 @@ class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, 
 
     final childElement = props.children.single as ReactElement;
     final childProps = domProps(getProps(childElement));
-    final phaseProps = props.childPropsByPhase[state.$transitionPhase];
+    final phaseProps = props.childPropsByPhase[state.$transitionPhase] ?? const {};
     final phaseClasses = ClassNameBuilder.fromProps(childProps)
       ..addFromProps(phaseProps);
 
--- a/lib/src/component/with_transition.dart
+++ b/lib/src/component/with_transition.dart
@@ -49,7 +49,7 @@ mixin WithTransitionPropsMixin on UiProps {
   ///   (props) {
   ///     final isShown = useState(false);
   ///
-  ///     (WithTransition()
+  ///     return (WithTransition()
   ///       ..isShown = isShown.value
   ///       ..childPropsByPhase = useMemo(() => {
   ///         TransitionPhase.PRE_SHOWING: domProps()..className = 'before-showing',
@@ -60,7 +60,7 @@ mixin WithTransitionPropsMixin on UiProps {
   ///       })
   ///     )(
   ///       // The child that has CSS transitions
-  ///     )
+  ///     );
   ///   },
   ///   $WithTransitionExampleConfig, // ignore: undefined_identifier
   /// );
@@ -97,13 +97,10 @@ class WithTransitionProps = UiProps with v2.TransitionPropsMixin, WithTransition
 /// }
 ///
 /// UiFactory<WithTransitionExampleProps> WithTransitionExample = uiFunction(
-///   (_props) {
-///     final defaultProps = useMemo(() {
-///       return WithTransitionExample()..initiallyShown = false;
-///     }, const []);
-///     final props = defaultProps..addProps(_props);
+///   (props) {
+///     final initiallyShown = props.initiallyShown ?? false;
 ///
-///     final isShown = useState(props.initiallyShown);
+///     final isShown = useState(initiallyShown);
 ///
 ///     final classes = ClassNameBuilder.fromProps(props)
 ///       ..add('fade')
@@ -166,28 +163,23 @@ class WithTransitionProps = UiProps with v2.TransitionPropsMixin, WithTransition
 /// );
 /// ```
 UiFactory<WithTransitionProps> WithTransition = memo(uiFunction(
-  (_props) {
+  (props) {
     // ----- Default Props ----- //
-    final defaultProps = useMemo(() {
-      return WithTransition()
-      ..isShown = false
-      ..transitionTimeout = const Duration(seconds: 1)
-      ..childPropsByPhase = const {
-        v2.TransitionPhase.PRE_SHOWING: {},
-        v2.TransitionPhase.SHOWING: {},
-        v2.TransitionPhase.SHOWN: {},
-        v2.TransitionPhase.HIDING: {},
-        v2.TransitionPhase.HIDDEN: {},
-      }
-      ..addProps(v2.TransitionPropsMixin.defaultProps);
-    }, const []);
-    final props = defaultProps..addProps(_props);
+    final isShown = props.isShown ?? false;
+    final transitionTimeout = props.transitionTimeout ?? const Duration(seconds: 1);
+    final childPropsByPhase = props.childPropsByPhase ?? const {
+      v2.TransitionPhase.PRE_SHOWING: {},
+      v2.TransitionPhase.SHOWING: {},
+      v2.TransitionPhase.SHOWN: {},
+      v2.TransitionPhase.HIDING: {},
+      v2.TransitionPhase.HIDDEN: {},
+    };
 
     assert(_hasSingleValidChild(props));
 
     // ----- State / Hooks ----- //
     final transitionNodeRef = useRef<Element>();
-    final transitionPhaseDerivedFromProps = props.isShown ? TransitionPhase.SHOWN : TransitionPhase.HIDDEN;
+    final transitionPhaseDerivedFromProps = isShown ? TransitionPhase.SHOWN : TransitionPhase.HIDDEN;
     final transitionPhase = useState(transitionPhaseDerivedFromProps);
 
     /// Begin showing the node unless the [currentPhase] is already shown or is in the process of showing.
@@ -205,13 +197,6 @@ UiFactory<WithTransitionProps> WithTransition = memo(uiFunction(
       tProps.onWillHide?.call();
       transitionPhase.set(
           _hasTransitionOut(tProps) ? TransitionPhase.HIDING : TransitionPhase.HIDDEN);
-    }
-
-    // getDerivedStateFromProps
-    if (props.isShown && _isOrWillBeHidden(transitionPhase.value)) {
-      _handleShow(props, transitionPhase.value);
-    } else if (!props.isShown && _isOrWillBeShown(transitionPhase.value)) {
-      _handleHide(props, transitionPhase.value);
     }
 
     // ----- transitionend event handling ----- //
@@ -252,7 +237,7 @@ UiFactory<WithTransitionProps> WithTransition = memo(uiFunction(
       _cancelTransitionEventListener();
       _cancelTransitionEndTimer();
 
-      _transitionEndTimer.current = Timer(props.transitionTimeout, () {
+      _transitionEndTimer.current = Timer(transitionTimeout, () {
         assert(ValidationUtil.warn(
             'The number of transitions expected to complete have not completed. Something is most likely wrong.',
         ));
@@ -287,6 +272,15 @@ UiFactory<WithTransitionProps> WithTransition = memo(uiFunction(
         _cancelTransitionEventListener();
       };
     }, const []);
+
+    useLayoutEffect(() {
+      // getDerivedStateFromProps
+      if (isShown && _isOrWillBeHidden(transitionPhase.value)) {
+        _handleShow(props, transitionPhase.value);
+      } else if (!isShown && _isOrWillBeShown(transitionPhase.value)) {
+        _handleHide(props, transitionPhase.value);
+      }
+    }, [isShown, props.onWillShow, props.onWillHide]);
 
     // componentDidUpdate
     useLayoutEffect(() {
@@ -348,14 +342,22 @@ UiFactory<WithTransitionProps> WithTransition = memo(uiFunction(
           props.onDidShow?.call();
           break;
       }
-    }, [transitionPhase.value]);
+    }, [
+      transitionPhase.value,
+      props.onDidHide,
+      props.onDidShow,
+      transitionTimeout,
+      props.transitionCount,
+      props.transitionInCount,
+      props.transitionOutCount,
+    ]);
 
     // ----- Rendering ----- //
 
     // NOTE: This cast is safe because we validate that it is a ReactElement via `_hasSingleValidChild`.
     final childElement = props.children.single as ReactElement;
     final childProps = domProps({...getProps(childElement)});
-    final phaseProps = props.childPropsByPhase[transitionPhase.value];
+    final phaseProps = childPropsByPhase[transitionPhase.value];
     final phaseClasses = ClassNameBuilder.fromProps(childProps)
       ..addFromProps(phaseProps);
 
--- a/lib/src/component/with_transition.dart
+++ b/lib/src/component/with_transition.dart
@@ -243,12 +243,10 @@ class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, 
         _handlePreShowing();
         break;
       case TransitionPhase.SHOWING:
-        // NOTE: props.onWillShow is called within _handleShow in the handler that
-        // triggers an update to `$transitionPhase` that will result in this effect being processed.
+        props.onWillShow?.call();
         break;
       case TransitionPhase.HIDING:
-        // NOTE: props.onWillHide is called within _handleHide in the handler that
-        // triggers an update to `$transitionPhase` that will result in this effect being processed.
+        props.onWillHide?.call();
         _transitionNotGuaranteed = state.$transitionPhase == TransitionPhase.SHOWING;
         _handleHiding();
         break;
@@ -311,7 +309,6 @@ class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, 
   Map _stateToBeginShowing(WithTransitionProps tProps, TransitionPhase currentPhase) {
     if (_isOrWillBeShown(currentPhase)) return null;
 
-    tProps.onWillShow?.call();
     return newState()
       ..$transitionPhase = _hasTransitionIn(tProps) ? TransitionPhase.PRE_SHOWING : TransitionPhase.SHOWN;
   }
@@ -321,7 +318,6 @@ class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, 
   Map _stateToBeginHiding(WithTransitionProps tProps, TransitionPhase currentPhase) {
     if (_isOrWillBeHidden(currentPhase)) return null;
 
-    tProps.onWillHide?.call();
     return newState()
       ..$transitionPhase = _hasTransitionOut(tProps) ? TransitionPhase.HIDING : TransitionPhase.HIDDEN;
   }
--- a/lib/src/component/with_transition.dart
+++ b/lib/src/component/with_transition.dart
@@ -289,7 +289,7 @@ class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, 
 
     return cloneElement(childElement, domProps()
       ..addTestId(childProps.getTestId())
-      ..addTestId('ovr.WithTransition.node')
+      ..addTestId('or.WithTransition.node')
       ..addProps(phaseProps)
       ..className = phaseClasses.toClassName()
       ..ref = chainRefs(childElement.ref, _transitionNodeRef)
--- a/lib/src/component/with_transition.dart
+++ b/lib/src/component/with_transition.dart
@@ -174,18 +174,6 @@ mixin WithTransitionState on UiState {
 }
 
 class WithTransitionComponent extends UiStatefulComponent2<WithTransitionProps, WithTransitionState> {
-  /// Whether the overlay is not guaranteed to transition in response to the current
-  /// state change.
-  ///
-  /// _Stored as variable as workaround for not adding breaking change to [_stateToBeginHiding] API._
-  ///
-  /// A transition may not always occur when the state moves from SHOWING to HIDING
-  /// if the PRE_SHOWING-->SHOWING-->HIDING transition happens back-to-back.
-  ///
-  /// Better to not always transition when the user is ninja-toggling a transitionable
-  /// component than to break state changes waiting for a transition that will never happen.
-  var _transitionNotGuaranteed = false;
-
   /// Timer used to determine if a transition timeout has occurred.
   Timer _transitionEndTimer;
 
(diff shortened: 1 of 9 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/component/abstract_transition_props.dart
```diff
@@ -25,6 +25,13 @@ mixin TransitionPropsMixin on UiProps {
   static final TransitionPropsMixin defaultProps = TransitionPropsMapView({})
     ..transitionCount = 1;
 
+  /// Prop that allows you to control the visibility of the component with props.
+  ///
+  /// __NOTE:__ In order for this work properly, if you have an implementation
+  /// of [AbstractTransitionComponent.initiallyShown] in your subclass - it must
+  /// use/reference this value.
+  bool shown;
```
aaronlademann-wf, line 33: Named this way instead of `isShown` to avoid gross prop namespacing / forwarding issues in consumer libs that have an `isShown` prop in a mixin that is used in props classes that also mix in `TransitionPropsMixin`.

test/over_react/component/abstract_transition_test.dart
```diff
@@ -31,613 +32,743 @@ part 'abstract_transition_test.over_react.g.dart';
...
-    group('renders with correct initial state when the `initiallyShown` getter returns', () {
-      test('true', () {
-        var renderedInstance = render(Transitioner()
-          ..initiallyShown = true
-        );
-        TransitionerComponent transitioner = getDartComponent(renderedInstance);
-
-        expect(transitioner.state.transitionPhase, equals(TransitionPhase.SHOWN));
+    @isTestGroup
+    void sharedTransitionerTests(BuilderOnlyUiFactory<AbstractSharedTransitionerProps> factory, {
```
aaronlademann-wf, line 36: Strongly recommend reviewing this file using a [whitespace-agnostic diff](https://github.com/Workiva/over_react/pull/621/files?w=1)

joebingham-wk (commented):
This all looks awesome! I just have one question, but it's a small one, and then I'm +1!

lib/src/component/abstract_transition.dart
```diff
@@ -100,37 +100,95 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps,
...
+    return (newState()
+      ..transitionPhase = this.initiallyShown ? TransitionPhase.SHOWN : TransitionPhase.HIDDEN
+    );
+  }
+
+  @mustCallSuper
+  @override
+  Map getDerivedStateFromProps(Map nextProps, Map prevState) {
+    final tNextProps = typedPropsFactory(nextProps);
+    if (tNextProps.shown == null) return prevState;
```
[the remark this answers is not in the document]
aaronlademann-wf: @joebingham-wk on initial mount, `getDerivedStateFromProps` is called - and the `prevState` value is the return value of the `initialState` getter.  So if you return `null`, the `initialState` value(s) will not be applied for the first render.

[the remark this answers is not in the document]
joebingham-wk: @aaronlademann-wf Oh, I didn't know that was expected. The React docs (which I've criticized for not being verbose enough) don't mention that though, and when I run [this reduced test case](https://gist.github.com/joebingham-wk/5ca1fa2492c9927d3fdaa5f7b60a908c), I'm seeing the initial state value come through... I'm sure either way this code works, I just want to make sure I understand the lifecycle method correctly. Does that test case miss something though?

lib/src/component/abstract_transition.dart
```diff
@@ -100,37 +100,95 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps,
   );
 
   @override
-  get initialState => (newState()
-    ..transitionPhase = this.initiallyShown ? TransitionPhase.SHOWN : TransitionPhase.HIDDEN
-  );
+  get initialState {
```
greglittlefield-wf, line 103: #nit this could stay as an arrow function

lib/src/component/abstract_transition.dart
```diff
@@ -100,37 +100,95 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps,
...
+
+  @mustCallSuper
+  @override
+  Map getDerivedStateFromProps(Map nextProps, Map prevState) {
+    final tNextProps = typedPropsFactory(nextProps);
+    if (tNextProps.shown == null) return prevState;
+
+    final tPrevState = typedStateFactory(prevState);
+    if (tNextProps.shown && _isOrWillBeHidden(tPrevState)) {
+      return _handleShow(tNextProps, tPrevState, this);
```
greglittlefield-wf, line 117: `this` cannot be used in `getDerivedStateFromProps`

lib/src/component/abstract_transition.dart
```diff
@@ -100,37 +100,95 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps,
...
+
   /// Whether transitions are enabled for this component.
-  bool get hasTransition => true;
+  ///
+  /// Will not work when the component is controlled using [TransitionPropsMixin.shown].
+  /// Use [TransitionPropsMixin.transitionCount] instead.
+  ///
+  /// __Deprecated.__ Use [TransitionPropsMixin.transitionCount] instead.
+  @Deprecated('Use `props.transitionCount` instead.')
+  bool get hasTransition => _hasTransition(props);
```
greglittlefield-wf, line 146: Same question as below; what's the migration path? https://sourcegraph.wk-dev.wdesk.org/search?q=get%5CshasTransition+repo:%5Egithub%5C.com/Workiva/web_skin_dart%24+&patternType=regexp#1

lib/src/component/abstract_transition.dart
```diff
@@ -311,9 +380,21 @@ abstract class AbstractTransitionComponent<T extends AbstractTransitionProps,
...
+  @Deprecated('Use props.onWillShow instead.')
   void prepareShow() {}
 
   /// Method that will be called right before the [AbstractTransitionComponent] begins to hide.
+  ///
+  /// Will not work when the component is controlled using [TransitionPropsMixin.shown].
+  /// Use [TransitionPropsMixin.onWillHide] instead.
+  ///
+  /// __Deprecated.__ Use [TransitionPropsMixin.onWillHide] instead.
+  @Deprecated('Use props.onWillHide instead.')
```
greglittlefield-wf, line 397: Do we have a migration path for WSD components consuming these APIs? https://sourcegraph.wk-dev.wdesk.org/search?q=prepare%28Hide%7CShow%29+repo:%5Egithub%5C.com/Workiva/web_skin_dart%24+&patternType=regexp#3

greglittlefield-wf (changes requested):
Given the `getDerivedStateFromProps` issues and deprecation concerns, is it possible to add controlled behavior to this component cleanly?

We might be better off starting fresh with a new `TransitionPrimitive` component that's entirely props-based

test/over_react/component/with_transition_test.dart
```diff
@@ -0,0 +1,522 @@
...
+import 'package:over_react/over_react.dart' hide TransitionPropsMixin, $TransitionPropsMixin, AbstractTransitionProps, AbstractTransitionState, AbstractTransitionComponent;
+import 'package:over_react/components.dart';
+import 'package:over_react_test/over_react_test.dart';
+import 'package:test/test.dart';
+
+import '../../test_util/test_util.dart';
+
+part 'with_transition_test.over_react.g.dart';
+
+/// Main entry point for AbstractTransition testing
```
joebingham-wk, line 30: #nit
```suggestion
/// Main entry point for WithTransition testing
```

joebingham-wk (dismissed):
+1! This all looks great to me, and shows a lot of cool things you can do with functional components, hooks, along with how to test for them!

maxwellpeterson-wf (dismissed):
This is working well in two different places in my shell refactor branch. The documentation is also excellent and easy to follow!

maxwellpeterson-wf:
+10
- Able to use `childPropsByPhase` to set various props on the child component
- Event callbacks (such as `onDidHide`) fire as expected
- Able to control the component using `isShown`

joebingham-wk (dismissed):
+1

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+  /// key(s) in the `Map` are the same.
+  ///
+  /// __Example:__
+  ///
+  /// ```dart
+  /// UiFactory<WithTransitionExampleProps> WithTransitionExample = uiFunction(
+  ///   (props) {
+  ///     final isShown = useState(false);
+  ///
+  ///     (WithTransition()
```
greglittlefield-wf, line 52: #nit this expression is unused;
```suggestion
  ///     return (WithTransition()
```

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+  ///       ..childPropsByPhase = useMemo(() => {
+  ///         TransitionPhase.PRE_SHOWING: domProps()..className = 'before-showing',
+  ///         TransitionPhase.SHOWING: domProps()..className = 'showing',
+  ///         TransitionPhase.SHOWN: domProps()..className = 'shown',
+  ///         TransitionPhase.HIDING: domProps()..className = 'hiding',
+  ///         TransitionPhase.HIDDEN: domProps()..className = 'hidden',
+  ///       })
+  ///     )(
+  ///       // The child that has CSS transitions
+  ///     )
```
greglittlefield-wf, line 63: #nit
```suggestion
  ///     );
```

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+/// }
+///
+/// UiFactory<WithTransitionExampleProps> WithTransitionExample = uiFunction(
+///   (_props) {
+///     final defaultProps = useMemo(() {
+///       return WithTransitionExample()..initiallyShown = false;
+///     }, const []);
+///     final props = defaultProps..addProps(_props);
+///
+///     final isShown = useState(props.initiallyShown);
```
greglittlefield-wf, line 106: #nit the preferred way to default props should be via null-awares (for a number of reasons, including promoting the values to non-nullable once nnbd is enabled)

```suggestion
///     final initiallyShown = props.initiallyShown ?? false;
///
///     final isShown = useState(initiallyShown);
```

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+///       ..add('fade')
+///       ..add('in', isShown.value);
+///
+///     return Dom.div()(
+///       Dom.div()(
+///         (Dom.button()
+///           ..type = 'button'
+///           ..onClick = (useCallback((_) {
+///             isShown.setWithUpdater((isShown) => !isShown);
+///           }, const []))
```
greglittlefield-wf, line 118: #nit unnecessary parentheses
```suggestion
///           ..onClick = useCallback((_) {
///             isShown.setWithUpdater((isShown) => !isShown);
///           }, const [])
```

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+      ..childPropsByPhase = const {
+        v2.TransitionPhase.PRE_SHOWING: {},
+        v2.TransitionPhase.SHOWING: {},
+        v2.TransitionPhase.SHOWN: {},
+        v2.TransitionPhase.HIDING: {},
+        v2.TransitionPhase.HIDDEN: {},
+      }
+      ..addProps(v2.TransitionPropsMixin.defaultProps);
+    }, const []);
+    final props = defaultProps..addProps(_props);
```
greglittlefield-wf, line 184: #nit prefer using null-awares to default props like mentioned above

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+      transitionPhase.set(
+          _hasTransitionOut(tProps) ? TransitionPhase.HIDING : TransitionPhase.HIDDEN);
+    }
+
+    // getDerivedStateFromProps
+    if (props.isShown && _isOrWillBeHidden(transitionPhase.value)) {
+      _handleShow(props, transitionPhase.value);
+    } else if (!props.isShown && _isOrWillBeShown(transitionPhase.value)) {
+      _handleHide(props, transitionPhase.value);
+    }
```
greglittlefield-wf, line 215: These calls mutating state and performing side-effects in `render`, which isn't allowed; they should go in `useEffect`

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+        _isMounted.current = false;
+
+        _cancelTransitionEndTimer();
+        _cancelTransitionEventListener();
+      };
+    }, const []);
+
+    // componentDidUpdate
+    useLayoutEffect(() {
+      if (!_isMounted.current) return;
```
greglittlefield-wf, line 293: Will this ever get called after the component is unmounted?

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+          // triggers an update to `transitionPhase` that will result in this effect being processed.
+          break;
+        case TransitionPhase.HIDING:
+          // NOTE: props.onWillHide is called within _handleHide in the handler that
+          // triggers an update to `transitionPhase` that will result in this effect being processed.
+          _transitionNotGuaranteed.current = transitionPhase.value == TransitionPhase.SHOWING;
+          _handleHiding();
+          break;
+        case TransitionPhase.HIDDEN:
+          props.onDidHide?.call();
```
greglittlefield-wf, line 345: I believe these need to also need to be included in the dependencies array: https://reactjs.org/docs/hooks-reference.html#conditionally-firing-an-effect

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,410 @@
...
+    }
+
+    void _cancelTransitionEndTimer() {
+      _transitionEndTimer.current?.cancel();
+      _transitionEndTimer.current = null;
+    }
+
+    /// Listens for the next `transitionend` event and invokes a callback after
+    /// the event is dispatched.
+    void onNextTransitionEnd(Function() complete) {
```
greglittlefield-wf, line 249: This references a number of props inside it (`props.transitionTimeout`, `_transitionInCount(props)`) that aren't declared in the calling `useLayoutEffect`

greglittlefield-wf (changes requested):
The hooks in this component are pretty complex; I'm see a couple issues, and then there are some areas in which I can't tell for sure whether rules of effect hooks are being broken.

It seems like this component's state logic doesn't translate super well to hooks, and that we'd need to rewrite it from the ground up with hooks in mind if we wanted to be confident in the implementation.

Perhaps for this time around, we should just use a class component under the hood, and expose a forwardRef component wrapper around it?

aaronlademann-wf:
@greglittlefield-wf can you be more specific with what you're seeing as far as breaking the rules of hooks?

lib/src/component/with_transition.dart
aaronlademann-wf: The over_react analyzer plugin complains about this if you remove them.

greglittlefield-wf:
>All tests we used for the old component are passing, and we have a consumer +10.

The current component has more edge cases and potentially incorrect code paths with the way it uses hooks, so I don't think that's enough to say the component in it's current state is bug-free.

Bugs introduced by improper uses of hooks might not always be surfaced in simple usages / tests, but rather by sequences of rerenders.

In this case, what about if you rerender the component when

lib/src/component/with_transition.dart
aaronlademann-wf: No, but it does get called once before the component mounts.

aaronlademann-wf:
@greglittlefield-wf I have addressed your feedback.

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
aaronlademann-wf: Addressed in https://github.com/Workiva/over_react/pull/621/commits/967d93dd4c6cbb6fac367b02d128a6002613d8b8

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+    _cancelTransitionEventListener();
+  }
+
+  @override
+  render() {
+    assert(_hasSingleValidChild(props));
+
+    final childElement = props.children.single as ReactElement;
+    final childProps = domProps({...getProps(childElement)});
+    final phaseProps = props.childPropsByPhase[state.$transitionPhase];
```
greglittlefield-wf, line 286: While all of the methods this is being used in allow `null`, I think it'd be good to add a null-aware here to prevent null exceptions from occurring in the future
```suggestion
    final phaseProps = props.childPropsByPhase[state.$transitionPhase] ?? const {};
```

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+    _cancelTransitionEndTimer();
+    _cancelTransitionEventListener();
+  }
+
+  @override
+  render() {
+    assert(_hasSingleValidChild(props));
+
+    final childElement = props.children.single as ReactElement;
+    final childProps = domProps({...getProps(childElement)});
```
greglittlefield-wf, line 285: #nit unnecessary map copy
```suggestion
    final childProps = domProps(getProps(childElement));
```

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+  get initialState => (newState()
+    ..$transitionPhase = _transitionPhaseDerivedFromProps(props)
+  );
+
+  @override
+  getDerivedStateFromProps(Map nextProps, Map prevState) {
+    final tNextProps = typedPropsFactory(nextProps);
+    final tPrevState = typedStateFactory(prevState);
+    if (tNextProps.isShown && _isOrWillBeHidden(tPrevState.$transitionPhase)) {
+      return _stateToBeginShowing(tNextProps, tPrevState.$transitionPhase);
```
greglittlefield-wf, line 221: getDerivedSTateFromProps shouldn't have side effects (e.g., calling a props callback in this case) since in concurrent mode this method can be called multiple times without actually rendering the result. Can we move those to `componentDidUpdate`?

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+      // changes waiting on `transitionend` events.
+      _cancelTransitionEventListener();
+    }
+
+    switch (state.$transitionPhase) {
+      case TransitionPhase.PRE_SHOWING:
+        _handlePreShowing();
+        break;
+      case TransitionPhase.SHOWING:
+        // NOTE: props.onWillShow is called within _handleShow in the handler that
```
greglittlefield-wf, line 246: > _handleShow
This comment is out of date

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+        // triggers an update to `$transitionPhase` that will result in this effect being processed.
+        break;
+      case TransitionPhase.HIDING:
+        // NOTE: props.onWillHide is called within _handleHide in the handler that
```
greglittlefield-wf, line 250: > _handleHide
This comment is out of date

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+
+    final childElement = props.children.single as ReactElement;
+    final childProps = domProps({...getProps(childElement)});
+    final phaseProps = props.childPropsByPhase[state.$transitionPhase];
+    final phaseClasses = ClassNameBuilder.fromProps(childProps)
+      ..addFromProps(phaseProps);
+
+    return cloneElement(childElement, domProps()
+      ..addTestId(childProps.getTestId())
+      ..addTestId('ovr.WithTransition.node')
```
greglittlefield-wf, line 292: #supernit I like this abbreviation better: over_react -> o_r => or
```suggestion
      ..addTestId('or.WithTransition.node')
```

greglittlefield-wf (changes requested):

lib/src/component/with_transition.dart
```diff
@@ -0,0 +1,426 @@
...
+        _handlePreShowing();
+        break;
+      case TransitionPhase.SHOWING:
+        // NOTE: props.onWillShow is called within _handleShow in the handler that
+        // triggers an update to `$transitionPhase` that will result in this effect being processed.
+        break;
+      case TransitionPhase.HIDING:
+        // NOTE: props.onWillHide is called within _handleHide in the handler that
+        // triggers an update to `$transitionPhase` that will result in this effect being processed.
+        _transitionNotGuaranteed = state.$transitionPhase == TransitionPhase.SHOWING;
```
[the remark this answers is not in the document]
aaronlademann-wf: If I change it to `tPrevState` - two tests fail.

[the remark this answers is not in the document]
greglittlefield-wf: So perhaps this flag isn't needed anymore? If the code isn't changed to prevState, then the flag should be eliminated, because right now it's always `false`.

Do the AbstractTransition tests fail when that code I linked is changed to `_transitionNotGuaranteed = false;`?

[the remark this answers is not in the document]
aaronlademann-wf: Yes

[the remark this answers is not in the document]
greglittlefield-wf: Cool, sounds like this impl just doesn't need it, likely because of the different way state transitions work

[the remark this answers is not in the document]
aaronlademann-wf: Removed in https://github.com/Workiva/over_react/pull/621/commits/e0208a6375576540b0fd0cccc5ffd2e5cc74f611

greglittlefield-wf (approved):

joebingham-wk (approved):

maxwellpeterson-wf:
+10
- pulled into https://github.com/Workiva/wdesk_sdk/pull/5096 where there are two separate usages of `WithTransition`. Both work as expected.

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #633, opened 2020-10-09
CPLAT-12194 Enable functional components to use consumed props

joebingham-wk:
## Motivation
  
In class based components, we have the API's `addUnconsumedProps` to facilitate passing along props more easily. We want the same behavior in functional components.

## Changes
  
- Create a new class (`PropsInstanceMeta`) that uses a mixin on `PropsMetaCollection` to manage consumed props operations (e0cde45)
- Add a property onto `UiProps` called `$meta` typed as `PropsInstanceMeta` (e0cde45)
- Update the builder to override the field on Component2 / functional component props instances (8855f7e)
- Add a mixin (`UiPropsMeta`) on `UiProps` that can be used to access `$meta`, along with adding APIs to derive consumed props (e0cde45)
- Add examples (51b5466)

### STILL LEFT TO DO:
- Receive feedback on the approach thus far
- Add doc comments / documentation
- Add tests

#### Release Notes
  
- Add APIs for using consumed props in functional components

## Review

  

Please review: 
@aaronlademann-wf @greglittlefield-wf @sydneyjodon-wk 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -32,13 +32,14 @@ class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
 class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
   @override
   render() {
-    final meta = UiPropsMeta(props).meta;
-    meta.consume(ParentOnlyPropsMixin);
+    final meta = UiPropsMeta(props).meta.forMixin(ParentOnlyPropsMixin);
 
     return (
         Dom.div()(
-          'The parent prop is: ${props.aParentProp}',
-          (SomeClassChild()..modifyProps((propsToChange) => meta.deriveUnconsumedProps(props, propsToChange)))(),
+          Dom.div()(
+            'The parent prop is: ${props.aParentProp}',
+          ),
+          (SomeClassChild()..addUnconsumedProps(props, meta))(),
         )
     );
   }
@@ -52,8 +53,8 @@ class SomeClassChildComponent extends UiComponent2<SomeChildProps> {
   @override
   render() {
     return (
-        Fragment()(
-          'The passed prop value is ${props.aPropToBePassed}',
+        Dom.div()(
+          'The passed prop value is: ${props.aPropToBePassed}',
         )
     );
   }
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -32,7 +32,7 @@ class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
 class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
   @override
   render() {
-    final meta = UiPropsMeta(props).meta.forMixin(ParentOnlyPropsMixin);
+    final meta = UiPropsMeta(props).staticMeta.forMixin(ParentOnlyPropsMixin);
 
     return (
         Dom.div()(
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -32,7 +32,7 @@ class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
 class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
   @override
   render() {
-    final meta = UiPropsMeta(props).staticMeta.forMixins({ParentOnlyPropsMixin});
+    final meta = props.staticMeta.forMixins({ParentOnlyPropsMixin});
 
     return (
         Dom.div()(
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -1,0 +1,60 @@
+// Copyright 2020 Workiva Inc.
+//
+// Licensed under the Apache License, Version 2.0 (the "License");
+// you may not use this file except in compliance with the License.
+// You may obtain a copy of the License at
+//
+//     http://www.apache.org/licenses/LICENSE-2.0
+//
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+import 'package:over_react/over_react.dart';
+
+// ignore_for_file: uri_has_not_been_generated
+part 'new_class_consumed_props.over_react.g.dart';
+
+UiFactory<SomeParentProps> SomeClassParent = _$SomeClassParent; // ignore: undefined_identifier
+
+mixin ParentOnlyPropsMixin on UiProps {
+  String aParentProp;
+}
+
+mixin SharedPropsMixin on UiProps {
+  String aPropToBePassed;
+}
+
+class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
+
+class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
+  @override
+  render() {
+    final meta = UiPropsMeta(props).meta;
+    meta.consume(ParentOnlyPropsMixin);
+
+    return (
+        Dom.div()(
+          'The parent prop is: ${props.aParentProp}',
+          (SomeClassChild()..modifyProps((propsToChange) => meta.deriveUnconsumedProps(props, propsToChange)))(),
+        )
+    );
+  }
+}
+
+UiFactory<SomeChildProps> SomeClassChild = _$SomeClassChild; // ignore: undefined_identifier
+
+class SomeChildProps = UiProps with SharedPropsMixin;
+
+class SomeClassChildComponent extends UiComponent2<SomeChildProps> {
+  @override
+  render() {
+    return (
+        Fragment()(
+          'The passed prop value is ${props.aPropToBePassed}',
+        )
+    );
+  }
+}
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -32,14 +32,14 @@ class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
 class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
   @override
   render() {
-    final meta = UiPropsMeta(props).staticMeta.forMixin(ParentOnlyPropsMixin);
+    final meta = UiPropsMeta(props).staticMeta.forMixins({ParentOnlyPropsMixin});
 
     return (
         Dom.div()(
           Dom.div()(
             'The parent prop is: ${props.aParentProp}',
           ),
-          (SomeClassChild()..addUnconsumedProps(props, meta.inList()))(),
+          (SomeClassChild()..addUnconsumedProps(props, meta))(),
         )
     );
   }
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -39,7 +39,7 @@ class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
           Dom.div()(
             'The parent prop is: ${props.aParentProp}',
           ),
-          (SomeClassChild()..addUnconsumedProps(props, meta))(),
+          (SomeClassChild()..addUnconsumedProps(props, meta.inList()))(),
         )
     );
   }
--- a/lib/src/builder/codegen/typed_map_impl_generator.dart
+++ b/lib/src/builder/codegen/typed_map_impl_generator.dart
@@ -226,7 +226,7 @@ abstract class TypedMapImplGenerator extends BoilerplateDeclarationGenerator {
 
       if (allPropsMixins != null) {
         generatePropsMeta(buffer, allPropsMixins,
-            classType: 'PropsMetaCollection', fieldName: r'$meta');
+            classType: 'PropsMetaCollection', fieldName: r'staticMeta');
       }
     }
 
--- a/lib/src/builder/codegen/typed_map_impl_generator.dart
+++ b/lib/src/builder/codegen/typed_map_impl_generator.dart
@@ -226,7 +226,7 @@ abstract class TypedMapImplGenerator extends BoilerplateDeclarationGenerator {
 
       if (allPropsMixins != null) {
         generatePropsMeta(buffer, allPropsMixins,
-            classType: 'PropsInstanceMeta', fieldName: r'$meta', isConst: false);
+            classType: 'PropsMetaCollection', fieldName: r'$meta');
       }
     }
 
@@ -379,9 +379,7 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
         factoryNames =
             declaration.factories.map((factory) => FactoryNames(factory.name.name)).toList(),
         member = declaration.props.either,
-        allPropsMixins = declaration.props.either.nodeHelper.mixins
-            ?.map<Identifier>((mixin) => mixin.name)
-            ?.toList(),
+        allPropsMixins = declaration.allPropsMixins,
         isProps = true,
         componentFactoryName = 'null',
         isFunctionComponentDeclaration = declaration.factories.first.shouldGenerateConfig,
--- a/lib/src/builder/codegen/typed_map_impl_generator.dart
+++ b/lib/src/builder/codegen/typed_map_impl_generator.dart
@@ -134,6 +134,7 @@ abstract class TypedMapImplGenerator extends BoilerplateDeclarationGenerator {
   String _generateConcretePropsOrStateImpl({
     String componentFactoryName,
     String propKeyNamespace,
+    List<Identifier> allPropsMixins,
   }) {
     if (isProps) {
       if (componentFactoryName == null || propKeyNamespace == null) {
@@ -222,6 +223,10 @@ abstract class TypedMapImplGenerator extends BoilerplateDeclarationGenerator {
             '  /// The default namespace for the prop getters/setters generated for this class.')
         ..writeln('  @override')
         ..writeln('  String get propKeyNamespace => ${stringLiteral(propKeyNamespace)};');
+
+      if (allPropsMixins != null) {
+        generatePropsMeta(buffer, allPropsMixins, classType: 'PropsInstanceMeta', fieldName: r'$meta', isConst: false);
+      }
     }
 
     // End of class body
@@ -345,10 +350,13 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
   @override
   final Version version;
 
+  final List<Identifier> allPropsMixins;
+
   _TypedMapImplGenerator.props(ClassComponentDeclaration declaration)
       : names = TypedMapNames(declaration.props.either.name.name),
         factoryNames = [FactoryNames(declaration.factory.name.name)],
         member = declaration.props.either,
+        allPropsMixins = declaration.allPropsMixins,
         isProps = true,
         componentFactoryName = ComponentNames(declaration.component.name.name).componentFactoryName,
         isFunctionComponentDeclaration = false,
@@ -358,6 +366,7 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
       : names = TypedMapNames(declaration.state.either.name.name),
         factoryNames = [FactoryNames(declaration.factory.name.name)],
         member = declaration.state.either,
+        allPropsMixins = null,
         isProps = false,
         componentFactoryName = ComponentNames(declaration.component.name.name).componentFactoryName,
         isFunctionComponentDeclaration = false,
@@ -369,6 +378,7 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
         factoryNames =
             declaration.factories.map((factory) => FactoryNames(factory.name.name)).toList(),
         member = declaration.props.either,
+        allPropsMixins = declaration.props.either.nodeHelper.mixins?.map<Identifier>((mixin) => mixin.name)?.toList(),
         isProps = true,
         componentFactoryName = 'null',
         isFunctionComponentDeclaration = declaration.factories.first.shouldGenerateConfig,
@@ -403,6 +413,7 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
       componentFactoryName: componentFactoryName,
       // This doesn't really apply to the new boilerplate
       propKeyNamespace: '',
+      allPropsMixins: allPropsMixins,
     ));
   }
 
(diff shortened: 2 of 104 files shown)


aviary-wf:
## Security Insights

## Action Items
---

example/builder/src/new_class_consumed_props.dart
```diff
@@ -0,0 +1,59 @@
...
+mixin SharedPropsMixin on UiProps {
+  String aPropToBePassed;
+}
+
+class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
+
+class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
+  @override
+  render() {
+    UiPropsMeta(props).meta.consume(ParentOnlyPropsMixin);
```
joebingham-wk, line 35: I didn't want to bring this up for discussion until after the APIs are a little more solid, but here I was wondering: could this replace the original `consumedProps` mechanism? It would keep the API consistent and reduce some of what the builder needs to do, since this is available on all `Component2` instances anyway. 

The main tradeoff is that this system isn't as intuitive for class based components and is really only manageable if they pass along props in `render` since they need to be re-consume the props every time a new props instance is received (which also has a performance implication, even if slight)

example/builder/src/new_class_consumed_props.dart
```diff
@@ -0,0 +1,60 @@
...
+  String aPropToBePassed;
+}
+
+class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
+
+class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
+  @override
+  render() {
+    final meta = UiPropsMeta(props).meta;
+    meta.consume(ParentOnlyPropsMixin);
```
joebingham-wk, line 36: Could this replace the original consumedProps mechanism? It would keep the API consistent and reduce some of what the builder needs to do, since this is available on all Component2 instances anyway.

The main tradeoff is that this system isn't as intuitive for class based components and could get unmanageable fast since they need to be re-consume the props every time a new props instance is received (which also has a performance implication, even if slight)

greglittlefield-wf: (I'm commenting on the overall approach in this thread, not just specific to class components.)

I'm not crazy about the idea of storing the consumed props themselves on the props instance because:
- This is storing additional state on what's supposed to be purely a typed view into a map class.
- The state needs to be recreated each time it's used (e.g. in each render). This isn't huge deal, but it means that we wouldn't be to perform future optimizations like caching the full set of keys under the hood.
- This blurs the line of separation of concerns between storing the metadata for all props mixins used, and storing the set of props the consumer has opted into.

I would lean toward having the API be closer to what we have for `consumedProps`, which separates these concerns, and also provides consistency with the current syntax:
```dart
final consumedProps = props.meta.forMixins({ParentOnlyPropsMixin});
```


With that, you would also be able to memoize it if you really wanted to:
```dart
final _getConsumedProps = memoize2<Iterable<ConsumedProps>, PropsMetaCollection>((meta) => meta.forMixins({
  ParentOnlyPropsMixin,
}));

// or with a helper function that fills in type parameters:

final _getConsumedProps = memoizeConsumedProps((meta) => meta.forMixins({
  ParentOnlyPropsMixin,
}));

UiFactory<FooProps> Foo = uiFunction((props) {
  final consumedProps = _getConsumedProps(props.meta);
});
```

---

As far as using the consumed props goes, currently we have this:
```dart
..modifyProps((propsToChange) => meta.deriveUnconsumedProps(props, propsToChange))
```
I think we could add some helpers to help reduce this boilerplate.

Since we're passing through `propsToChange`, instead of creating that closure ourselves, we could instead return a function:
```dart
..modifyProps(meta.addUnconsumedPropsFrom(props))
```
Or even add something to UiProps:
```dart
..addUnconsumedProps(props, meta)
// Or if we end up using something like that `consumedProps` API I mentioned above
..addUnconsumedProps(props, consumedProps)
```

lib/src/builder/codegen/typed_map_impl_generator.dart
```diff
@@ -369,6 +379,9 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
         factoryNames =
             declaration.factories.map((factory) => FactoryNames(factory.name.name)).toList(),
         member = declaration.props.either,
+        allPropsMixins = declaration.props.either.nodeHelper.mixins
+            ?.map<Identifier>((mixin) => mixin.name)
+            ?.toList(),
```
greglittlefield-wf, line 384: I think this implementation might not be correct when using the shorthand syntax, since `.nodeHelper.mixins` on a mixin will be empty.

For class components, we're using `allPropsMixins`, which does account for that case. Could we share that same implementation between `ClassComponentDeclaration` and `PropsMapViewOrFunctionComponentDeclaration` so that we can use it here as well?

lib/src/builder/codegen/util.dart
```diff
@@ -66,3 +67,22 @@ String generatedMixinWarningCommentLine(TypedMapNames mixinNames, {@required boo
...
+  StringBuffer buffer,
+  List<Identifier> mixins, {
+  String classType = 'PropsMetaCollection',
+  String fieldName = 'propsMeta',
+  bool isConst = true,
+}) {
+  buffer
+    ..writeln()
+    ..writeln('  @override')
+    ..writeln('  $classType get $fieldName => ${isConst ? 'const' : ''} $classType({');
```
greglittlefield-wf, line 81: Why is the instance conditionally const? Can't it always be const?

joebingham-wk: I might be missing something, but not with the current implementation of `InstancePropsMeta` because it has instance fields. If I change that to not keep that state though, then that will clean this up as well!

example/builder/src/new_class_consumed_props.dart
joebingham-wk: Your APIs looked awesome - thanks for writing that up! Added those API changes in 9c7ebbe.

lib/src/builder/codegen/util.dart
joebingham-wk: Updated this in ea0e4aa!

lib/src/builder/codegen/typed_map_impl_generator.dart
joebingham-wk: Definitely, good catch. Updated in ea0e4aa!

example/builder/src/new_class_consumed_props.dart
```diff
@@ -32,13 +32,14 @@ class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
 class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
   @override
   render() {
-    final meta = UiPropsMeta(props).meta;
-    meta.consume(ParentOnlyPropsMixin);
+    final meta = UiPropsMeta(props).meta.forMixin(ParentOnlyPropsMixin);
```
greglittlefield-wf, line 35: #nit usage of the UiPropsMeta wrapper isn't necessary in cases where there isn't already a `meta field`
```suggestion
    final meta = props.meta.forMixin(ParentOnlyPropsMixin);
```
I think we should omit this where possible

lib/src/component_declaration/builder_helpers.dart
```diff
@@ -126,102 +124,93 @@ abstract class UiProps extends component_base.UiProps with GeneratedClass {
...
 
-  void consumeAllExceptFor(Set<Type> mixins) {
-    _consumedProps.addAll(allExceptForMixins(mixins));
-  }
+  /// Copies props from the provided [props] instance to this [UiProps] instance,
+  /// filtering out props found in [consumedProps].
+  ///
+  /// [consumedProps] should be either a [PropsMeta] or `Iterable<PropsMeta>` instance.
+  /// These are both return types of [PropsMetaCollection]'s related APIs
+  /// (`forMixin`, `forMixins`, `allExceptForMixins`).
```
greglittlefield-wf, line 150: I think we should make it just `Iterable<PropsMeta>` to provide better typing and make it easier for consumers to tell if they're passing in the wrong thing.

The `forMixin` use-case should be able be changed to `forMixins` instead in most cases, or in edge cases wrap it in a list literal like we do in the consumedProps getter in components.

lib/src/util/map_util.dart
```diff
@@ -75,6 +76,9 @@ Map getPropsToForward(Map props, {
 ///
 /// Based upon configuration, the function will overlook [props] that are not
 /// meant to be passed on, such as non-DOM props or specified values.
+///
+/// DEPRECATED: Use [forwardUnconsumedPropsV2] instead.
+@Deprecated('This implementation does not filter DOM props correctly. Use forwardUnconsumedPropsV2 instead.')
```
greglittlefield-wf, line 81: 👏

greglittlefield-wf (commented):
Noticed a couple things but this generally looks great!

lib/src/util/react_util.dart
```diff
@@ -40,6 +40,8 @@ class UiPropsMapView extends MapView
   bool get $isClassGenerated =>
       throw UnimplementedError('@PropsMixin instances do not implement \$isClassGenerated');
 
+  PropsMetaCollection get $meta => throw UnimplementedError('@PropsMixin instances do not implement instance meta');
```
greglittlefield-wf, line 43: Unrelated to this PR, but we should really deprecate the `UiPropsMapView` class... It's made obsolete by the new boilerplate for props map views, which are real `UiProps` classes

lib/src/component_declaration/builder_helpers.dart
joebingham-wk: Yep, good call, that is a lot cleaner!

joebingham-wk: Addressed as part of ec43447!

example/builder/src/new_class_consumed_props.dart
joebingham-wk: Unfortunately this always seems necessary :/ Without it, I get `error: Static field 'meta' can't be accessed through an instance. (instance_access_to_static_member at [over_react] example/builder/src/new_class_consumed_props.dart:35)`

joebingham-wk: I could rename it to `propsMeta`, `instanceMeta`, or something to that effect though. I can't think of anything really succinct, but those could still be better than requiring always wrapping

greglittlefield-wf: Oh gross, because it's conflicting with the generated static `meta` fields on the legacy props mixins mixed into UiProps itself...

I knew that would be an issue for legacy props classes annd mixins, but I didn't realize that those were being mixed into `UiProps`...

For example, removing these fixes it: 
<details>
<summary> <em>Click to expand</em> </summary>

```diff
diff --git a/example/builder/src/new_class_consumed_props.dart b/example/builder/src/new_class_consumed_props.dart
index c2f9a972..55f9b4c8 100644
--- a/example/builder/src/new_class_consumed_props.dart
+++ b/example/builder/src/new_class_consumed_props.dart
@@ -32,7 +32,7 @@ class SomeParentProps = UiProps with ParentOnlyPropsMixin, SharedPropsMixin;
 class SomeClassParentComponent extends UiComponent2<SomeParentProps> {
   @override
   render() {
-    final meta = UiPropsMeta(props).meta.forMixin(ParentOnlyPropsMixin);
+    final meta = props.meta.forMixin(ParentOnlyPropsMixin);
 
     return (
         Dom.div()(
@@ -58,4 +58,4 @@ class SomeClassChildComponent extends UiComponent2<SomeChildProps> {
         )
     );
   }
-}
\ No newline at end of file
+}
diff --git a/lib/src/component/prop_mixins.over_react.g.dart b/lib/src/component/prop_mixins.over_react.g.dart
index 2141f846..d980d03d 100644
--- a/lib/src/component/prop_mixins.over_react.g.dart
+++ b/lib/src/component/prop_mixins.over_react.g.dart
@@ -11,7 +11,6 @@ abstract class ReactPropsMixin implements _$ReactPropsMixin {
   @override
   Map get props;
 
-  static const PropsMeta meta = _$metaForReactPropsMixin;
   @override
   String get key => props['key'];
   @override
@@ -6765,7 +6764,6 @@ abstract class UbiquitousDomPropsMixin implements _$UbiquitousDomPropsMixin {
   @override
   Map get props;
 
-  static const PropsMeta meta = _$metaForUbiquitousDomPropsMixin;
   @override
   @Accessor(doNotGenerate: true)
   AriaPropsMapView _aria;
diff --git a/lib/src/util/class_names.over_react.g.dart b/lib/src/util/class_names.over_react.g.dart
index 391c9f00..d7a7a80f 100644
--- a/lib/src/util/class_names.over_react.g.dart
+++ b/lib/src/util/class_names.over_react.g.dart
@@ -11,8 +11,6 @@ abstract class CssClassPropsMixin implements _$CssClassPropsMixin {
   @override
   Map get props;
 
-  static const PropsMeta meta = _$metaForCssClassPropsMixin;
-
   /// String of space-delimited CSS classes to be added to the resultant DOM.
   ///
   /// All over_react components merge any added classes with this prop and the [classNameBlacklist] prop (see
```

</details>

Ugh we really need to update those to be non-legacy mixins at some point... I'm gonna add that to the list for the next breaking change.

Anyways, yeah, I'm in favor of renaming it so that you don't have to wrap it. It's probably a good thing that this happened since the name `meta` was pretty ambiguous to begin with...

I think it'd be nice if the name made it clear that it's specific to the type/class and not the instance itself, so my vote would be for something like `propsMeta`/`staticMeta`/`classMeta`. At some point, I think it'd be good to get some input from the rest of the squad/team, and maybe even from consumers, to make sure the name is a good fit.

joebingham-wk: Sweeet, yeah, I agree it could be more descriptive and better to just rename it. I made a poll in slack for our squad (and anyone else in the channel). Based off of that, I'll rename it!

lib/src/component_declaration/builder_helpers.dart
greglittlefield-wf: Consumers should prefer using `forMixins` instead of `forMixin`, (since that wraps it inside an iterable for them, and also makes it easier to add additional mixins later without accidentally duplicating them), so I don't think we should optimize for that use-case.

There's also the `.all` that can be used to get all of the mixins in an iterable.

So instead of:
```dart
final consumedProps = props.staticMeta.forMixin(ParentOnlyPropsMixin);
(SomeChild()..addUnconsumedProps(props, consumedProps.inList()))();
```
I think we should recommend:
```dart
final consumedProps = props.staticMeta.forMixins({ParentOnlyPropsMixin});
(SomeChild()..addUnconsumedProps(props, consumedProps))();
```

and for consumers who want to forward all props:
```dart
final consumedProps = props.staticMeta.all;
(SomeChild()..addUnconsumedProps(props, consumedProps))();
```

...and then we can eliminate `inList`.

Another benefit to that is that, `consumedProps` variable is typed as `Iterable<PropsMeta>`, just like the `consumedProps` getter is in class components, making it easier to translate between the two.

lib/src/component_declaration/builder_helpers.dart
```diff
@@ -120,9 +124,61 @@ abstract class UiProps extends component_base.UiProps with GeneratedClass {
...
+  PropsMetaCollection get staticMeta => $meta;
+
+
+  /// Copies props from the provided [props] instance to this [UiProps] instance,
+  /// filtering out props found in [consumedProps].
+  ///
+  /// [consumedProps] should be a `Iterable<PropsMeta>` instance.
+  /// This is the return type of [PropsMetaCollection]'s related APIs `forMixins`
+  /// and `allExceptForMixins`. `forMixin` can be used easily by then calling `.inList()`
+  /// on the returned [PropsMeta] instance.
```
greglittlefield-wf, line 151: Like I commented above, this shouldn't mention `forMixin` and should maybe mention `.all`.

Also, I think it'd be helpful to mention this is useful for forwarding props in function components, and to provide a small example showing how it will be used in the real world.

I think the example can be fairly simple, and doesn't need all the boilerplate and stuff; something like:
```dart
// Consider props in FooProps "consumed"...
final consumedProps = props.staticMeta.forMixins({FooProps});
// ...and filter them out when forwarding props to Bar.
return (Bar()..addUnconsumedProps(props, consumedProps))();
```

lib/src/component_declaration/builder_helpers.dart
```diff
@@ -120,9 +124,61 @@ abstract class UiProps extends component_base.UiProps with GeneratedClass {
...
+  ///
+  /// Synonymous with [UiComponent2]'s `propsMeta`.
+  ///
+  /// This can be used to derive consumed props by usage in conjunction with [addUnconsumedProps]
+  /// and [addUnconsumedDomProps].
+  PropsMetaCollection get staticMeta => $meta;
+
+
+  /// Copies props from the provided [props] instance to this [UiProps] instance,
+  /// filtering out props found in [consumedProps].
```
greglittlefield-wf, line 146: #nit redundant type reference https://dart.dev/guides/language/effective-dart/documentation#avoid-redundancy-with-the-surrounding-context

and #nit it's a little abiguous where the filtering occurs in this context

```suggestion
  /// Copies key-value pairs from the provided [props] map into this map, 
  /// excluding those with keys found in [consumedProps].
```

greglittlefield-wf (changes requested):
Found a couple things, but this is looking great; should be my final pass!

lib/src/component_declaration/builder_helpers.dart
joebingham-wk: Sweeet, that sounds good. I removed `inList` and updated the current examples in 501b2c8!

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/function_component_test.dart
```diff
@@ -232,6 +232,75 @@ void functionComponentTestHelper(UiFactory<TestProps> factory,
...
+      test('', () {
+        secondProps.addUnconsumedDomProps(initialProps, null);
+        expect(secondProps.stringProp, isNull);
+        expect(secondProps.anotherProp, isNull);
+        expect(secondProps.className, className);
+      });
+
+      test('and consumed props are correctly filtered', () {
+        expect(initialProps.className, isNotNull, reason: 'Test setup sanity check');
+        secondProps.addUnconsumedDomProps(initialProps, PropsMeta(fields: [PropDescriptor('className')], keys: ['className']).inList());
```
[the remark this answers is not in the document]
joebingham-wk: Oooh that's dope!

lib/src/component_declaration/builder_helpers.dart
```diff
@@ -120,9 +124,61 @@ abstract class UiProps extends component_base.UiProps with GeneratedClass {
       ' Will be removed in 4.0.0.')
   @toBeGenerated String get propKeyNamespace => throw UngeneratedError(member: #propKeyNamespace);
 
+  @toBeGenerated PropsMetaCollection get $meta => throw UngeneratedError(member: #meta);
+
   @override @toBeGenerated Map get props => throw UngeneratedError(member: #props);
 }
 
+extension UiPropsMeta on UiProps {
```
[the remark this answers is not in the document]
joebingham-wk: That's a really good question. I went back and forth a little myself but I agree that the separation seemed like more effort and confusing to maintain. I updated this in d28a61e!

lib/src/component_declaration/builder_helpers.dart
```diff
@@ -139,42 +133,55 @@ extension UiPropsMeta on UiProps {
...
+  /// // Consider props in FooProps "consumed"...
+  /// final consumedProps = props.staticMeta.forMixins({FooProps});
+  /// // ...and filter them out when forwarding props to Bar.
+  /// return (Bar()..addUnconsumedProps(props, consumedProps))();
+  /// ```
   ///
   /// To only add DOM props, use [addUnconsumedDomProps].
   ///
   /// Related: [UiComponent2]'s `addUnconsumedProps`
   void addUnconsumedProps(Map props, Iterable<PropsMeta> consumedProps) {
```
greglittlefield-wf, line 158: Is there a a reason these are only implemented in the builder_helpers subclass of UiProps and not the component_base version? Since they don't depend directly on generated code, I think they should be moved to the base class.

If we did that, we also wouldn't have to add new `throw UnimplementedError()` overrides to classes that don't extend from this one.

lib/src/util/react_util.dart
```diff
@@ -74,6 +80,8 @@ class UiPropsMapView extends MapView
   @override
   String get testId => getTestId();
 
+
+
```
greglittlefield-wf, line 84: #nit extra newlines

greglittlefield-wf (changes requested):
Just one more small comment

lib/src/util/react_util.dart
joebingham-wk: Oh woops, haha thank you!

lib/src/component_declaration/builder_helpers.dart
joebingham-wk: Nope, good call!

lib/src/component/strictmode_component.dart
```diff
@@ -30,6 +32,17 @@ class StrictModeProps extends component_base.UiProps
...
+  PropsMetaCollection get staticMeta => throw UnimplementedError('StrictModeProps instances do not implement instance meta');
+
+  @override
+  void addUnconsumedProps(Map props, Iterable<PropsMeta> consumedProps) =>
+      throw UnimplementedError('StrictModeProps instances do not implement addUnconsumedProps');
+
+  @override
+  void addUnconsumedDomProps(Map props, Iterable<PropsMeta> consumedProps) =>
+      throw UnimplementedError('StrictModeProps instances do not implement addUnconsumedDomProps');
+
```
greglittlefield-wf, line 45:
```suggestion
```

Since this class inherits a concrete implementation that functions properly, there's no need for us to throw in overrides

lib/src/util/react_util.dart
```diff
@@ -57,6 +63,14 @@ class UiPropsMapView extends MapView
...
       throw UnimplementedError('@PropsMixin instances do not implement modifyProps');
 
+  @override
+  void addUnconsumedProps(Map props, Iterable<PropsMeta> consumedProps) =>
+      throw UnimplementedError('@PropsMixin instances do not implement addUnconsumedProps');
+
+  @override
+  void addUnconsumedDomProps(Map props, Iterable<PropsMeta> consumedProps) =>
+      throw UnimplementedError('@PropsMixin instances do not implement addUnconsumedDomProps');
+
```
greglittlefield-wf, line 73:
```suggestion
```

Same comment as above

greglittlefield-wf (changes requested):

lib/src/util/react_util.dart
joebingham-wk: The `StrictMode` component was as expected, but I'm seeing it as necessary here. And we have the other members (`addProps`) overridden here. Did I miss something though?

greglittlefield-wf: Ah, right; sorry about that!

greglittlefield-wf (approved):

lib/src/util/react_util.dart
joebingham-wk: No worries!!

greglittlefield-wf:
QA +1

aaronlademann-wf (approved):
+1

This looks awesome @joebingham-wk !!!

joebingham-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #608, opened 2020-07-16
Use propsOrStateMapsEqual in connect so that tearoffs don't cause unnecessary rerenders

greglittlefield-wf:
> Discovered while testing #499

## Motivation
[Tearoffs aren't guaranteed to be identical](https://github.com/dart-lang/sdk/issues/31665#issuecomment-352678783), and we don't want the use of them to trigger unnecessary rerenders in connect-ed components

## Changes
Use `propsOrStateMapsEqual` instead of default JS equality

## Review
_[See CONTRIBUTING.md][contributing-review-types] for more details on review types (+1 / QA +1 / +10) and code review process._

  

Please review: 

### QA Checklist
- [x] Tests were updated and provide good coverage of the changeset and other affected code
- [x] Manual testing was performed if needed
    - [x] Steps from PR author: 
       - [x] Verify in redux example app that there are no unnecessary rerenders introduced by these changes
       - [x] Passing consumer tests
    - [x] Anything falling under manual testing criteria [outlined in CONTRIBUTING.md][contributing-manual-testing]

## Merge Checklist
While we perform many automated checks before auto-merging, some manual checks are needed:
- [x] A Client Platform member has reviewed these changes
- [x] There are no unaddressed comments _- this check can be automated if reviewers use the "Request Changes" feature_
- [ ] _For release PRs -_ Version metadata in Rosie comment is correct

[contributing-review-types]: https://github.com/Workiva/over_react/blob/master/CONTRIBUTING.md#review-types
[contributing-manual-testing]: https://github.com/Workiva/over_react/blob/master/CONTRIBUTING.md#manual-testing-criteria

PR #643, opened 2020-11-12
Use propsOrStateMapsEqual in memo so that tearoffs don't cause unnecessary rerenders

aaronlademann-wf:
## Motivation
[Tearoffs aren't guaranteed to be identical](https://github.com/dart-lang/sdk/issues/31665#issuecomment-352678783), and we don't want the use of them to trigger unnecessary rerenders in `memo`ized function components.

[Closely related to how we handled this same problem for `connect`ed components](https://github.com/Workiva/over_react/pull/608)

## Changes
1. Use `propsOrStateMapsEqual` for `areEqual` instead of default JS equality when the consumer does not specify their own custom handler for the `areEqual` parameter.
1. Note that we're using our own utility to do the default shallow comparison in the doc comment for `memo`.
1. Add a regression test.

### DIFF
--- a/lib/src/util/memo.dart
+++ b/lib/src/util/memo.dart
@@ -32,11 +32,11 @@ import 'package:over_react/component_base.dart';
 /// ```dart
 /// import 'package:over_react/over_react.dart';
 ///
-/// UiFactory<UiProps> MemoExample = memo<UiProps>(uiFunction(
+/// UiFactory<UiProps> MemoExample = memo(uiFunction(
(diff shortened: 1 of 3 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

greglittlefield-wf (approved):
+10

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #650, opened 2020-12-09
CPLAT-12620 Make factory config syntax use private variables

sydneyjodon-wk:
## Motivation
  

In Dart 2.9, undefined_identifier errors can no longer be ignored - which causes analyzer errors in the over_react boilerplate when generated code has not yet been generated. To work around this, names beginning with `_$` are now assumed to be generated. The component factory boilerplate already uses the private `_$` prefix, but we need to update generated factory configs to also be private.

## Changes
  
- f676b56 Updated code generation and parsing logic to support both private and public factory configs (deprecating the public config in favor of the private config)
- 1024112 Updated public factory config usages to private
- b49e726 Updated tests to test both private and public factory configs

#### Release Notes
  
Update generated factory configs to be private.
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] CI Passes
      - [ ] Verify examples still work as expected
      - [ ] Verify that tests adequately cover both public and private factory config syntax

## Merge Checklist

### DIFF
--- a/lib/src/builder/codegen/names.dart
+++ b/lib/src/builder/codegen/names.dart
@@ -53,8 +53,7 @@ class FactoryNames {
   /// The name of the public generated function component props config for the factory.
   ///
   /// Example: `$FooConfig
-  @Deprecated('The public config is incompatible with Dart >2.9.0 and will be '
-      'removed. Use `privateConfigName` instead.')
+  @Deprecated('The public config is deprecated and will be removed. Use `privateConfigName` instead.')
   String get publicConfigName => '$_prefix$publicGeneratedPrefix${unprefixedConsumerName}Config';
 }
 
--- a/lib/src/builder/codegen/names.dart
+++ b/lib/src/builder/codegen/names.dart
@@ -48,14 +48,14 @@ class FactoryNames {
   /// The name of the private generated function component props config for the factory.
   ///
   /// Example: `_$FooConfig
-  String get privateConfigName => '_\$${consumerName}Config';
+  String get privateConfigName => '$_prefix$privateSourcePrefix${unprefixedConsumerName}Config';
 
   /// The name of the public generated function component props config for the factory.
   ///
   /// Example: `$FooConfig
   @Deprecated('The public config is incompatible with Dart >2.9.0 and will be '
       'removed. Use `privateConfigName` instead.')
-  String get publicConfigName => '\$${consumerName}Config';
+  String get publicConfigName => '$_prefix$publicGeneratedPrefix${unprefixedConsumerName}Config';
 }
 
 /// A set of names of the different generated members for a given component class.
--- a/lib/src/builder/codegen/names.dart
+++ b/lib/src/builder/codegen/names.dart
@@ -47,12 +47,14 @@ class FactoryNames {
 
   /// The name of the private generated function component props config for the factory.
   ///
-  /// Example: `$FooConfig
-  String get privateConfigName => '_$publicConfigName';
+  /// Example: `_$FooConfig
+  String get privateConfigName => '_\$${consumerName}Config';
 
   /// The name of the public generated function component props config for the factory.
   ///
   /// Example: `$FooConfig
+  @Deprecated('The public config is incompatible with Dart >2.9.0 and will be '
+      'removed. Use `privateConfigName` instead.')
   String get publicConfigName => '\$${consumerName}Config';
 }
 
--- a/lib/src/builder/codegen/names.dart
+++ b/lib/src/builder/codegen/names.dart
@@ -45,10 +45,15 @@ class FactoryNames {
   /// - Output: `_$Foo`
   String get implName => '$_prefix$privateSourcePrefix$unprefixedConsumerName';
 
-  /// The name of the generated function component props config for the factory.
+  /// The name of the private generated function component props config for the factory.
   ///
   /// Example: `$FooConfig
-  String get configName => '\$${consumerName}Config';
+  String get privateConfigName => '_$publicConfigName';
+
+  /// The name of the public generated function component props config for the factory.
+  ///
+  /// Example: `$FooConfig
+  String get publicConfigName => '\$${consumerName}Config';
 }
 
 /// A set of names of the different generated members for a given component class.
--- a/lib/src/builder/codegen/names.dart
+++ b/lib/src/builder/codegen/names.dart
@@ -53,7 +53,8 @@ class FactoryNames {
   /// The name of the public generated function component props config for the factory.
   ///
   /// Example: `$FooConfig
-  @Deprecated('The public config is deprecated and will be removed. Use `privateConfigName` instead.')
+  @Deprecated(
+      'The public config is deprecated and will be removed. Use `privateConfigName` instead.')
   String get publicConfigName => '$_prefix$publicGeneratedPrefix${unprefixedConsumerName}Config';
 }
 
--- a/lib/src/builder/codegen/typed_map_impl_generator.dart
+++ b/lib/src/builder/codegen/typed_map_impl_generator.dart
@@ -390,11 +390,14 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
 
   String _generateUiFactoryConfig(FactoryNames factoryName) {
     return 'final UiFactoryConfig<${names.implName}> '
-        '${factoryName.configName} = UiFactoryConfig(\n'
+        '${factoryName.privateConfigName} = UiFactoryConfig(\n'
         'propsFactory: PropsFactory(\n'
         'map: (map) => ${names.implName}(map),\n'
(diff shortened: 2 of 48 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/builder/codegen/typed_map_impl_generator.dart
```diff
@@ -390,11 +390,14 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
...
   String _generateUiFactoryConfig(FactoryNames factoryName) {
     return 'final UiFactoryConfig<${names.implName}> '
-        '${factoryName.configName} = UiFactoryConfig(\n'
+        '${factoryName.privateConfigName} = UiFactoryConfig(\n'
         'propsFactory: PropsFactory(\n'
         'map: (map) => ${names.implName}(map),\n'
         'jsMap: (map) => ${names.jsMapImplName}(map),),\n'
-        'displayName: \'${factoryName.consumerName}\');\n\n';
+        'displayName: \'${factoryName.consumerName}\');\n\n'
+        '@Deprecated(\'Use `_\\${factoryName.publicConfigName}` instead.\')\n'
```
joebingham-wk, line 398: I feel like it'd be useful to add more context here. Just something like (feel free to change my suggestion to anything that makes more sense to you though 🤷‍♂️ ):
```suggestion
        '@Deprecated(\'This member is incompatible with Dart >2.9.0 and should not be used. Use `_\\${factoryName.publicConfigName}` instead.\')\n'
```

joebingham-wk (commented):
This looks awesome to me! I had a couple small thoughts around deprecations, but all the significant changes look great

lib/src/builder/codegen/names.dart
```diff
@@ -45,10 +45,15 @@ class FactoryNames {
...
+  /// The name of the private generated function component props config for the factory.
   ///
   /// Example: `$FooConfig
-  String get configName => '\$${consumerName}Config';
+  String get privateConfigName => '_$publicConfigName';
+
+  /// The name of the public generated function component props config for the factory.
+  ///
+  /// Example: `$FooConfig
+  String get publicConfigName => '\$${consumerName}Config';
```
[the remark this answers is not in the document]
sydneyjodon-wk: That makes sense! It definitely will need to be removed when we stop generating the public config.

lib/src/builder/codegen/typed_map_impl_generator.dart
sydneyjodon-wk: Updated in 8ce88ad !

joebingham-wk (dismissed):

lib/src/builder/codegen/typed_map_impl_generator.dart
```diff
@@ -390,11 +390,15 @@ class _TypedMapImplGenerator extends TypedMapImplGenerator {
...
   String _generateUiFactoryConfig(FactoryNames factoryName) {
     return 'final UiFactoryConfig<${names.implName}> '
-        '${factoryName.configName} = UiFactoryConfig(\n'
+        '${factoryName.privateConfigName} = UiFactoryConfig(\n'
         'propsFactory: PropsFactory(\n'
         'map: (map) => ${names.implName}(map),\n'
         'jsMap: (map) => ${names.jsMapImplName}(map),),\n'
-        'displayName: \'${factoryName.consumerName}\');\n\n';
+        'displayName: \'${factoryName.consumerName}\');\n\n'
+        '@Deprecated(r\'This member is incompatible with Dart >2.9.0. Use `${factoryName.privateConfigName}` instead.\')\n'
```
greglittlefield-wf, line 398: 1. Dart >**=**2.9.0
2. "Incompatible" is too harsh; it will still work, but the dev experience is worse

There's a lot of context that's hard to include in this string, so perhaps we should just link to it?

```suggestion
        '@Deprecated(r\'Use the private variable, `${factoryName.privateConfigName}`, instead. For information on why this is deprecated, see https://github.com/Workiva/over_react/pull/650\')\n'
```

lib/src/builder/parsing/members/factory.dart
```diff
@@ -87,7 +87,7 @@ class BoilerplateFactory extends BoilerplateMember {
 
     final names = FactoryNames(factoryName);
     final generatedFactoryName = names.implName;
-    final generatedConfigName = names.configName;
+    final generatedConfigName = names.privateConfigName;
 
     final initializer = variable.initializer;
     final referencesGeneratedFactory = initializer != null &&
```
greglittlefield-wf, line 93: For backward compatibility, this should be looking for either the public or the private version, and not just the private one.

This logic is duplicated by `.usesAGeneratedConfig`, so we should just use that instead.

I would expect this to cause validation to fail on the public config... is it not? It looks like there's test coverage around it, but for some reason it's not failing... 😕

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/function_component_test.dart
```diff
@@ -381,7 +385,22 @@ UiFactory<TestProps> Test = uiFunction(
...
+      ..addTestId('testId')
+      ..addProp('data-prop-string-prop', props.stringProp)
+      ..addProp('data-prop-dynamic-prop', props.dynamicProp)
+      ..addProp('data-prop-untyped-prop', props.untypedProp)
+      ..addProp('data-prop-custom-key-prop', props.customKeyProp)
+      ..addProp('data-prop-custom-namespace-prop', props.customNamespaceProp)
+      ..addProp('data-prop-custom-key-and-namespace-prop',
+          props.customKeyAndNamespaceProp))('rendered content');
+  },
+  $TestConfig, // ignore: undefined_identifier, deprecated_member_use_from_same_package
```
greglittlefield-wf, line 403: This should use its own config 
```suggestion
  $TestPublicConfig, // ignore: undefined_identifier, deprecated_member_use_from_same_package
```

test/vm_tests/builder/codegen/names_test.dart
```diff
@@ -37,6 +40,9 @@ main() {
 
         test('consumerName', () => expect(names.consumerName, r'foo.Foo'));
         test('implName', () => expect(names.implName, r'foo._$Foo'));
+        test('privateConfigName', () => expect(names.privateConfigName, r'_$foo.FooConfig'));
+        // ignore: deprecated_member_use_from_same_package
+        test('publicConfigName', () => expect(names.publicConfigName, r'$foo.FooConfig'));
```
greglittlefield-wf, line 45: These should actually be:
```suggestion
        test('privateConfigName', () => expect(names.privateConfigName, r'foo._$FooConfig'));
        // ignore: deprecated_member_use_from_same_package
        test('publicConfigName', () => expect(names.publicConfigName, r'foo.$FooConfig'));
```
My bad for messing up configName and not testing it 😅 

You can look at `implName` to see how they should be implemented

greglittlefield-wf (changes requested):
Just a couple small things; this looks great!

lib/src/builder/codegen/typed_map_impl_generator.dart
greglittlefield-wf: Oh, and maybe we should add a reminder to set the lower bound to the over_react version we expect this to be released in, so that people don't start using these new members and then get pulled into a project with a lower over_react, causing build errors

sydneyjodon-wk: Good idea! I'll update that

lib/src/builder/parsing/members/factory.dart
sydneyjodon-wk: @greglittlefield-wf `generatedConfigName` is only used in the [error message](https://github.com/Workiva/over_react/blob/8ce88addddd95308d85403900d3dae3e8e4cb8e2/lib/src/builder/parsing/members/factory.dart#L98-L107) which is why the build isn't failing. I was thinking it was okay to just use the private config name here since the error message is showing examples of what should be done, but I can add the public config name to the error message if you think it's needed!

greglittlefield-wf: Oh wow, I read this wrong, and thought this code:
```dart
    final referencesGeneratedFactory = initializer != null &&
        anyDescendantIdentifiers(
            initializer, (identifier) => identifier.name == generatedFactoryName);
```
was referencing `generatedConfigName`, but it's referencing `generatedFactoryName` instead. My bad!

No, the private config is good here. Sorry for the confusion!

sydneyjodon-wk: Oh I see! No worries!

greglittlefield-wf (dismissed):

greglittlefield-wf (approved):
+1 (I'll come back and test once I get a chance)

greglittlefield-wf (approved):
+10

PR #657, opened 2021-01-13
Set up Github actions CI

aaronlademann-wf:
Using Github actions for CI is cool, and Travis is dead.


aviary-wf:
## Security Insights

## Action Items
---

.travis.yml
```diff
@@ -1,31 +0,0 @@
...
-        - ./tool/travis_tests.sh
-    - dart: dev
-      name: "OverReact Lib (SDK: dev)"
-      script:
-        - ./tool/travis_quality.sh
-        - ./tool/travis_tests.sh
-    # Analyzer Plugin CI Jobs
-    - dart: "2.7.2"
-      name: "Analyzer Plugin (SDK: 2.7)"
-      script: dart tool/travis_link_plugin_deps.dart && cd ./tools/analyzer_plugin && ./tool/travis.sh
```
sydneyjodon-wk, line 19: Was there a still valid reason to also run analyzer plugin CI on Dart 2.7.2?

sydneyjodon-wk (approved):
+1 Very fancy!

.travis.yml
aaronlademann-wf: Running Github CI on a specific version of Dart is not something I was successful doing... If @greglittlefield-wf feels we should still run the analyzer plugin on 2.7.2, I'll have to look into another option (possibly using Workiva build or something).

.github/workflows/dart_ci.yml
```diff
@@ -0,0 +1,90 @@
...
+
+      - name: Validate dependencies
+        run: dart pub run dependency_validator --no-fatal-pins -i analyzer,build_runner,build_web_compilers,built_value_generator
+        if: always() && steps.install.outcome == 'success'
+
+      - name: Verify formatting
+        run: dart pub run dart_dev format --check
+        if: always() && ${{ matrix.sdk }} == 'stable' && steps.install.outcome == 'success'
+
+      # Analyze before generated files are created to verify that component boilerplate analysis is "clean" without the need for building
```
greglittlefield-wf, line 37: 👏 

This is great, but since we check in generated files in this repo, this doesn't have the desired effect. 

In a separate PR, though (which I've been wanting to do for a while), we can stop checking in files outside of lib, so that components within example and test directories do actually get analyzed without generated code.

.travis.yml
greglittlefield-wf: I think we should still run in 2.7.2 since that's what most of our consumers use for local dev

greglittlefield-wf (approved):
+10

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #669, opened 2021-01-14
CPLAT-13171: Disable implicit casts

greglittlefield-wf:
## Motivation
- Implicit casts will need to be disabled eventually to enable NNBD
- We want to be able to analyze test cases and examples with implicit casts disabled so that we can ensure boilerplate doesn't unknowingly rely on implicit casts

## Changes
_For reviewing, I recommend looking at the commits separately so that generated file changes don't clog up the diff_

- Fix implicit cast issues in source code
   - A lot of issues were solved by adding missing types to arguments or return values, since their types were `dynamic` and thus resulted in implicit casts
   - I tried to use Dart's type promotion where possible in order to avoid explicit casts, so there are cases where I stored a field as a local variable to make that possible.
        ```dart
        // Before
        if (foo.bar is Baz) return foo.bar;

        // is-check type promotion not possible for fields, so you need a cast
        if (foo.bar is Baz) return foo.bar as Baz;

        // With type promotion
        final bar = foo.bar;
        if (bar is Baz) return bar;
        ```
   - The rest of the cases just needed an explicit (`as`) cast
       - Most of these cases were due to `dynamic` values coming from either JS APIs, Maps, or older, untyped react-dart APIs
- Fix implicit cast issues in generated code by updating the builder

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        - [ ] Smoke-test examples and web directory
        
## Merge Checklist

### DIFF
--- a/analysis_options.yaml
+++ b/analysis_options.yaml
@@ -1,6 +1,8 @@
 include: package:pedantic/analysis_options.1.8.0.yaml
 
 analyzer:
+  strong-mode:
+    implicit-casts: false
   exclude:
     - _site/**
     - test_fixtures/**
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -259,14 +259,19 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           docComment = '';
         }
 
+        String castGetterMapValueIfNecessary(String expression) {
+          if (typeSource == null) return expression;
+          return '($expression) as $typeSource';
+        }
+
         String generatedAccessor = '$docComment'
             // TODO reinstate inlining once https://github.com/dart-lang/sdk/issues/36217 is fixed and all workarounds are removed
             // inlining is necessary to get mixins to use $index/$indexSet instead of $index$asx
             // '  @tryInline\n'
             '  @override\n'
             '$metadataSrc'
-            '  ${typeString}get $accessorName => $proxiedMapName[$keyConstantName] ?? null;'
-            ' // Add ` ?? null` to workaround DDC bug: <https://github.com/dart-lang/sdk/issues/36052>;\n'
+            // Add ` ?? null` to workaround DDC bug: <https://github.com/dart-lang/sdk/issues/36052
+            '  ${typeString}get $accessorName => ${castGetterMapValueIfNecessary('$proxiedMapName[$keyConstantName] ?? null')};\n'
             '$docComment'
             // '  @tryInline\n'
             '  @override\n'
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -270,7 +270,9 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
             // '  @tryInline\n'
             '  @override\n'
             '$metadataSrc'
-            // Add ` ?? null` to workaround DDC bug: <https://github.com/dart-lang/sdk/issues/36052
+            // Add ` ?? null` to work around DDC bug: <https://github.com/dart-lang/sdk/issues/36052
+            // Apply this workaround ASAP, before the cast, to limit where undefined can leak into
+            // and potentially cause issues (for instance, DDC cast internals).
             '  ${typeString}get $accessorName => ${castGetterMapValueIfNecessary('$proxiedMapName[$keyConstantName] ?? null')};\n'
             '$docComment'
             // '  @tryInline\n'
--- a/lib/src/util/safe_render_manager/safe_render_manager.dart
+++ b/lib/src/util/safe_render_manager/safe_render_manager.dart
@@ -192,8 +192,8 @@ class SafeRenderManager extends Disposable {
     }
   }
 
-  void _helperRef(ref) {
-    _helper = ref;
+  void _helperRef(dynamic ref) {
+    _helper = ref as SafeRenderManagerHelperComponent;
     if (_helper != null) {
       if (_state == _RenderState.mounting) {
         _state = _RenderState.mountedOrErrored;
@@ -222,7 +222,8 @@ class SafeRenderManager extends Disposable {
         // an indefinitely long lifetime.
         completer = null;
       });
-    }, onError: (error, stackTrace) {
+    // ignore: avoid_types_on_closure_parameters
+    }, onError: (error, StackTrace stackTrace) {
       completer?.completeError(error, stackTrace);
       // Clear out to not retain it in the onError closure, which has
       // an indefinitely long lifetime.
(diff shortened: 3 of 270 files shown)


aviary-wf:
## Security Insights

## Action Items
---

analysis_options.yaml
```diff
@@ -1,6 +1,8 @@
 include: package:pedantic/analysis_options.1.8.0.yaml
 
 analyzer:
+  strong-mode:
+    implicit-casts: false
```
greglittlefield-wf, line 5: For reviewing, I recommend looking at the commits separately so that generated file changes don't clog up the diff

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -259,14 +259,19 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
...
+        }
+
         String generatedAccessor = '$docComment'
             // TODO reinstate inlining once https://github.com/dart-lang/sdk/issues/36217 is fixed and all workarounds are removed
             // inlining is necessary to get mixins to use $index/$indexSet instead of $index$asx
             // '  @tryInline\n'
             '  @override\n'
             '$metadataSrc'
-            '  ${typeString}get $accessorName => $proxiedMapName[$keyConstantName] ?? null;'
-            ' // Add ` ?? null` to workaround DDC bug: <https://github.com/dart-lang/sdk/issues/36052>;\n'
```
greglittlefield-wf, line 269: I took this comment out of the generated code since it cluttered up the output, and seemed more appropriate in the code generation logic itself

.github/workflows/dart_ci.yml
```diff
@@ -2,9 +2,12 @@ name: Dart CI
 
 on:
   push:
-    branches: [ master ]
```
greglittlefield-wf, line 5: These changes are just merged in from https://github.com/Workiva/over_react/pull/670

analysis_options.yaml
greglittlefield-wf, line 5: Reminder: for reviewing, I recommend looking at the commits separately so that generated file changes don't clog up the diff

test/over_react_redux/value_mutation_checker_test.dart
```diff
@@ -22,7 +22,7 @@ import 'package:test/test.dart';
 main() {
   group('Value Mutation Checker:', () {
     group('CollectionLengthHasher', () {
-      sharedHashTests(useLengthHasher: false);
```
greglittlefield-wf, line 25: This should have been `true`, so shared tests for this class weren't actually being run

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -259,14 +259,19 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
...
+            // Add ` ?? null` to workaround DDC bug: <https://github.com/dart-lang/sdk/issues/36052
+            '  ${typeString}get $accessorName => ${castGetterMapValueIfNecessary('$proxiedMapName[$keyConstantName] ?? null')};\n'
```
sydneyjodon-wk, line 274: why is `?? null` also being type casted?

test/over_react_redux/value_mutation_checker_test.dart
```diff
@@ -148,9 +147,6 @@ void sharedHashTests({bool useLengthHasher}) {
 
       map['newField'] = true;
       expect(hasher.hasHashChanged(map), isTrue);
-
-      map['test'] = false;
-      expect(hasher.hasHashChanged(map), isTrue);
```
sydneyjodon-wk, line 153: What was the reason for deleting this?

sydneyjodon-wk (commented):
Looks great! I just had a couple minor questions.

test/over_react/component/_deprecated/abstract_transition_test.dart
```diff
@@ -753,6 +753,6 @@ class TransitionerComponent extends AbstractTransitionComponent<TransitionerProp
   @override
   void setState(dynamic newState, [Function() callback]) {
     super.setState(newState, callback);
-    transitionPhasesSet.add(newState.transitionPhase);
+    transitionPhasesSet.add(typedStateFactory(newState as Map).transitionPhase);
   }
```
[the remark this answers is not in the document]
greglittlefield-wf: Good eye! It's because in `Component`, the arg is typed as `dynamic` (since it can take either a Map or an updater function), whereas in `Component2` we made it only accept `Map` and added `setStateWithUpdater`.

lib/src/util/safe_render_manager/safe_render_manager.dart
```diff
@@ -192,8 +192,8 @@ class SafeRenderManager extends Disposable {
     }
   }
 
-  void _helperRef(ref) {
-    _helper = ref;
+  void _helperRef(dynamic ref) {
+    _helper = ref as SafeRenderManagerHelperComponent;
```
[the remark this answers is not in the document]
greglittlefield-wf: No reason, I believe that works as well!

I'm not quite sure I did it this way; I think we never really talked about what the preferred syntax should be in implicit-cast-disabled repos, so I wasn't sure which one to pick.

test/over_react_redux/value_mutation_checker_test.dart
greglittlefield-wf: Ah, I should have commented on this before, sorry!

This test suite was previously being run on the `CollectionShallowHasher` twice instead of once on it and once on `CollectionLengthHasher`, and this test was written in a way that it failed for `CollectionLengthHasher`. I assumed this was a mistake that just never got caught since tests weren't running properly, so I removed it.

And since what this expectation was exercising—that the hash changes when updating a value—is already being tested by the `hash` tests in the non-shared suite, I didn't think I needed to add any additional coverage.

test/over_react/component/_deprecated/abstract_transition_test.dart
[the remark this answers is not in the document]
sydneyjodon-wk: ohh that makes sense!

lib/src/builder/codegen/accessors_generator.dart
greglittlefield-wf: I figured applying the DDC workaround ASAP would be more robust, so that any potential `undefined` value can't make its way into DDC runtime cast logic and cause issues (for example, if the DDC implementation changes in a newer SDK version and suddenly doesn't work well with `undefined`).

That's a good question, though; I can add that info in a comment!

sydneyjodon-wk (approved):
+1

lib/src/builder/codegen/accessors_generator.dart
greglittlefield-wf: Added in https://github.com/Workiva/over_react/pull/669/commits/79ef12a8c13737cf47cb1999021fc32c0b7e5c7b

joebingham-wk (approved):
+1 to 79ef12a... QA'ing in progress!

joebingham-wk (commented):
QA +1
- [x] Smoke test web / example
- [x] CI is passing (runs that failed are expected to)

PR #674, opened 2021-01-20
CPLAT-11863 Add Redux Hooks

aaronlademann-wf:
## Motivation
The react-redux JS package that we wrap / expose from our `over_react_redux.dart` library offers function component hook APIs that we don't have wrapped for our consumers.

## Core Changes
1. Add `useSelector`, `createSelectorHook`, `useDispatch`, `createDispatchHook`, `useStore` and `createStoreHook` top level functions, along with the necessary JS interop pieces to call into the analogous [JS hooks](https://react-redux.js.org/api/hooks).
1. Add examples, tests and documentation.

## Misc Changes
1. Move over_react_redux test utils into a directory within `src/` so we can reuse some of it in our `example/`s

#### Release Notes
Add React Redux hooks for function components


aviary3-wk:
## Security Insights

## Action Items
---

test/over_react_redux/hooks/use_dispatch_test.dart
```diff
@@ -0,0 +1,65 @@
...
+      jacket = null;
+
+      counterStore = Store(counterStateReducer, initialState: CounterState());
+      bigCounterStore = Store(bigCounterStateReducer, initialState: BigCounterState(bigCount: 9));
+    });
+
+    test('dispatches an action', () async {
+      jacket = mount(
+        (ReduxProvider()..store = counterStore)(
+          CounterFn()(),
```
greglittlefield-wf, line 41: I'm concerned that this shared component could be accidentally updated to not use `useDispatch`. Could we create a copy of the component local to this test, or add a comment in that implementation saying that this test depends on it?

test/over_react_redux/hooks/use_selector_test.dart
```diff
@@ -0,0 +1,169 @@
...
+          expectCountValue(jacket, 0);
+          await clickSiblingConnectedIncrementButton(jacket);
+          expectCountValue(jacket, 1);
+        });
+
+        test('unless a custom equalityFn returns true', () async {
+          jacket = mount(
+            (ReduxProvider()..store = counterStore)(
+              (CounterFn()..countEqualityFn = (nextCount, prevCount) {
+                return nextCount == 1;
```
greglittlefield-wf, line 64: there's no test for when equalityFn returns false; could we make this `2` so that it rerenders at least once?

test/over_react_redux/hooks/use_selector_test.dart
```diff
@@ -0,0 +1,169 @@
...
+    group('subscribes to custom context when createSelectorHook is used and redraws when', () {
+      group('the selector value is updated', () {
+        test('', () async {
+          jacket = mount(
+              (ReduxProvider()..store = counterStore)(
+                (ReduxProvider()
+                  ..store = bigCounterStore
+                  ..context = bigCounterContext
+                )(
+                  CustomContextCounterFn()(),
```
greglittlefield-wf, line 120: I will say that using shared component code, in another file, makes it harder to tell at a glance how exactly these things are being tested, since you have to dig to find the `createSelectorHook` call as opposed to it being in this test/group block

test/over_react_redux/hooks/use_store_test.dart
```diff
@@ -0,0 +1,61 @@
...
+
+import '../../test_util/test_util.dart';
+import 'package:over_react/src/over_react_redux/test_fixtures/counter_fn.dart';
+import 'package:over_react/src/over_react_redux/test_fixtures/store.dart';
+
+import 'utils.dart';
+
+main() {
+  group('useStore hook', () {
+    test('subscribes to the store and redraws when it updates', () async {
```
greglittlefield-wf, line 27: This component also uses useSelector, which causes it to redraw, so this test can't actually test if useStore makes the component redraw.

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,183 @@
...
+///     MyComponent()(),
+///   );
+///
+///   react_dom.render(app, querySelector('#id_of_mount_node'));
+/// }
+/// ```
+_SelectorFnHook<TReduxState> createSelectorHook<TReduxState>([Context<TReduxState> context]) {
+  final jsHook = _jsCreateSelectorHook(context?.jsThis ?? JsReactRedux.ReactReduxContext);
+
+  return <TSelector>(selector, [equalityFn]) {
```
greglittlefield-wf, line 159: #nit it would be nice if this were a local function, to provide better stack traces if something throws

lib/src/over_react_redux/hooks/use_store.dart
```diff
@@ -0,0 +1,68 @@
...
+/// use custom [context].
+///
+/// This is useful if you're building a complex reusable component, and you don't want your [Store] to collide
+/// with any Redux [Store] your consumers' applications might use.
+///
+/// See the [react-redux JS documentation](https://react-redux.js.org/api/hooks#custom-context) for more details.
+///
+/// See the [createSelectorHook] documentation for an example of creating / using custom context.
+Store<TReduxState> Function() createStoreHook<TReduxState>([Context<TReduxState> context]) {
+  return () {
```
greglittlefield-wf, line 60: Same as above on local functions

doc/over_react_redux_documentation.md
```diff
@@ -275,6 +279,221 @@ UiFactory<CounterProps> Counter = connect<CounterState, CounterProps>(
...
+final useMySelector = createSelectorHook(MyContext);
+final useMyDispatch = createDispatchHook(MyContext);
+final useMyStore = createStoreHook(MyContext);
+
+final myStore = Store(myReducer);
+
+// ------------------------------------
+//  2. (Optional) Create a custom
+//  provider component that uses the
+//  context defined above.
```
greglittlefield-wf, line 446: #nit this seems like it provides limited value and adds a bit of boilerplate, and may not be worth the cognitive overhead it adds for the reader (same goes for other examples)

doc/over_react_redux_documentation.md
```diff
@@ -275,6 +279,221 @@ UiFactory<CounterProps> Counter = connect<CounterState, CounterProps>(
...
+// ------------------------------------
+//  3. Create a function component that
+//  uses `useMySelector`.
+// ------------------------------------
+mixin MyComponentProps on UiProps {}
+
+UiFactory<MyComponentProps> MyComponent = uiFunction(
+  (props) {
+    final count = useMySelector((state) => state.count);
+    final store = useMyStore();
```
greglittlefield-wf, line 472: This hook and its initialization usage above should have similar warnings about not using useStore hooks

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,183 @@
...
+/// ```dart
+/// import 'package:over_react/over_react.dart';
+/// import 'package:over_react/over_react_redux.dart';
+/// import 'counter_state.dart';
+///
+/// mixin CounterProps on UiProps {}
+///
+/// UiFactory<CounterProps> Counter = uiFunction(
+///   (props) {
+///     final count = useSelector<int, CounterState>((state) => state.count);
```
greglittlefield-wf, line 59: I feel like having the generic parameters the other way around feels more natural:
```dart
// Current:
final count = useSelector<int, CounterState>((state) => state.count);
// My preference
final count = useSelector<CounterState, int>((state) => state.count);
```

I'd be curious to see what other people think, though.

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,183 @@
...
+///
+///     return Dom.div()('The current count is $count');
+///   },
+///   $CounterConfig, // ignore: undefined_identifier
+/// );
+/// ```
+///
+/// > Related: [useDispatch]
+TSelector useSelector<TSelector, TReduxState>(TSelector Function(TReduxState state) selector, [
```
greglittlefield-wf, line 68: I think a better name for `TSelector` would be `TValue`, as in the value you're selecting

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,183 @@
...
+///
+/// // ------------------------------------
+/// //  3. Create a function component that
+/// //  uses `useMySelector`.
+/// // ------------------------------------
+/// mixin MyComponentProps on UiProps {}
+///
+/// UiFactory<MyComponentProps> MyComponent = uiFunction(
+///   (props) {
+///     final count = useMySelector<int, MyState>((state) => state.count);
```
greglittlefield-wf, line 137: I think this should be
```suggestion
///     final count = useMySelector((state) => state.count);
```

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,183 @@
...
+/// ```dart
+/// import 'package:over_react/over_react.dart';
+/// import 'package:over_react/over_react_redux.dart';
+/// import 'package:redux/redux.dart';
+///
+/// // ------------------------------------
+/// //  1. Declare the custom context
+/// // ------------------------------------
+/// final MyContext = createContext<MyState>();
+/// final useMySelector = createSelectorHook(MyContext);
```
greglittlefield-wf, line 106: I wonder if we should should recommend shadowing the top-level useSelector/useStore/etc hooks in our doc comments and examples, for consistency with what they show in React Redux https://react-redux.js.org/api/hooks#custom-context

greglittlefield-wf (changes requested):
I had some comments, but this looks awesome!! I'm so excited for these hooks to be available!

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,183 @@
...
+/// // ------------------------------------
+/// main() {
+///   final app = MyProvider()(
+///     MyComponent()(),
+///   );
+///
+///   react_dom.render(app, querySelector('#id_of_mount_node'));
+/// }
+/// ```
+_SelectorFnHook<TReduxState> createSelectorHook<TReduxState>([Context<TReduxState> context]) {
```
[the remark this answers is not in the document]
aaronlademann-wf: It is nicer, but there is the extra step of creating the hook, whereas the `useSelector` hook is a "catchall" for the default `Context`.

doc/over_react_redux_documentation.md
aaronlademann-wf: Yeah, I was just trying to make our example as close to the react-redux docs as possible.

lib/src/over_react_redux/hooks/use_selector.dart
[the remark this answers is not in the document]
aaronlademann-wf: It also runs the risk of them exporting the shadowed entity unintentionally, which could create confusion for consumers.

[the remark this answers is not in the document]
aaronlademann-wf: @greglittlefield-wf I tried removing the generic param from `Context` as you suggested above - but it makes it so that the typing is not inferred correctly for consumers.

Screen Shot 2021-01-22 at 7 36 26 AM

lib/src/over_react_redux/over_react_redux.dart
```diff
@@ -508,13 +508,21 @@ JsReactReduxStore _reduxifyStore(Store store) {
...
 @JS()
 @anonymous
 class JsReactReduxStore {
+  /// The Dart Redux `Store` instance.
+  ///
+  /// Set once within [_reduxifyStore], and should only be used internally.
+  ///
+  /// > See: `createStoreHook` / `useStore`.
+  @protected
+  Store dartStore;
```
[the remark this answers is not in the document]
aaronlademann-wf: This seems to cause problems with the hook generated by calling `createStoreHook`...

```
NoSuchMethodError: tried to call a non-function, such as null: 'counter_fn.useBigCountStore'
```

This is what I have... did I miss something? I'm not really familiar with `external` - so it feels like magic to me that `dartStore` would ever have a value as a result of passing a value to the `dartStore` constructor arg 🤷 

```dart
@JS()
@anonymous
class JsReactReduxStore {
  /// The Dart Redux `Store` instance.
  ///
  /// Set once within [_reduxifyStore], and should only be used internally.
  ///
  /// > See: `createStoreHook` / `useStore`.
  external Store get dartStore;

  external factory JsReactReduxStore({
    ReactInteropValue Function() getState,
    Dispatcher dispatch,
    Function Function(Function) subscribe,
    Store dartStore,
  });
}
```

doc/over_react_redux_documentation.md
greglittlefield-wf: Oh, okay, well if they include it, I suppose that's fine (though I don't agree with them 🙂 )

lib/src/over_react_redux/hooks/use_selector.dart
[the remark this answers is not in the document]
greglittlefield-wf: > It is nicer, but there is the extra step of creating the hook, whereas the useSelector hook is a "catchall" for the default Context.

Yeah, I was thinking it'd be worth it though in a lot of cases to take that extra step. For instance:
```dart
final useMySelector = useSelector<MyState>();

... = uiFunction(props) {
  final foo = useMySelector((state) => state.foo);
  final bar = useMySelector((state) => state.bar);
  final baz = useMySelector((state) => state.baz);
}...
```
vs
```dart
... = uiFunction(props) {
  final foo = useSelector<MyState, int>((state) => state.foo);
  final bar = useSelector<MyState, String>((state) => state.bar);
  final baz = useSelector<MyState, Baz>((state) => state.baz);
}...
```

I think we should at call it out prominently as an option to help avoid having to specifying the types.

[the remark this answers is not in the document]
greglittlefield-wf: > It also runs the risk of them exporting the shadowed entity unintentionally, which could create confusion for consumers.
Yeah, that's fair; I suppose they can decide to name it whatever they want.


>  I tried removing the generic param from Context as you suggested above - but it makes it so that the typing is not inferred correctly for consumers.

Did you retain the generic parameter in the method and explicitly add it to the call site (since it no longer gets inferred from the context generic parameters)?
```diff
-final useMySelector = createSelectorHook(MyContext);
+final useMySelector = createSelectorHook<MyState>(MyContext);
```

> Since Redux uses the redux Store.state as the value for Context.Provider.value, it seems like the generic arg should probably be something like this:

My understanding is that Redux stores the store itself in context, not the state. See https://react-redux.js.org/api/provider#provider for more info

And like I mentioned above, it's not the Dart store that's stored directly in the context. I thought it was the JS store, but it looks like it's actually an internal Redux object wrapped around the JS store. Take this example:

```dart
      (ReduxProvider()
        ..store = store
        ..context = MyContext
      )(
      MyContext.Consumer()((value) {
        window.console.log('context value');
        window.console.log(value);
      }),
```
Screen Shot 2021-01-22 at 9 25 30 AM

So, typing it as the Dart store would be incorrect and lead to potential runtime type errors.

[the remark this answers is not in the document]
aaronlademann-wf: > Did you retain the generic parameter in the method and explicitly add it to the call site (since it no longer gets inferred from the context generic parameters)?

Nope... that works.

lib/src/over_react_redux/test_fixtures/store.dart
```diff
@@ -53,8 +72,21 @@ Reducer<int> counterActionsReducer = combineReducers<int>([
   TypedReducer<int, ResetAction>(_resetCounterReducer),
 ]);
 
+Reducer<DartModelCounter> modelCounterActionsReducer = combineReducers<DartModelCounter>([
+  TypedReducer<DartModelCounter, ResetAction>((currentModel, action) {
+    return DartModelCounter(count: 0);
+  }),
+  TypedReducer<DartModelCounter, IncrementModelCountAction>((currentModel, action) {
+    return DartModelCounter(count: currentModel.count + (action?.value != null ? action.value : 1));
```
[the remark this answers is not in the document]
aaronlademann-wf: When I make this change, a `Uncaught Error: NoSuchMethodError: '<Unexpected Null Value>'` exception is thrown.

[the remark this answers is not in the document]
greglittlefield-wf: My bad, I messed up the order of operations and was missing parens 
```suggestion
    return DartModelCounter(count: currentModel.count + (action.value ?? 1));
```

aaronlademann-wf:
@greglittlefield-wf all feedback has been addressed.

example/hooks/over_react_redux/counter.dart
```diff
@@ -0,0 +1,85 @@
...
+
+import 'package:over_react/over_react.dart';
+import 'package:over_react/over_react_redux.dart';
+
+import 'redux_actions.dart';
+
+part 'counter.over_react.g.dart';
+
+@Factory()
+UiFactory<CounterProps> Counter = _$Counter;
```
joebingham-wk, line 23: This component appears to be unused

example/hooks/over_react_redux/redux_actions.dart
```diff
@@ -0,0 +1,48 @@
+// Copyright 2020 Workiva Inc.
```
joebingham-wk, line 1: #nit

```suggestion
// Copyright 2021 Workiva Inc.
```

example/hooks/over_react_redux/redux_actions.dart
```diff
@@ -0,0 +1,48 @@
...
+
+class DecrementModelCountAction extends Action {
+  DecrementModelCountAction([value]) : super(type: 'DECREMENT_MODEL_COUNT', value: value);
+}
+
+class ResetAction extends Action {
+  ResetAction() : super(type: 'RESET');
+}
+
+class MutateStoreDirectlyAction extends Action {
```
joebingham-wk, line 46: This action appears to be unused

example/hooks/over_react_redux/store.dart
```diff
@@ -0,0 +1,144 @@
+// Copyright 2020 Workiva Inc.
```
joebingham-wk, line 1: #nit
```suggestion
// Copyright 2021 Workiva Inc.
```

example/hooks/use_selector_example.dart
```diff
@@ -0,0 +1,31 @@
+import 'package:over_react/over_react.dart';
```
joebingham-wk, line 1: #nit this file could use the OSS license

test/over_react_redux/hooks/use_selector_test.dart
```diff
@@ -0,0 +1,185 @@
...
+          expectBigCountValue(jacket, 9);
+          await clickSiblingConnectedBigCountIncrementButton(jacket);
+          expectBigCountValue(jacket, 109);
+
+          expectCountValue(jacket, 0);
+          await clickSiblingConnectedIncrementButton(jacket);
+          expectCountValue(jacket, 1);
+        });
+
+        test('unless a custom equalityFn returns false', () async {
```
joebingham-wk, line 152: Should this be when the equality function is true?
```suggestion
        test('unless a custom equalityFn returns true', () async {
```

joebingham-wk (commented):
Everything looks really good to me - the nesting on those hooks does not look fun haha. I only found nits!

aaronlademann-wf:
All feedback addressed @joebingham-wk!

joebingham-wk (approved):
LGTM!

lib/src/over_react_redux/hooks/use_selector.dart
```diff
@@ -0,0 +1,228 @@
...
+///   )(
+///     MyComponent()(),
+///   );
+///
+///   react_dom.render(app, querySelector('#id_of_mount_node'));
+/// }
+/// ```
+_SelectorFnHook<TReduxState> createSelectorHook<TReduxState>([Context context]) {
+  final jsHook = _jsCreateSelectorHook(context?.jsThis ?? JsReactRedux.ReactReduxContext);
+  // Using a var instead of a named function here to avoid having to redeclare the type on the LHS
```
greglittlefield-wf, line 200: Can we make this a named function? I think the benefit of getting better stack traces is worth having to redeclare the types.

doc/over_react_redux_documentation.md
```diff
@@ -275,6 +279,270 @@ UiFactory<CounterProps> Counter = connect<CounterState, CounterProps>(
 
 > [More information about the `connect` function](https://react-redux.js.org/api/connect#connect)
 
+## Hooks
+OverReact offers a set of hook APIs as an alternative to the existing [`connect()`](#connect) Higher Order Component. 
+These APIs allow you to subscribe to the Redux store and dispatch actions, without having to wrap your components 
+in [`connect()`](#connect).
```
greglittlefield-wf, line 285: We should clarify these APIs' relationship to React Redux and link to that documentation up front, similar to the `connect` section:

```suggestion
OverReact exposes wrappers around React Redux hook APIs, which serve as an alternative to the existing [`connect()`](#connect) Higher Order Component. 
These APIs allow you to subscribe to the Redux store and dispatch actions, without having to wrap your components 
in [`connect()`](#connect).

> See: https://react-redux.js.org/api/hooks
```

doc/over_react_redux_documentation.md
```diff
@@ -275,6 +279,270 @@ UiFactory<CounterProps> Counter = connect<CounterState, CounterProps>(
...
+
+The use of this hook will also subscribe your component to the Redux `Store`, and run your `selector` whenever
+an action is dispatched.
+
+* The `selector` should be pure since it is potentially executed multiple times and at arbitrary points in time.
+* The `selector` is approximately equivalent to the `mapStateToProps` argument of [`connect`](#connect) conceptually.
+* The `selector` will be called with the entire Redux `Store` state as its only argument.
+* The `selector` will be run whenever the function component renders.
+
+By default, the return value of `selector` is compared using strict (`===`) equality. If you want to
```
greglittlefield-wf, line 323:
```suggestion
By default, the return value of `selector` is compared using strict JavaScript (`===`) equality. If you want to
```

doc/over_react_redux_documentation.md
```diff
@@ -275,6 +279,270 @@ UiFactory<CounterProps> Counter = connect<CounterState, CounterProps>(
...
+```
+
+> __CAUTION:__ Be sure to not export the shadowed values of `useSelector`, `useDispatch` or `useStore` unless you know 
+  exactly what you're doing, and the consumers of your library expect the hook to always have the context of the 
+  state you parameterize it with.
+
+> See the [react-redux JS documentation](https://react-redux.js.org/api/hooks#custom-context) for more details.
+
+
 ## Using Multiple Stores
```
greglittlefield-wf, line 546: Can we update this section to also mention / link to the "Custom Context For Hooks" section for consumers using hooks?

greglittlefield-wf (changes requested):
Just a couple more small things

aaronlademann-wf:
@greglittlefield-wf all feedback addressed.

greglittlefield-wf (approved):

greglittlefield-wf:
QA +1

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #696, opened 2021-07-22
Fix analyzer plugin build in Dart stable

greglittlefield-wf:
## Motivation
  

The build for the analyzer plugin was failing in Dart 2.13 with errors that seemed like null safety was being incorrectly opted into.

Example: https://github.com/Workiva/over_react/runs/3306575416

I forget exactly why, but increasing analyzer_plugin (and thus increasing the resolvable upper bound of `analyzer`) seem to fix the issue.

## Changes
- Increase upper bound of analyzer_plugin (thus also allowing newer versions of `analyzer`)
- Fix issues caused by analyzer breakages 
    - (e.g., analyzer 0.40.0 made `getDisplayString`'s `withNullability` arg required)
- Address deprecated API usages
- Copy over analyzer APIs imported from `src` directories that were removed/moved in newer analyzer versions 
- Update build to only run the format check in a single Dart version
- Remove mockito dependency since its builder was timing out test runs in newer Dart SDKs for some reason

__Note this gets CI passing, but for some reason, diagnostics aren't showing up in the Dart Analysis panel of IDEs 😕 .__ Other things like suggested edits (e.g., "Add ref") work as expected. The `AnalysisErrorsParams` notification being sent by the plugin seems to be getting constructed properly with the correct errors and gets sent, but for some reason the analysis server isn't piping it through... **Update:** upgrading to analyzer_plugin ^0.6.0 (and analyzer ^1.7.0) fixes the issue! We'll do that in a separate PR, since upgrading those dependencies in over_react and over_react_analyzer_plugin is going to take some effort.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        - [ ] Analyzer plugin job runs tests, and passes in all SDKs

## Merge Checklist


aviary2-wf:
## Security Insights

## Action Items
---

.github/workflows/dart_ci.yml
```diff
@@ -94,13 +97,23 @@ jobs:
...
       - id: install
         name: Install dependencies
         run: pub get
+        if: always() && steps.link.outcome == 'success'
 
-      - name: Run analyzer_plugin checks
-        run: |
-          dart tool/travis_link_plugin_deps.dart
-          cd ./tools/analyzer_plugin
-          ./tool/travis.sh
```
greglittlefield-wf, line 105: I broke this up so that we could more easily conditionally run formatting step

.github/workflows/dart_ci.yml
```diff
@@ -94,13 +97,23 @@ jobs:
...
+      - name: Validate dependencies
+        run: pub run dependency_validator --no-fatal-pins -i analyzer,build_runner,build_web_compilers,built_value_generator
```
evanweible-wf, line 110: fyi if you want to upgrade to dependency_validator v2 or v3, then it should detect the executable from `build_runner` and the builders from the last two, and then you'd only need to ignore `analyzer` which you could do statically. You'd also have to drop `--no-fatal-pins` and ignore whichever packages are ignored.

tools/analyzer_plugin/lib/src/util/ast_util.dart
```diff
@@ -3,15 +3,12 @@ library over_react_analyzer_plugin.src.ast_util;
 
 import 'dart:collection';
 import 'dart:mirrors';
-// This is necessary for `ConstantEvaluator`. If that API is removed, it can just
-// be copied and pasted into this analyzer package (if still needed).
-// ignore: deprecated_member_use
-import 'package:analyzer/analyzer.dart' show ConstantEvaluator;
```
evanweible-wf, line 9: Oh, looks like this answers my previous question. Did they replace this with an alternative or a better option?

.github/workflows/dart_ci.yml
evanweible-wf: (probably out-of-scope since this is a small PR, just thought I'd throw it out there for future reference)

aaronlademann-wf (approved):
+1

aaronlademann-wf:

rmconsole-wf (approved):
+1 from RM

PR #703, opened 2021-08-27
Improve FluxUiComponent disposed store logging

greglittlefield-wf:
## Motivation
  
We have some warning logs in place for before FluxUiComponent(2) components attempt to subscribe to disposed stores.

These logs weren't very helpful, though, since they didn't contain a stack trace, component name, or store name.

## Changes
- Add more information to logs
    - Add a utility to get a Dart component name from its Dart instance
- Add basic unit tests for this logging

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -21,6 +21,7 @@ import 'dart:async';
 import 'package:logging/logging.dart';
 import 'package:meta/meta.dart';
 import 'package:over_react/component_base.dart' as component_base;
+import 'package:over_react/src/util/component_debug_name.dart';
 import 'package:w_flux/w_flux.dart';
 
 import 'annotations.dart' as annotations;
@@ -235,10 +236,12 @@ mixin _FluxComponentMixin<TProps extends FluxUiProps> on component_base.UiCompon
 
   void _validateStoreDisposalState(Store store) {
     if (store.isOrWillBeDisposed) {
+      final componentName = getDebugNameForDartComponent(this);
+
       // Include the component name in the logger so that
       // 1. it's included in the log somewhere
       // 2. logs from the same component can be easily grouped together
-      final logger = Logger('over_react._FluxComponentMixin.$displayName');
+      final logger = Logger('over_react._FluxComponentMixin.$componentName');
 
       var storeNameOrType = store.disposableTypeName;
       // Detect if they don't override disposableTypeName and use the default name.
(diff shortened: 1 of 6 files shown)


lib/src/component_declaration/flux_component.dart
```diff
@@ -228,24 +229,40 @@ abstract class FluxUiStatefulComponent2<TProps extends FluxUiProps, TState exten
...
-  static final Logger _logger = Logger('over_react._FluxComponentMixin');
-
   /// List of store subscriptions created when the component mounts.
   ///
   /// These subscriptions are canceled when the component is unmounted.
   List<StreamSubscription> _subscriptions = [];
 
   void _validateStoreDisposalState(Store store) {
-    String message = 'Cannot listen to a disposed/disposing Store.';
+    if (store.isOrWillBeDisposed) {
```
greglittlefield-wf, line 238: I wrapped both the assert and the log in this check so we could conditionally build a shared message instead of computing it unnecessarily. The control-flow of this method should be the same as it was before.

lib/src/component_declaration/flux_component.dart
```diff
@@ -228,24 +229,40 @@ abstract class FluxUiStatefulComponent2<TProps extends FluxUiProps, TState exten
...
+
+      var storeNameOrType = store.disposableTypeName;
+      // Detect if they don't override disposableTypeName by checking for the default name.
+      if (storeNameOrType == 'Store') {
+        storeNameOrType = store.runtimeType.toString();
+      }
 
-    var isDisposedOrDisposing = store.isOrWillBeDisposed ?? false;
+      final message = 'Cannot listen to a disposed/disposing Store.'
+          ' (storeNameOrType: $storeNameOrType, shouldBatchRedraw: $shouldBatchRedraw)';
```
todbachman-wf, line 253: Would it be better to use a `ContextualMessage` and put `storeNameOrType` and `shouldBatchRedraw` in the context of the message instead?

greglittlefield-wf: I really wanted to do that, but unfortunately I'd have to depend on app_intelligence_dart to get that class (which we can't do since this is an OSS repo, but even if it weren't I'd be wary about).

It'd be nice if you could pass in a `Map` or some other core type and have it treat it the same as a `ContextualMessage`.

todbachman-wf: Ahh. I didn't realize that was an AID-specific class.

todbachman-wf (approved):

todbachman-wf:
QA +1. The unit tests added should be sufficient for this change.

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #730, opened 2022-02-03
CPLAT-16997 Fix selector hooks redrawing when values haven't changed

greglittlefield-wf:
## Motivation
The `useSelector` hook and hooks created by `createSelectorHook` allow consumers to subscribe to store updates and rerender the component only when a value returned from the selector function changes.

However, the Dart versions of these APIs were always rerendering, even when the selected value didn't change.

This was caused by two issues:
1. We were always wrapping the selected value in a new `ReactInteropValue` object, which meant that when you didn't provide a custom equality function, the default behavior on the JS side of comparing the value using `===` would always think the value changed
1. We were always passing `null` into the hook as an equality function as opposed to conditionally calling it, which was causing null errors in React Redux JS that [were getting ignored](https://github.com/reduxjs/react-redux/blob/86e962edf8ed077cf720e8e089876227d943dfe8/src/hooks/useSelector.js#L70)

## Changes
- Instead of always wrapping in an interop value, conditionally wrap only when necessary (Dart functions), just like we do in react-dart's JsBackedMap
- Conditionally pass the equality function instead of always passing in null
- Add regression tests to ensure renders only occur when we expect them to, for both hooks and connect
- Add test coverage for values of different types, which also tests conditional wrapping of values
- Add useSelector example

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        - Serve the examples locally `webdev serve`
        - Go to <http://localhost:8080/over_react_redux/examples/simple/>
        - Start a React Dev Tools profile, click one of the `BigCounter` buttons, and stop the profile
        - Verify that only the "BigCounter" components rerender, and not the "Counter" or "HookCounter" components

## Merge Checklist


aviary-wf:
## Security Insights

## Action Items
---

lib/src/component/pure_component_mixin.dart
```diff
@@ -27,8 +27,7 @@ import 'package:over_react/over_react.dart';
 ///   list of children and pass that in as shown in the example below:
 ///
 /// ```dart
-/// import 'package:memoize/memoize.dart';
-///
+/// // imemo1 is from the `memoize` package
```
greglittlefield-wf, line 30: I changed this to work around dependency_validator flagging imports in comments

lib/src/over_react_redux/over_react_redux.dart
```diff
@@ -497,14 +501,9 @@ class ReactJsReactReduxComponentFactoryProxy extends ReactJsContextComponentFact
 
 /// Converts a Redux.dart [Store] into a Javascript object formatted for consumption by react-redux.
 JsReactReduxStore _reduxifyStore(Store store) {
-  // Memoize this so that the same ReactInteropValue instances will be used
-  // for a given state, allowing JS `===` checks to not fail when the same
-  // state object is passed.
```
greglittlefield-wf, line 502: Since we pass through the value directly (and memoize it in the cases we wrap it in `DartValueWrapper.wrapIfNeeded`), this is no longer necessary

test/over_react_redux/hooks/use_selector_test.dart
```diff
@@ -1,183 +0,0 @@
...
-          expectBigCountValue(jacket, 109, reason: 'Component should not update when equalityFn returns true');
-
-          expectCountValue(jacket, 0);
-          await clickSiblingConnectedIncrementButton(jacket);
-          expectCountValue(jacket, 1);
-        });
-      });
-    });
-  });
-}
```
greglittlefield-wf, line 183: These tests have moved to store_bindings_tests.dart. I started by refactoring them to use custom components so they could be run with multiple types of selected values, and they eventually morphed into tests that could also be shared with connect, thus the new file name.

test/over_react_redux/store_bindings_tests.dart
```diff
@@ -0,0 +1,420 @@
+// Copyright 2022 Workiva Inc.
```
greglittlefield-wf, line 1: I'm open to better name suggestions for this file

aaronlademann-wf (approved):
+10

test/over_react_redux/store_bindings_tests.dart
```diff
@@ -0,0 +1,420 @@
...
+  String Function(T) renderValue,
+}) {
+  renderValue ??= (value) => value.toString();
+
+  group(name, () {
+    setUpAll(() {
+      final allValues = [initialValue, updatedValue1, updatedValue2];
+      if (allValues.toSet().length != allValues.length) {
+        throw ArgumentError('initialValue and updatedValues must all be unique.');
+      }
```
joebingham-wk, line 99: Oohh this is clever. TIL

joebingham-wk (approved):
+1, looks great! Those are some dope tests 🎉

KealJones (approved):
+1

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #733, opened 2022-02-15
SafeRenderManager retention workaround

greglittlefield-wf:
## Motivation
SafeRenderManager gets retained by the rendered SafeRenderManagerHelper, which in some cases was being retained by React, seemingly because of the following behavior:

_From https://github.com/facebook/react/issues/16087:_
> Props/child trees retained by alternate children. Similarly, children that was just removed can sometimes between retained by the alternate copy of that. That is until that node gets another update on it which clears out the old children. These cases are fairly unusual and fix themselves eventually as the app lives on.

These retentions eventually resolve themselves, so they're not true leaks, but they can show up as leaks in memory leak tests that look for disposed-but-retained objects (via Disposable's LeakFlag mechanism).

There are cases where it's very difficult to avoid these kinds of React retention paths. However, this case happens to be easier to avoid, and doing so would reduce noise and some false positives in these memory leak tests, so we want to update SafeRenderManager to not be retained by the tree it renders.

This won't prevent React from retaining the React tree it renders, which may retain other Disposable instances and still be flagged as leaks; only the SafeRenderManager object itself will no longer be retained. However, it's possible that the way content is "safely" unmounted (by rerendering SafeRenderManagerHelper with empty children before unmounting) helps prevent consumer-provided content being retained, but we're not sure at this time. 

## Changes
- Prevent SafeRenderManager from being retained by the React tree it renders. The retainers were two callback refs that directly retained SafeRenderManager due to being tearoffs:
    - `_contentCallbackRef` - fixed by using a ref object instead
    - `_helperRef` - fixed by instead retaining a `_selfRef` ref object which holds a SafeRenderManager, and gets cleared out on disposal
- Add additional test coverage to ensure callback chaining works as intended with different component types

Note that callback refs aren't inherently problematic, nor was how SafeRenderManager was being retained by the tree it rendered. The root issue here is React retaining things longer than we'd like/expect it to, and this issue helps work around that to reduce noise in tests. 

#### Release Notes
- Update `SafeRenderManager` to not be retained by the React tree it renders, to aid in memory leak testing. Outside of that context, this change has negligible effects.

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/util/safe_render_manager/safe_render_manager.dart
+++ b/lib/src/util/safe_render_manager/safe_render_manager.dart
@@ -52,11 +52,13 @@ class SafeRenderManager extends Disposable {
   /// If not specified, a new div will be used.
   final Element mountNode;
 
+  final _contentRefObject = createRef<dynamic>();
+
   /// The ref to the component rendered by [render].
   ///
   /// Due to react_dom.render calls not being guaranteed to be synchronous.
   /// this may not be populated until later than expected.
-  dynamic contentRef;
+  dynamic get contentRef => _contentRefObject.current;
 
   _RenderState _state = _RenderState.unmounted;
 
@@ -105,7 +107,7 @@ class SafeRenderManager extends Disposable {
           content = null;
           return value;
         }
-        ..contentRef = _contentCallbackRef
+        ..contentRef = _contentRefObject
       )(), mountNode);
     } catch (_) {
       _state = _RenderState.unmounted;
@@ -201,10 +203,6 @@ class SafeRenderManager extends Disposable {
       _renderQueue.forEach(_helper.renderContent);
       _renderQueue = [];
     }
-  }
-
-  void _contentCallbackRef(ref) {
-    contentRef = ref;
   }
 
   @override
--- a/lib/src/util/safe_render_manager/safe_render_manager.dart
+++ b/lib/src/util/safe_render_manager/safe_render_manager.dart
@@ -66,8 +66,21 @@ class SafeRenderManager extends Disposable {
   /// of rendering.
   List<ReactElement> _renderQueue = [];
 
-  SafeRenderManager({Element mountNode, this.autoAttachMountNode = false})
-      : mountNode = mountNode ?? DivElement();
+  /// A ref to `this`, allowing us to create closures with references to this object that can be "revoked",
+  /// so that any trees retained before React clears out old children https://github.com/facebook/react/issues/16087
+  /// don't retain a reference to this object.
+  ///
+  /// In those cases, usually it's only SafeRenderManagerHelper/SafeRenderManager that get retained,
+  /// potentially since SafeRenderManagerHelper's tryUnmountContent plus the actual unmount are enough for React to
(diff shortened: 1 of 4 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/util/safe_render_manager/safe_render_manager_helper.dart
```diff
@@ -57,24 +54,6 @@ class SafeRenderManagerHelperComponent extends UiStatefulComponent2<SafeRenderMa
...
   }
-
-  void _contentRef(ref) {
-    props.contentRef?.call(ref);
-  }
-}
-
-// AF-3369 This will be removed once the transition to Dart 2 is complete.
-// ignore: mixin_of_non_class, undefined_class
-class SafeRenderManagerHelperProps extends _$SafeRenderManagerHelperProps with _$SafeRenderManagerHelperPropsAccessorsMixin {
```
greglittlefield-wf, line 70: I upgraded the boilerplate while I was in here, but this should be fine this component is completely internal

danielway-wk (approved):
I've confirmed this resolves the specific leak I was investigating for spreadsheets, and it appears that several other tests were also impacted by this issue and are resolved by this fix (still waiting on a final test run).

lib/src/util/safe_render_manager/safe_render_manager.dart
```diff
@@ -52,20 +52,35 @@ class SafeRenderManager extends Disposable {
...
   final Element mountNode;
 
+  final _contentRefObject = createRef<dynamic>();
+
   /// The ref to the component rendered by [render].
   ///
   /// Due to react_dom.render calls not being guaranteed to be synchronous.
   /// this may not be populated until later than expected.
-  dynamic contentRef;
+  dynamic get contentRef => _contentRefObject.current;
```
greglittlefield-wf, line 61: This is technically a breaking change since it removes a setter, but setting `contentRef` shouldn't have ever been supported, and I can't find any code that sets this.

greglittlefield-wf: Actually, I decided to avoid a breakage here ee8ab08a8c48308b9218605ce62c1b15544b66c3

keatoncarter-wf (approved):

gregthompson-wk (approved):

jediaelfraser-wk (approved):

danielway-wk (approved):

aaronlademann-wf (approved):
+1 / QA +1 (CI passing)

danielway-wk:
FYI Luke is QAing in DPC before we merge.

lukeanderson-wk:
> FYI Luke is QAing in DPC before we merge.

Yep waiting on https://ci.webfilings.com/build/3406321

lukeanderson-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #752, opened 2022-05-18
[Analyzer Plugin] Add non-defaulted prop diagnostic tests

sydneyjodon-wk:
Add non-defaulted prop diagnostic tests


aaronlademann-wf (approved):
+1

PR #756, opened 2022-05-20
[Analyzer Plugin] Diagnostic prefer useRef over createRef in function components

sydneyjodon-wk:
Diagnostic prefer useRef over createRef in function components

### DIFF
--- a/tools/analyzer_plugin/lib/src/diagnostic/create_ref_usage.dart
+++ b/tools/analyzer_plugin/lib/src/diagnostic/create_ref_usage.dart
@@ -1,0 +1,96 @@
+import 'package:analyzer/dart/ast/ast.dart';
+import 'package:analyzer/dart/ast/visitor.dart';
+import 'package:analyzer_plugin/protocol/protocol_common.dart';
+import 'package:over_react_analyzer_plugin/src/diagnostic_contributor.dart';
+import 'package:over_react_analyzer_plugin/src/util/function_components.dart';
+import 'package:over_react_analyzer_plugin/src/util/react_types.dart';
+
+const _desc = r'Prefer useRef over createRef within a function component.';
+// <editor-fold desc="Documentation Details">
+const _details = r'''
+
+This diagnostic detects when `createRef`, instead of `useRef`, is used inside a function component.
+
+**BAD:** 
+```
+import 'package:over_react/over_react.dart';
+
+part 'create_ref_usage.over_react.g.dart';
+
+UiFactory<FooProps> Foo = uiFunction(
+  (props) {
+    final ref = createRef();
+
+    return (Dom.div()..ref = ref)();
+  },
+  _$FooConfig, // ignore: undefined_identifier
+);
+
+mixin FooProps on UiProps {}
+```
+
+**GOOD:**
+```
+import 'package:over_react/over_react.dart';
+
+part 'create_ref_usage.over_react.g.dart';
+
+UiFactory<FooProps> Foo = uiFunction(
+  (props) {
+    final ref = useRef();
+
+    return (Dom.div()..ref = ref)();
+  },
+  _$FooConfig, // ignore: undefined_identifier
+);
+
+mixin FooProps on UiProps {}
+```
+
+''';
+// </editor-fold>
+
+class CreateRefUsageDiagnostic extends DiagnosticContributor {
+  @DocsMeta(_desc, details: _details)
+  static const code = DiagnosticCode(
+    'over_react_create_ref_usage',
+    "Prefer useRef over createRef within a function component.",
+    AnalysisErrorSeverity.WARNING,
+    AnalysisErrorType.STATIC_WARNING,
+  );
+
+  static final fixKind = FixKind(code.name, 400, "Replace with useRef");
+
+  @override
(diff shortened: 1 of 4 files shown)


aviary-wf:
## Security Insights

## Action Items
---

tools/analyzer_plugin/lib/src/diagnostic/create_ref_usage.dart
```diff
@@ -0,0 +1,96 @@
...
+            builder.addSimpleReplacement(range.node(createRefInvocation), 'useRef');
+          }),
+        );
+      }
+    }));
+  }
+}
+
+class CreateRefUsageVisitor extends GeneralizingAstVisitor<void> {
+  final createRefInvocations = <SimpleIdentifier, Uri>{};
```
timothysteward-wk, line 84: Is there a reason to store the `Uri`? I don't see it used anywhere.

Would a `Set` be sufficent?

timothysteward-wk (approved):

tools/analyzer_plugin/test/integration/diagnostics/create_ref_usage_test.dart
```diff
@@ -0,0 +1,143 @@
...
+  }
+
+  /// Helper function to expect that the `createRef` usage error exists and verify that
+  /// the fix will replace it with `useRef`.
+  Future<void> _expectErrorAndFix(String input, String expectedOutput) async {
+    var source = newSource('test.dart', input);
+    final selection = createSelection(source, "#createRef#");
+
+    // Verify error.
+    expect((await getAllErrors(source)).single, isAnErrorUnderTest(locatedAt: selection, hasFix: true));
```
greglittlefield-wf, line 73: #nit I think you can use `expectSingleErrorAt` here instead:
```suggestion
    await expectSingleErrorAt(selection);
```

tools/analyzer_plugin/test/integration/diagnostics/create_ref_usage_test.dart
```diff
@@ -0,0 +1,143 @@
...
+    final ref = $createRefUsage<InputElement>();
+    
+    return (Dom.div()..ref = ref)();
+  },
+  _\$FooConfig, // ignore: undefined_identifier
+);
+''';
+    await _expectErrorAndFix(content('createRef'), content('useRef'));
+  }
+}
```
greglittlefield-wf, line 143: Can we add test cases for when this diagnostic shouldn't apply? E.g., `createRef` inside a class component or in another context (e.g., a top-level function or a test or something).

tools/analyzer_plugin/lib/src/diagnostic/create_ref_usage.dart
```diff
@@ -0,0 +1,96 @@
...
+    "Prefer useRef over createRef within a function component.",
+    AnalysisErrorSeverity.WARNING,
+    AnalysisErrorType.STATIC_WARNING,
+  );
+
+  static final fixKind = FixKind(code.name, 400, "Replace with useRef");
+
+  @override
+  computeErrors(result, collector) async {
+    result.unit!.accept(FunctionComponentVisitor((component) async {
```
greglittlefield-wf, line 66: I just noticed that this visitor body is `async`, but the result of those futures isn't awaited. Can we change this to `getAllFunctionComponents`, so we can iterate them in a for-loop or something where we can `await` calls properly?

greglittlefield-wf: Oh I just thought, we also should warn when people use `createRef` in custom hook functions. I'm down to do that in a follow up PR, though. It should just be iterating over `getAllHookUsages` as well.

greglittlefield-wf (changes requested):

tools/analyzer_plugin/lib/src/diagnostic/create_ref_usage.dart
sydneyjodon-wk: Oh good catch! I was previously doing something with the Uri, but now that's unnecessary - I'll remove that!

timothysteward-wk: I had another thought. Maybe we shouldn't be using `SimpleIdentifier` as a key since it doesn't implement a custom `hashCode`.

tools/analyzer_plugin/test/integration/diagnostics/create_ref_usage_test.dart
```diff
@@ -23,6 +23,43 @@ class CreateRefUsageDiagnosticTest extends DiagnosticTestBase {
...
+
+mixin FooProps on UiProps {}
+
+class FooComponent extends UiComponent2<FooProps> {
+  @override
+  get defaultProps => (newProps());
+
+  @override
+  render() {
+    final ref = createRef();
```
greglittlefield-wf, line 55: #nit This will result in an error after we add a diagnostic for not calling `createRef` inside a class method (since refs should be stored as fields).

But, we can address that when we get to it.

greglittlefield-wf (approved):

greglittlefield-wf:
QA +1

PR #769, opened 2022-07-13
FED-178 Add Documentation for `uiJsComponent`

joebingham-wk:
## Motivation
  
The process for wrapping JS components in Dart is opaque despite us having the APIs available for it to be possible. We need a document that describes the process to make it feasible for consumers to adopt the pattern.

## Changes
  
- Add a doc that describes wrapping JS components

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] Formatting is all correct
      - [ ] Links work as expected

## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

.gitignore
```diff
@@ -25,6 +25,7 @@ nbproject
 *.sublime-project
 *.sublime-workspace
 .idea/*
+.vscode/*
```
greglittlefield-wf, line 28: I'm totally fine adding this here, but FYI, if you add this to [a global .gitignore](http://egorsmirnov.me/2015/05/04/global-gitignore-file.html), you don't have to do it for each repo you work in.

doc/wrapping_js_components.md
```diff
@@ -0,0 +1,693 @@
...
+  @Deprecated('Use onClick_ArbitraryComponent for proper typing')
+  @override
+  set onClick(value);
+}
+```
+
+The steps are:
+
+1. Add overridden getters and setters for the prop. Overriding is necessary because `UiProps` already contributes the property.
+1. Deprecate those getters and setters. This isn't necessary to avoid runtime errors, but it's an important step because those properties should not be accessed by devs.
```
greglittlefield-wf, line 567:
```suggestion
> This isn't necessary to avoid runtime errors
```
I think the opposite is true. If consumers set callbacks that don't match the signature the JS code calls them with, you'll get a runtime error.

doc/wrapping_js_components.md
```diff
@@ -0,0 +1,693 @@
...
+
+```dart
+@JS('MagicalJsPackage.ArbitraryComponent')
+```
+
+> Reminder: The JavaScript bundle for the JS must be loaded into the application, [potentially similar to how React is added][react-script], or else this interop step will always fail at runtime.
+
+To check that the name is correct, you can use the browser devtools to attempt to access the property on the `Window`. If the property is valid and generally where you expect it to be then it's likely correct.
+
+<img src='./images/ui_js_component/interop_ref.gif' alt='Accessing the property in the devtools' />
```
greglittlefield-wf, line 238: Don't forget to optimize these images and rewrite the history.

doc/wrapping_js_components.md
```diff
@@ -0,0 +1,693 @@
...
+
+#### Creating the Getter
+
+Creating the getter is very simple.
+
+```dart
+external ReactClass get _jsArbitraryComponent;
+```
+
+The code here can basically be copied and pasted from the code found in this doc, just changing the name of the getter to be something more specific than `_jsArbitraryComponent`. However, the name can really be anything. We have the norm of using `_js{ComponentName}` because:
```
greglittlefield-wf, line 213: #nit In the context of stylistic programming guidelines, "convention" would be more standard way to refer to this than "norm".
```suggestion
The code here can basically be copied and pasted from the code found in this doc, just changing the name of the getter to be something more specific than `_jsArbitraryComponent`. However, the name can really be anything. We have the convention of using `_js{ComponentName}` because:
```

greglittlefield-wf (changes requested):
Final pass, this is looking super close! Awesome stuff

.gitignore
joebingham-wk: TIL, that's a good point! 👍

joebingham-wk:
Applied a hold because I'm going to pass this along to at least one consumer to make sure it all makes sense for them!

doc/wrapping_js_components.md
```diff
@@ -0,0 +1,699 @@
...
+part 'arbitrary_component.over_react.g.dart';
+```
+
+After that, move on to converting the TS props interface to a Dart mixin. Say you are given a props interface (or type) that looks like:
+
+```ts
+interface ArbitraryComponentProps {
+  anObjProp: Record<string, bool>;
+  content: string;
+  onClick: (e: React.MouseEvent, id: string) => void;
```
aaronlademann-wf, line 286:
```suggestion
  onClick: (e: React.SyntheticMouseEvent, id: string) => void;
```

aaronlademann-wf: Also - this "intricate" conversion seems like something that may be unique to MUI JS... Should we mention that if the JS component just uses the HTML `onClick` function signature - that the namespacing is not necessary?

aaronlademann-wf: Or maybe include a dom callback like `onBlur` as well - that doesn't require a custom conversion and be really explicit that the `onClick` used in the example is for demonstration purposes only - and that most JS component libs don't customize dom prop callbacks like this?

aaronlademann-wf (changes requested):
This is an incredible level of detail @joebingham-wk ... nice work!

Just have a couple of suggestions.

doc/wrapping_js_components.md
joebingham-wk: In TS, is it `React.SyntheticMouseEvent`? I only saw [types like these](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/5a03f363621433711005972e084f8e2111662299/types/react/v17/index.d.ts#L1156-L1160), where the name is just something like `ClipboardEvent` and it extends from `SyntheticEvent` (or `UiEvent`, which extends from `SyntheticEvent`).

I agree though that we should specific this is a rare case. I'll update that!

aaronlademann-wf: mmm.... you might be right that its not a synthetic event in dom callback cases in typescript land.

joebingham-wk: Ohh I may be misunderstanding, but I don't think that's quite what I was saying. I meant that in TS, synthetic events don't have the word "synthetic" in them. A `React.MouseEvent` will map to a Dart `SyntheticMouseEvent`

aaronlademann-wf (approved):
+1

doc/wrapping_js_components.md
```diff
@@ -0,0 +1,702 @@
...
+```dart
+@JS('MagicalJsPackage.ArbitraryComponent')
+external ReactClass get _jsArbitraryComponent;
+```
+
+That getter can then be used whenever it's necessary to reference the raw JS component, such as when instantiating `ReactJsComponentFactoryProxy`.
+
+In order for that getter to be valid though, it must be pointing to tangible JavaScript somewhere. The interop APIs look for that JavaScript as a property of the `window`. Your JavaScript component can be included in several ways, but it must happen before the `UiFactory` attempts to reference the interop getter. Otherwise, a runtime error will fire.
+
+One way to load the JavaScript is to use a bundle like [React usually is][react-dart]. If you are wrapping a JS library without any customizations, it's likely the bundle already exists as a [UMD module](https://jameshfisher.com/2020/10/04/what-are-umd-modules/) or similar format. If you need to write your own JS or make customizations, it's necessary to create and load a custom build. If you have any questions around this, let us know in [Slack][slack]!
```
greglittlefield-wf, line 154: > One way to load the JavaScript is to use a bundle like [React usually is][react-dart].

I think there's a wording issue here. Maybe something like this?

> One way to load the JavaScript is to use a bundle, like how [React is loaded in Dart][react-dadt].

doc/wrapping_js_components.md
```diff
@@ -0,0 +1,702 @@
...
+
+1. Create a `JsMap` property that will represent the raw JS data for the component. This is necessary because the raw JS is not in a consumable form and should not be accessed directly.
+1. Add an `@Accessor` annotation to the property with a key that matches the prop name. This will link the `JsMap` property to the JS prop.
+1. Add `Map` getters and setters named the same as the prop.
+   - The getter should use `unjsifyMapProp` to convert the JS property to a Dart `Map`
+   - The setter should use `jsifyMapProp` to convert the Dart `Map` to JS.
+
+###### Converting Refs
+
+If you read the "Index Signature" section of [Converting Maps](#converting-maps), this section will be very similar (just with different interop utilities). Say you have props that look like:
```
greglittlefield-wf, line 505: This link is broken/outdated; I think it should be:
```suggestion
If you read the "Index Signature" section of [Converting Javascript Object Types](#converting-javascript-object-types), this section will be very similar (just with different interop utilities). Say you have props that look like:
```

greglittlefield-wf (commented):
Just spotted two more tiny things! Other than those and the image optimizations, this looks good to go!

greglittlefield-wf (approved):
- Documentation looks awesome 🎉 
- No broken links

+10

joebingham-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #795, opened 2022-12-06
Add `jsifyMapListProp` and `unjsifyMapListProp` conversion utils

sydneyjodon-wk:
## Motivation
  

We need prop conversion utils to convert and unconvert `List<JsMap>` using `(un)jsifyMapProp` utils for [DataGridPro generation](https://github.com/Workiva/react_material_ui/pull/376#discussion_r1039911301)

## Changes
  

- Add `jsifyMapListProp` and `unjsifyMapListProp` conversion utils
- Add tests 

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] Verify good test coverage
      - [ ] CI passes

## Merge Checklist

### DIFF
--- a/lib/src/util/prop_conversion.dart
+++ b/lib/src/util/prop_conversion.dart
@@ -51,6 +51,16 @@ Map<K, V> unjsifyMapProp<K, V>(JsMap value) {
   return JsBackedMap.backedBy(value).cast();
 }
 
+/// Runs [jsifyMapProp] on every Dart Map in the [value].
+List<JsMap> jsifyMapListProp(List<Map> value) {
+  return value?.map(jsifyMapProp)?.toList();
+}
+
+/// Runs [unjsifyMapProp] on every JS Map in the [value].
+List<Map<K, V>> unjsifyMapListProp<K, V>(List<JsMap> value) {
+  return value?.map(unjsifyMapProp)?.toList()?.cast();
+}
+
 /// Returns [value] converted to its JS ref representation for storage in a props map, or null of the [value] is null.
 ///
 /// For use in JS component prop getters where the component expects a JS ref, but accepting Dart refs
--- a/test/over_react/util/prop_conversion_test.dart
+++ b/test/over_react/util/prop_conversion_test.dart
@@ -28,15 +28,13 @@ main() {
 
   group('prop conversion', () {
     group('utilities:', () {
+      Matcher hasJsBackedMapValue(dynamic matcher) => isA<JsMap>()
+          .having((jsMap) => JsBackedMap.backedBy(jsMap), 'JsBackedMap.backedBy', matcher);
+
       group('jsifyMapProp', () {
         test('passes through null values', () {
           expect(jsifyMapProp(null), null);
         });
-
-        Matcher hasJsBackedMapValue(dynamic matcher) => isA<JsMap>().having(
-            (jsMap) => JsBackedMap.backedBy(jsMap),
-            'JsBackedMap.backedBy',
-            matcher);
 
         test('converts maps to JS objects', () {
           expect(jsifyMapProp({'foo': 'bar'}),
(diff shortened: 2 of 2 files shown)


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/util/prop_conversion.dart
```diff
@@ -51,6 +51,16 @@ Map<K, V> unjsifyMapProp<K, V>(JsMap value) {
...
 }
 
+/// Runs [jsifyMapProp] on every Dart Map in the [value].
+List<JsMap> jsifyMapListProp(List<Map> value) {
+  return value?.map(jsifyMapProp)?.toList();
+}
+
+/// Runs [unjsifyMapProp] on every JS Map in the [value].
+List<Map<K, V>> unjsifyMapListProp<K, V>(List<JsMap> value) {
+  return value?.map(unjsifyMapProp)?.toList()?.cast();
```
greglittlefield-wf, line 61: I think the `cast` here might not be necessary
```suggestion
  return value?.map(unjsifyMapProp)?.toList();
```

test/over_react/util/prop_conversion_test.dart
```diff
@@ -75,16 +73,103 @@ main() {
...
+              ]),
+              equals([
+                {'foo': 'bar'},
+                {'bar': 1},
+                {'foo2': true},
+              ]),
+            );
+          });
+
+          test('casting the returned map to the correct type', () {
```
greglittlefield-wf, line 152: Since we're verifying the list itself here
```suggestion
          test('casting the returned list to the correct type', () {
```

test/over_react/util/prop_conversion_test.dart
```diff
@@ -75,16 +73,103 @@ main() {
...
+              {'bar': 1},
+              {'foo2': true},
+            ]),
+            orderedEquals([
+              hasJsBackedMapValue({'foo': 'bar'}),
+              hasJsBackedMapValue({'bar': 1}),
+              hasJsBackedMapValue({'foo2': true}),
+            ]),
+          );
+        });
```
greglittlefield-wf, line 107: Could we add a test around how `null` entries in the list are handled?

greglittlefield-wf (changes requested):
Couple small things, plus the additional test coverage from that other thread, otherwise this looks good!

lib/src/util/prop_conversion.dart
sydneyjodon-wk: I'm getting this error without the `cast`:
```
A value of type 'List<Map<dynamic, dynamic>>' can't be returned from the function 'unjsifyMapListProp' because it has a return type of 'List<Map<K, V>>'.
```

sydneyjodon-wk: Oh wait I guess it works if I take out that `<K, V>` - is that what you were thinking?

test/over_react/util/prop_conversion_test.dart
```diff
@@ -75,16 +73,111 @@ main() {
...
+              ]),
+            );
+          });
+
+          test('casting the returned list to the correct type', () {
+            expect(
+                unjsifyMapListProp<String, String>([
+                  jsify({'foo': 'bar'}) as JsMap
+                ]),
+                isA<List<Map<String, String>>>());
```
greglittlefield-wf, line 165: Related to the comment about `.cast`, could we also add checks here that the elements are the correct type, and not just the container list? Normally it shouldn't be possible for the wrong types to slip through, but it may be possible depending on where type checks are being compiled.
```suggestion
                allOf(
                  isA<List<Map<String, String>>>(),
                  everyElement(isA<Map<String, String>>()),
                ));
```

lib/src/util/prop_conversion.dart
greglittlefield-wf: Oh I was thinking that the generics on `.map` would get inferred and passed to `unjsifyMapProp`, but it looks like they don't because of the `toList` call, so `dynamic` is getting used instead.

So I think the current implementation:
```dart
return value?.map((e) => unjsifyMapProp(e as JsMap))?.toList()?.cast();
```
is equivalent to:
```dart
return value?.map((e) => unjsifyMapProp<dynamic, dynamic>(e as JsMap))?.toList()?.cast<Map<K, V>>();
```

And what we really want is to pass the generics to `unjsifyMapProp` so the returned maps have the right type, which should also eliminate the need for `.cast()`:
```dart
return value?.map((e) => unjsifyMapProp<K, V>(e as JsMap))?.toList();
```

sydneyjodon-wk: Oh makes sense! Updated that!

test/over_react/util/prop_conversion_test.dart
```diff
@@ -512,6 +608,232 @@ main() {
...
+                // For this test, it's important to mutate the the value returned from the `builder.listOfProps` getter,
+                // and not the original Map passed into `builder.listOfProps`.
+                builder.listOfProps[0]['data-foo'] = 'bar';
+                expect(builder.listOfProps, [
+                  {'data-foo': 'bar'}
+                ]);
+              });
+
+              group('and then reading a nested value that gets unconverted:', () {
+                group('A nested map prop', () {
```
greglittlefield-wf, line 688: #nit
```suggestion
                group('A nested map list prop', () {
```

greglittlefield-wf (approved):
+10 pending passing CI

sydneyjodon-wk:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #855, opened 2023-11-13
FED-1889 Null Safety: Audit public utility components

aaronlademann-wf:
- Prop nullability, upgrade props/state to late for public components, update class component prop defaulting
    - ErrorBoundary
    - ResizeSensor
    - AbstractTransition
    - WithTransition
    - FluxUiComponent
    - ConnectFluxPropsMixin
    - ReduxProviderPropsMixin
    - ReduxMultiProvider (storesByContext)

### DIFF
--- a/lib/src/component_declaration/flux_component.dart
+++ b/lib/src/component_declaration/flux_component.dart
@@ -50,13 +50,17 @@ mixin FluxUiPropsMixin<ActionsT, StoresT> on UiProps implements FluxUiProps<Acti
   /// The prop defined by [ActionsT] that holds all [Action]s that
   /// this component needs access to.
   ///
+  /// ** Required **
+  ///
   /// There is no strict rule on the [ActionsT] type. Depending on application
   /// structure, there may be [Action]s available directly on this object, or
   /// this object may represent a hierarchy of actions.
   @override
-  ActionsT? actions;
+  late ActionsT actions;
 
   /// The flux [Store] instance(s) to be used by a [FluxUiComponent2] instance, or a reference to one.
+  ///
+  /// ** Required **
   ///
   /// __Instead of storing state within this component via `setState`, it is recommended that data be
   /// pulled directly from these stores.__ This ensures that the data being used is always up to date
@@ -71,7 +75,7 @@ mixin FluxUiPropsMixin<ActionsT, StoresT> on UiProps implements FluxUiProps<Acti
   /// Then, you can explicitly select the [Store] instances that should be
   /// listened to by overriding [_FluxComponentMixin.redrawOn].
   @override
-  StoresT? store;
+  late StoresT store;
 
   @override
   String get _actionsPropKey {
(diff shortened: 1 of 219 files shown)


aviary-wf:
## Security Insights

## Action Items
---

lib/src/component_declaration/flux_component.dart
```diff
@@ -50,14 +50,18 @@ mixin FluxUiPropsMixin<ActionsT, StoresT> on UiProps implements FluxUiProps<Acti
...
   /// this component needs access to.
   ///
+  /// ** Required **
+  ///
   /// There is no strict rule on the [ActionsT] type. Depending on application
   /// structure, there may be [Action]s available directly on this object, or
   /// this object may represent a hierarchy of actions.
   @override
-  ActionsT? actions;
+  late ActionsT actions;
```
greglittlefield-wf, line 59: Hmmm thinking it a little more, making `actions` required might be a little tricky.

While it'll be present in most cases, there are usages that only wire up a component to redraw on the store, because the actions aren't used. Internally, it looks like ~5% of flux component usages had a store with no actions.

So our options are either:
- Make it optional
- Make it required, and plan on migrating consumers to also set `..actions =` (either to an actions class that will be unused, or to set `..actions = null` and potentially update the `TActions` generic of the consuming component to be `with FluxUiPropsMixin<MyStore, Null>`)

5% isn't a lot, so I'm leaning toward making it required, but it'll be a little bit of a pain to update those usages.

Thoughts?

lib/src/over_react_redux/redux_multi_provider.dart
```diff
@@ -68,7 +67,7 @@ class ReduxMultiProviderComponent
   get propTypes => {
         keyForProp((p) => p.storesByContext): (props, info) {
           final storesByContext = props.storesByContext;
-          if (storesByContext != null && storesByContext.isEmpty) {
```
greglittlefield-wf, line 71: Looks like there's a failing test:
```
❌ [Chrome, Dart2Js] test/over_react_redux_test.dart: ReduxMultiProvider works as expected when storesByContext is null (failed)
  Expected: emits the propType warning some element contains 'RequiredPropError: Prop ReduxMultiProviderProps.storesByContext is required.'
    Actual: <Closure: () => TestJacket<Component> from: () => {
                        let t9;
                        return jacket$.mount(react.Component, redux_multi_provider.ReduxMultiProvider().call((t9 = dom_components.Dom.div(), (() => {
                          t9.addTestId("content");
                          return t9;
                        })()).call("foo")));
                      }>
     Which: has propType warning with value [
              'Warning: Failed prop type: Expected a value of type \'Map<Context<dynamic>, Store<dynamic>>\', but got one of type \'Null\''
            ]
```

That test should probably just be removed, since when we add required prop validation it'll fail for that case when invoking the builder.

greglittlefield-wf (changes requested):
Just two small comments, one of em a question. This looks great!

lib/src/component_declaration/flux_component.dart
aaronlademann-wf: I like making it required.

lib/src/over_react_redux/redux_multi_provider.dart
aaronlademann-wf: Removed the test in https://github.com/Workiva/over_react/pull/855/commits/136eb834d6cc04b101a76857716aa18ac9e04702

greglittlefield-wf (approved):
Changes look good, +10 assuming CI passes

PR #862, opened 2023-12-01
FED-1886 Null-safe props: runtime validation of required props

sydneyjodon-wk:
## Motivation
  

FED-1717 completes support for null-safe props, and this ticket will implement runtime validation that all required late props are specified.

## Changes
  
- Implement runtime validation in UiProps.build that all non-nullable props are specified
    - Use the "more static way" method [documented in the prop null safety RFD](https://sandbox.wdesk.com/a/QWNjb3VudB82NzE5NTMwMjcyOTQ4MjI0/doc/018477c7c9c64667a25db6a867305f2e/r/-1/v/1/sec/018477c7c9c64667a25db6a867305f2e_55)
        ```dart
        mixin FooProps on UiProps {
          late String requiredProp1;
          String? optionalProp;
          late String requiredProp2;
        }
        mixin $FooProps on FooProps {
          @override
          @mustCallSuper
          void validateRequiredProps() {
            super.validateRequiredProps();
            if (!props.containsKey('FooProps.requiredProp1')) throw RequiredPropError(/* ... */);
            if (!props.containsKey('FooProps.requiredProp2')) throw RequiredPropError(/* ... */);
          }
        }
        ```
    - Do not include props considered required by annotation (`@requiredProp`). It's too disruptive/risky of a changeto make this start to throw; and we'll be deprecating this prop once `late` is available.
    - Determine when not to run this. `assert` vs `inDevMode`.
- Add escape hatch for disabling this validation, e.g., to support cases where required props are cloned onto an element.
   - Perhaps something like `UiProps.disableRequiredPropValidation()`, which unsets a flag? Or maybe we should have people invoke the component differently, or use the componentFactory directly?
- Add the ability to disable required prop validation on a certain prop
```
mixin QuxProps on UiProps {
  // Never validate that this prop is set since we know it will always get cloned on
  @disableRequiredPropValidation
  late String qux;
}
```
- Add tests

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -352,6 +364,15 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
 
     output.write(staticVariablesImpl);
 
+    if (requiredPropChecks.isNotEmpty) {
+      final validateRequiredPropsMethod = '\n  @override\n'
+          '  @mustCallSuper\n'
+          '  void validateRequiredProps() {\n'
```
sydneyjodon-wk, line 370: I'm not sure if this was supposed to call `super.validateRequiredProps()`, but I get this error in generated files when I add that `error: The method 'validateRequiredProps' is always abstract in the supertype.`

greglittlefield-wf: I tried adding this in locally to look more closely at the issues, and it looks like this only affects legacy Dart-2-only boilerplate...

I tried seeing if we could add a super-constraint to those mixins:
```diff
diff --git a/lib/src/builder/codegen/accessors_generator.dart b/lib/src/builder/codegen/accessors_generator.dart
index 1003b2ed..f19daee3 100644
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -82,9 +82,7 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
       final implementsClause = 'implements ${names.consumerName}$typeParamsOnSuper';
       generatedClass
         ..write(generatedCodeUseOnlyDeprecation)
-        ..writeln('abstract class $accessorsMixinName$typeParamsOnClass $implementsClause {\n'
-            '  @override'
-            '  Map get ${type.isProps ? 'props' : 'state'};\n');
+        ..writeln('mixin $accessorsMixinName$typeParamsOnClass on ${type.isProps ? 'UiProps' : 'UiState'} $implementsClause {\n');
       if (type.isMixin) {
         generatedClass.writeln(_generateStaticMetaDecl(names.publicName, type.isProps));
         generatedClass.write(_copyStaticMembers(node));
@@ -367,6 +365,7 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
       final validateRequiredPropsMethod = '\n  @override\n'
           '  @mustCallSuper\n'
           '  void validateRequiredProps() {\n'
+          '    super.validateRequiredProps();\n'
           '    ${requiredPropChecks.join('\n')}\n'
           '  }\n';
       output.write(validateRequiredPropsMethod);
```
...but that caused a whole bunch of other typing issues that don't seem feasible to fix without a lot of effort and potential for regressions.

So I'm thinking we just don't support required prop validation for that boilerplate version; that boilerplate is really only used within over_react, so I don't think that should be a huge issue.

With these changes, it seems to work!
```diff
diff --git a/lib/src/builder/codegen/accessors_generator.dart b/lib/src/builder/codegen/accessors_generator.dart
index 1003b2ed..c388384a 100644
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -363,10 +363,11 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
 
     output.write(staticVariablesImpl);
 
-    if (requiredPropChecks.isNotEmpty) {
+    if (type.isProps && version != Version.v3_legacyDart2Only) {
       final validateRequiredPropsMethod = '\n  @override\n'
           '  @mustCallSuper\n'
           '  void validateRequiredProps() {\n'
+          '    super.validateRequiredProps();\n'
           '    ${requiredPropChecks.join('\n')}\n'
           '  }\n';
       output.write(validateRequiredPropsMethod);
```

Also, I'm thinking it'd probably be safer to not conditionally output the override based on `requiredPropChecks.isNotEmpty`. If we always output it, then it'll help our tests catch any cases where the generated code is valid when there are no required props, but might become invalid if required props are ever added. And I'd expect those empty overrides, where it just calls super, to get compiled out in dart2js.

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -149,6 +150,8 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
 
     StringBuffer output = StringBuffer();
 
+    final requiredPropChecks = [];
```
greglittlefield-wf, line 153: This is missing a type:
```suggestion
    final requiredPropChecks = <String>[];
```

greglittlefield-wf (changes requested):
Made a partial pass and got through most of the lib changes; one area of feedback but otherwise this looks great! Will make another pass at the tests in a bit

lib/src/builder/codegen/accessors_generator.dart
sydneyjodon-wk: Oh nice! I didn't realize it was just for some boilerplates! I added that change in [8890aad](https://github.com/Workiva/over_react/pull/862/commits/8890aad7c2104f68200e310a12ac52be49c07501), but I'm still getting a few `error: The method 'validateRequiredProps' is always abstract in the supertype.` Like [this one](https://github.com/Workiva/over_react/blob/c346c0c371957eceb59fefafb0750c9017c1cb3d/test/over_react/component/fixtures/dummy_component.over_react.g.dart#L51) - what boilerplate version is that?

greglittlefield-wf: Oh that's the v2 backwards compat boilerplate. Hmm it looks like that version has the same issues... 

Dang, sorry about that, I could have sworn it all looked good locally when I was testing it out; maybe my analyzer wasn't catching up? Or maybe I had other local generation changes I forgot to comment out?

Hmm tweaking the generation of that specific case in legacy boilerplate, so that the accessors mixin is a mixin on the base class instead of implementing it, seems to fix the issue:
```diff
diff --git a/lib/src/builder/codegen/accessors_generator.dart b/lib/src/builder/codegen/accessors_generator.dart
index 3d72fb35..85be0eae 100644
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -79,16 +79,20 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
       }
     }
     if (version.isLegacy) {
+      if (type.isMixin) {
         final implementsClause = 'implements ${names.consumerName}$typeParamsOnSuper';
         generatedClass
           ..write(generatedCodeUseOnlyDeprecation)
           ..writeln('abstract class $accessorsMixinName$typeParamsOnClass $implementsClause {\n'
               '  @override'
               '  Map get ${type.isProps ? 'props' : 'state'};\n');
-      if (type.isMixin) {
         generatedClass.writeln(_generateStaticMetaDecl(names.publicName, type.isProps));
         generatedClass.write(_copyStaticMembers(node));
         generatedClass.write(_copyClassMembers(node));
+      } else {
+        generatedClass
+          ..write(generatedCodeUseOnlyDeprecation)
+          ..writeln('mixin $accessorsMixinName$typeParamsOnClass on ${names.consumerName}$typeParamsOnSuper {\n');
       }
     } else {
       generatedClass
```
Though, I think that would only fix it for props declared in props classes, and not mixins... (that `if (type.isMixin)` case was the one I had issues with [here](https://github.com/Workiva/over_react/pull/862/commits/c346c0c371957eceb59fefafb0750c9017c1cb3d#r1420812787)).

I'm starting to think we should just not support this validation for either of the legacy boilerplates :/. How does that sound to you?

sydneyjodon-wk: Not supporting validation for legacy boilerplates works for me! I think some of them already throw errors for required props anyway

test/over_react_component_declaration_test.dart
```diff
@@ -136,4 +137,7 @@ main() {
   new_boilerplate_required_accessor_integration_test.main();
   new_boilerplate_stateful_component_integration_test.main();
   new_boilerplate_unassigned_prop_integration_test.main();
+
+  // FIXME: Move these tests to only be run opted in to null safety.
```
greglittlefield-wf, line 141: Hmm is this fixme still relevant?

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/null_safety_validate_required_props_test.dart
```diff
@@ -0,0 +1,233 @@
...
+
+import '../../../../test_util/test_util.dart';
+
+part 'null_safety_validate_required_props_test.over_react.g.dart';
+
+void main() {
+  group('(New boilerplate) validates required props:', () {
+    group('non-nullable required prop', () {
+      group('throws when a prop is required and not set', () {
+        test('on mount', () {
```
greglittlefield-wf, line 27: I think we want to verify more specifically that these throw when the props are invoked, before they're even attempted to be rendered.

So the setup should be more like:
```dart
          expect(() {
            (ComponentTest()
              ..requiredNullable = true
            )();
          },
              throwsA(isA<MissingRequiredPropsError>()...
```

Though I do think that also having test coverage that they render properly, like you have now, is probably a good idea.

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/null_safety_validate_required_props_test.dart
```diff
@@ -0,0 +1,233 @@
...
+          rtl.render((MultipleMixinsTest()
+            ..requiredNullable = true
+            ..requiredNonNullable = true
+            ..secondRequiredProp = true
+          )());
+        },
+          returnsNormally);
+      });
+    });
+  }, tags: 'ddc');
```
greglittlefield-wf, line 195: Can we also verify that validation is disabled in dart2js? Doesn't have to be exhaustive will all the cases; just one test that missing required props don't cause failures should be sufficient.

lib/src/over_react_redux/over_react_redux.dart
```diff
@@ -51,7 +52,8 @@ abstract class _$ConnectPropsMixin implements UiProps {
   @override
   Map get props;
 
-  dynamic Function(dynamic action)? dispatch;
+  @disableRequiredPropValidation
```
greglittlefield-wf, line 55: I think a comment here would be helpful
```suggestion
  // Disable validation since this prop is set by the `connect` HOC, and does not need to be set by consumers.
  @disableRequiredPropValidation
```

lib/src/component_declaration/component_base.dart
```diff
@@ -621,11 +621,12 @@ abstract class UiProps extends MapBase
...
-    assert(() {
-      validateRequiredProps();
-      return true;
-    }());
+      assert(() {
+        if(_shouldValidateRequiredProps) {
+            validateRequiredProps();
+        }
+        return true;
+      }());
```
greglittlefield-wf, line 629: #nit formatting
```suggestion
    assert(() {
      if (_shouldValidateRequiredProps) {
        validateRequiredProps();
      }
      return true;
    }());
```

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -198,6 +203,12 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           annotationCount++;
           isRequired = true;
           isPotentiallyNullable = true;
+
+          if (type.isProps && disableRequiredPropValidation == null) {
+            requiredPropChecks.add('  if(!props.containsKey($keyValue)) {\n'
+                '  throw MissingRequiredPropsError(\'Required prop `$accessorName` is missing.\');\n'
```
greglittlefield-wf, line 209: This needs to be escaped in case the name of the prop has a `$` character in it. (In general it's a good idea to always escape strings by default.)

You can use `stringLiteral` for that here:
```suggestion
                '  throw MissingRequiredPropsError(${stringLiteral('Required prop `$accessorName` is missing.')});\n'
```

lib/src/component_declaration/annotations.dart
```diff
@@ -15,6 +15,8 @@
 // Dummy annotations that would be used by Pub code generator
 library over_react.component_declaration.annotations;
 
+export 'package:meta/meta.dart' show mustCallSuper;
```
greglittlefield-wf, line 18: Oh interesting, this is so that generated code can use `@mustCallSuper`, right? Could we add a comment here to that effect.

Kind of gross that it has to be exported from over_react and added to its public API, though I don't think it would cause any issues... 

And I can't think of any better alternatives, unless we created an alias for it that was named something else, and maybe deprecated (e.g., `@Deprecated('For generated use only) const $overReact$mustCallSuper = mustCallSuper;`)... Not sure if that's worth the trouble though.

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/null_safety_validate_required_props_test.dart
greglittlefield-wf, line 1: Sorry, I know you just ripped out the tests for the old boilerplate, but would you mind adding some tests that verify that late props aren't validated upon UiProps invocation for those? That way we're guarded against any regressions to that behavior in the future.

greglittlefield-wf (changes requested):
Couple more comments; sorry it took me so long to circle back to the tests!

greglittlefield-wf (approved):
+10

PR #874, opened 2024-01-23
FED-2015 Opt out of required props from mixin declaration

sydneyjodon-wk:
## Motivation
  
Consumers can currently opt props they declare out of required prop vaildation, by annotating individual props with `@disableRequiredPropValidation`, but they can't opt out of validation for props in other mixins.

This becomes an issue for certain use-cases, primarily the declaration of wrapper components.

For example:

```
mixin FooProps on UiProps {
  late String requiredProp1AlwaysSetInWrapper;
  late String requiredProp2;
}

UiFactory<FooProps> Foo = uiFunction((props) {
  // ...
}, _$FooConfig);

class WrapperProps = UiProps with FooProps, WrapperPropsMixin;

UiFactory<WrapperProps> Wrapper = uiForwardRef((props, ref) {
  return (Foo()
    ..requiredProp1AlwaysSetInWrapper = 'foo'
    ..addProps(props.getPropsToForward(exclude: {WrapperPropsMixin}))
    ..ref = ref
  )();
}, _$WrapperConfig);
```

`Wrapper` sets required props on the component it wraps, but doing `(Wrapper()..requiredProp2 = '')()` will still result in runtime errors and lint warnings because `requiredProp1AlwaysSetInWrapper` isn't set set.

This pattern is pretty common, and we need a way for consumers to be able to continue to write these wrappers without required prop validation getting in their way.
## Changes
  
- Finish spike branch and add ignore option
- Write tests
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist

### DIFF
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -205,7 +205,8 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           isPotentiallyNullable = true;
 
           if (type.isProps && disableRequiredPropValidation == null) {
-            requiredPropChecks.add('  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains($keyValue)) {\n'
+            requiredPropChecks.add(
+                '  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains(\'$accessorName\')) {\n'
                 '  throw MissingRequiredPropsError(${stringLiteral('Required prop `$accessorName` is missing.')});\n'
                 '}\n');
           }
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -206,7 +206,7 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
 
           if (type.isProps && disableRequiredPropValidation == null) {
             requiredPropChecks.add(
-                '  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains(\'$accessorName\')) {\n'
+                '  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains(${stringLiteral(accessorName)})) {\n'
                 '  throw MissingRequiredPropsError(${stringLiteral('Required prop `$accessorName` is missing.')});\n'
                 '}\n');
           }
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -205,8 +205,8 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           isPotentiallyNullable = true;
 
           if (type.isProps && disableRequiredPropValidation == null) {
-            requiredPropChecks.add('  if(!props.containsKey($keyValue)) {\n'
-                '  throw MissingRequiredPropsError(${stringLiteral('Required prop `$accessorName` is missing.')});\n'
+            requiredPropChecks.add('  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains($keyValue)) {\n'
+                '  throw MissingRequiredPropsError(requiredPropNamesToSkipValidation.join(\',\'));\n'
                 '}\n');
           }
         }
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -374,11 +374,23 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
     if (type.isProps &&
         version != Version.v3_legacyDart2Only &&
         version != Version.v2_legacyBackwardsCompat) {
+      List<String> implLines;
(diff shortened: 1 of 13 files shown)


aviary2-wf:
## Security Insights

## Action Items
---

lib/src/builder/parsing/members/props_and_state_util.dart
```diff
@@ -62,6 +64,18 @@ annotations.TypedMap getPropsOrStateAnnotation(bool isProps, AnnotatedNode node)
           // ignore: deprecated_member_use_from_same_package
           InstantiatedMeta.fromNode<annotations.StateMixin>(node));
 
+  if (meta != null && meta.potentiallyIncompleteValue is annotations.Props) {
+    // fixme error when this case isn't well formed
```
sydneyjodon-wk, line 68: @greglittlefield-wf was there a certain case you had in mind to throw an error for here?

greglittlefield-wf: Yeah I was thinking if we couldn't parse the value, then we should throw, but it looks like that's probably already handled by accessing `meta?.value`... So maybe this was just a bad FIXME

lib/src/builder/parsing/members/props_and_state_util.dart
```diff
@@ -62,6 +64,18 @@ annotations.TypedMap getPropsOrStateAnnotation(bool isProps, AnnotatedNode node)
...
+    if (meta.unsupportedArguments.length == 1) {
+      final arg = meta.unsupportedArguments[0];
+      if (arg is NamedExpression && arg.name.label.name == 'ignoreRequiredProps') {
+        props_ignoreRequiredProps_source[meta.potentiallyIncompleteValue] =
```
greglittlefield-wf, line 71: Ooh I initially used this expando solution because there wasn't a way to instantiate the `Props` with real references to the types for `ignoreRequiredPropsFrom`, and was pretty unhappy with how sketchy it was. E.g., if we ended up with a different instance of props somehow, then we'd lose that value.

Ideally I'd love to use parse the expression and pass that data into the real `Props` object so we can eliminate that expando, if it's not too much more effort.

Something like this:
```dart
      if (arg is NamedExpression && arg.name.label.name == 'ignoreRequiredProps') {
        // Attempt to parse the value, and fall through if something goes wrong,
        // and let `meta?.value` below throw.
        final expression = arg.expression;
        if (expression is SetOrMapLiteral && expression.isSet) {
          final simpleStringElements =
              expression.elements.whereType<SimpleStringLiteral>().toList();
          if (simpleStringElements.length == expression.elements.length) {
            return annotations.Props(
              keyNamespace: meta.potentiallyIncompleteValue.keyNamespace,
              ignoreRequiredProps: simpleStringElements.map((e) => e.value).toSet(),
            );
          }
        }
      }
```

I'd be fine doing this as a follow-up ticket, too

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -205,7 +205,8 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           isPotentiallyNullable = true;
 
           if (type.isProps && disableRequiredPropValidation == null) {
-            requiredPropChecks.add('  if(!props.containsKey($keyValue)) {\n'
+            requiredPropChecks.add(
+                '  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains(\'$accessorName\')) {\n'
```
greglittlefield-wf, line 209: We should should make sure this string is properly escaped
```suggestion
                '  if(!props.containsKey($keyValue) && !requiredPropNamesToSkipValidation.contains(${stringLiteral(accessorName)})) {\n'
```

lib/src/component_declaration/component_base.dart
```diff
@@ -693,6 +693,10 @@ abstract class UiProps extends MapBase
   void disableRequiredPropValidation() {
     _shouldValidateRequiredProps = false;
   }
+
+  /// Names of props to opt out of required prop validation for.
```
greglittlefield-wf, line 697: #nit Could be helpful to provide maintainers with a little more context
```suggestion
  /// Names of props to opt out of required prop validation for.
  ///
  /// Overridden in generated code, based on the value of `@Props(ignoreRequiredProps: ...)`.
```

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/null_safety_validate_required_props_test.dart
```diff
@@ -262,4 +271,17 @@ mixin MultipleMixinsTestPropsMixin on UiProps {
   late bool secondRequiredProp;
 }
 
+@Props(ignoreRequiredProps: {'testIgnoreOnMixin'})
```
greglittlefield-wf, line 274: Could we update the tests using `MultipleMixinsTest` to indicate that ignored props are also being tested for that case? E.g., in the test descriptions or in comments within the tests.

Just some way so it's clear by looking at the tests themselves that we have test coverage for this case: ignoring some required props, and not others.

greglittlefield-wf (changes requested):
Just a couple more comments

greglittlefield-wf (approved):
+10 assuming CI passes

mitchellsharma-wk (approved):
+1 lgtm

PR #888, opened 2024-02-15
FED-2242 Stop exporting meta

greglittlefield-wf:
## Motivation
I noticed when pulling v5_wip into a test project that there were a lot of infos that looked like:

> info: The import of 'package:meta/meta.dart' is unnecessary because all of the used elements are also provided by the import of 'package:over_react/over_react.dart'. (unnecessary_import at …)

This is because we export `mustCallSuper` so we can use it in generated code. 

Initially I thought this was going to be okay: https://github.com/Workiva/over_react/pull/862#discussion_r1427412583
> > ```diff
> > +export 'package:meta/meta.dart' show mustCallSuper;
> > ```
> ##### @greglittlefield-wf
> Oh interesting, this is so that generated code can use @mustCallSuper, right? Could we add a comment here to that effect.
> 
> Kind of gross that it has to be exported from over_react and added to its public API, though I don't think it would cause any issues...
> 
> And I can't think of any better alternatives, unless we created an alias for it that was named something else, and maybe deprecated (e.g., `@Deprecated('For generated use only) const $overReact$mustCallSuper = mustCallSuper;`)... Not sure if that's worth the trouble though.

These new lints are definitely enough to make me think it is worth the trouble 🙂.

## Changes
**Note: diff best reviewed commit-by-commit**

- 9306f6b3412990f0337c60397dd7acaf43dfd1f6 f236c025f89a04ab3777556c510644676a9cf59d Make ConnectPropsMixin implement `builder_helpers.UiProps` instead of `component_base.UiProps`
    - Without this, because of the way things were imported, the generated code was referencing `component_base.UiProps.$mustCallSuper`, when it needs to reference `component_base.UiProps.$mustCallSuper`
    - We should have been doing this anyway; this is technically a breaking change for any existing consumers extending only from `component_base.UiProps`, but we don't really support consumers using the component_base version in declarations, and I don't see any consumers that would be impacted. If there are any that would be impacted, it'd be a compilation error we'd catch in testing.
- 53999664806ae3657575c5ac7a5840a7ae2f2795 Stop exporting package:meta, and use a custom constant instead (I was able to use `@protected` to help prevent other code from using it)
- 1cd247b78a3ec3e43f637a993042a43421440427 Update generated code
- fca76c6c0c2f2ce4e700deed615c87af8ce2cafc Update generated gold files

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        - Everything analyzes cleanly with these changes (CI passes)

## Merge Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,10 @@
     - `ReduxMultiProviderProps.storesByContext`
     - `ReduxProviderProps.store`
 - UiPropsMapView (deprecated) is now abstract and requires subclasses to override `selfFactory`
-- `PropsMeta`/`StateMeta` constructor arguments `fields` and `keys` are now required
-- `ProviderProps.props` type has been widened from `JsMap` to `Map` (now matches `ConsumerProps` and other props classes)
+- Other changes that we don't expect to affect consumers:
+  - `PropsMeta`/`StateMeta` constructor arguments `fields` and `keys` are now required
+  - `ProviderProps.props` type has been widened from `JsMap` to `Map` (now matches `ConsumerProps` and other props classes)
+  - `ConnectPropsMixin` now `implements UiProps` (the one exported from `package:over_react/over_react.dart`) instead of `UiProps`'s superclass of the name, `component_base.UiProps`.
  
 ## [4.10.3](https://github.com/Workiva/over_react/compare/4.10.2...4.10.3)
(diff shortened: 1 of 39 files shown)


aviary-wf:
## Security Insights

## Action Items
---

robbecker-wf (approved):

PR #891, opened 2024-02-21
FED-2267 Turn off prop type warnings for ignored prop validation

sydneyjodon-wk:
## Motivation
  
Based on [this discussion](https://github.com/Workiva/wdesk_sdk/pull/12266#discussion_r1496136886) - we need to turn off prop type warnings for props that have been ignored in `@Props(disableRequiredPropValidation)`
## Changes
  
- Coniditionally turn off prop type warnings
## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] Verify that new tests fail without the fix - https://github.com/Workiva/over_react/actions/runs/7995851924/job/21837092967
      - [ ] CI passes on this PR
      - [ ] Also see this branch fixing failed tests in https://github.com/Workiva/wdesk_sdk_legacy/pull/65

## Merge Checklist

### DIFF
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -216,9 +216,11 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component with 
 
   /// Validates that props with the `@requiredProp` annotation are present.
   void validateRequiredProps(Map appliedProps) {
+    final requiredPropNamesToSkipValidation = newProps().requiredPropNamesToSkipValidation;
     consumedProps?.forEach((consumedProps) {
       consumedProps.props.forEach((prop) {
         if (!prop.isRequired) return;
+        if (requiredPropNamesToSkipValidation.contains(prop.key.split('.').last)) return;
         if (prop.isNullable && appliedProps.containsKey(prop.key)) return;
         if (!prop.isNullable && appliedProps[prop.key] != null) return;
 
--- a/lib/src/component_declaration/component_base.dart
+++ b/lib/src/component_declaration/component_base.dart
@@ -216,11 +216,11 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component with 
 
   /// Validates that props with the `@requiredProp` annotation are present.
   void validateRequiredProps(Map appliedProps) {
-    final requiredPropNamesToSkipValidation = newProps().requiredPropNamesToSkipValidation;
     consumedProps?.forEach((consumedProps) {
       consumedProps.props.forEach((prop) {
         if (!prop.isRequired) return;
-        if (requiredPropNamesToSkipValidation.contains(prop.key.split('.').last)) return;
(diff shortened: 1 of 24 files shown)


lib/src/component_declaration/component_base.dart
```diff
@@ -216,9 +216,11 @@ abstract class UiComponent<TProps extends UiProps> extends react.Component with
 
   /// Validates that props with the `@requiredProp` annotation are present.
   void validateRequiredProps(Map appliedProps) {
+    final requiredPropNamesToSkipValidation = newProps().requiredPropNamesToSkipValidation;
     consumedProps?.forEach((consumedProps) {
       consumedProps.props.forEach((prop) {
         if (!prop.isRequired) return;
+        if (requiredPropNamesToSkipValidation.contains(prop.key.split('.').last)) return;
```
greglittlefield-wf, line 223: Hmm I'm not a big fan of pulling the name out of the prop key here is a good idea, since it won't behave well for edge-cases:
- props with the same name declared in different prop mixins
- props with custom keys that don't match the name

Some other paths forward that I can think of off the top of my head, in order from most to least preference:
- Update PropDescriptor to be able to tell the difference between `@requiredProp` and `late`, and just disable this validation in the component for `late` props entirely (since they're already validated by `UiProps.call`/`.build`)
   -  Maybe this is a new field like `PropDescriptor.isLate`?
- Update PropDescriptor to include the prop name and use that in check
- Remove this old automatic required prop validation (in favor of the new validation in `UiProps.call`/`.build` and the analyzer plugin)

Thoughts?

greglittlefield-wf (changes requested):

lib/src/component_declaration/component_base.dart
greglittlefield-wf: Yeahh 😕. I'm not 100% sure how I feel about that option, but my thought was if, as part of consumers migrating to null safety, they'll be getting rid of `@reuiredProp`, then maybe it's worth it to safe us some headaches in getting this working.

But, thinking more about that... we probably have consumers relying on that behavior now in tests... My thought was that if they were only using it as a debugging tool during development, we could probably remove it without breaking anyone, but if it'll cause existing tests to fail, then it's probably not worth the added friction after all.

sydneyjodon-wk: I just updated it to add an `isLate` property to `PropDescriptor`, but I realized that turning off proptype validation for all late props means that late props in old boilerplates won't have any validation since the validation in call/build only runs in the newest boilerplates, right? Are we cool with that?

greglittlefield-wf: Yeah I'm cool with that 👍

greglittlefield-wf (approved):
+1, can QA once test branch rerun finishes

sydneyjodon-wk:
@greglittlefield-wf looks like it passed!

greglittlefield-wf:
QA +1

PR #893, opened 2024-03-01
FED-2299 Utilize new ReactNode typedef

aaronlademann-wf:
## Motivation
https://github.com/Workiva/react-dart/pull/384 adds a new typedef that is an alias for `Object?`. We should use it!

## Changes
Use it where appropriate

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

lib/src/component/prop_mixins.dart
```diff
@@ -46,10 +46,10 @@ abstract class _$ReactPropsMixin {
 
   // This private field is namespaced to avoid colliding with other classes.
   @Accessor(key: 'children')
-  dynamic _raw$ReactProps$children;
+  ReactNode _raw$ReactProps$children;
 
   /// The children that were passed in to this component when it was built.
-  List<dynamic>? get children {
+  List<ReactNode>? get children {
```
greglittlefield-wf, line 52: Unfortunately this is a breaking change, since consumers will no longer be able to dynamically invoke methods on children (which they probably shouldn't be doing, and instead using proper type checks). 

And I think it might mess up implicit casting in some cases too, not 100% sure.

I'd just be surprised if there aren't some usages out there that would be broken. 

So, I'd say either we revert this change, or if we really wanted to, perhaps we could do some consumer static analysis testing to see if it affects anything.

lib/src/util/react_wrappers.dart
```diff
@@ -195,7 +195,7 @@ Map getProps(dynamic/* ReactElement|ReactComponent */ instance, {bool traverseWr
 ///
 /// This method simply wraps react.findDOMNode with strong typing for the return value
 /// (and for the function itself, which is declared using `var` in react-dart).
-Element? findDomNode(dynamic instance) => react_dom.findDOMNode(instance) as Element?;
+Element? findDomNode(ReactNode instance) => react_dom.findDOMNode(instance) as Element?;
```
greglittlefield-wf, line 198: I don't think `ReactNode` is the appropriate type here, since it should be the mounted component instance (`ReactComponent|react.Component|Element`), not the VDOM instance. So, I think we should switch this back to `dynamic`.

lib/src/util/react_wrappers.dart
```diff
@@ -223,7 +223,7 @@ bool isDomElement(dynamic instance) {
 /// Returns whether [instance] is a composite [ReactComponent].
 ///
 /// __Not for external use.__
-bool _isCompositeComponent(Object? instance) {
+bool _isCompositeComponent(ReactNode instance) {
```
greglittlefield-wf, line 226: Same comment as for findDomNode

lib/src/component/suspense_component.dart
```diff
@@ -67,11 +67,11 @@ mixin SuspensePropsMixin on UiProps {
   /// The actual UI you intend to render. If children suspends while rendering, the Suspense boundary will
   /// switch to rendering fallback.
   @override
-  /*ReactNode*/ List<dynamic>? get children;
+  List<ReactNode>? get children;
```
greglittlefield-wf, line 70: Similar to above, the changes to props in this component are technically breaking changes, but I think we could get away with these in particular since, by comparison, there should be almost no components that read the children/fallback props from this props mixin.

greglittlefield-wf (changes requested):
Just a couple comments, otherwise looks good!

lib/src/component/prop_mixins.dart
aaronlademann-wf: Kicked off [a static analysis check](https://github.com/Workiva/gha-dart/actions/runs/10375376792)

aaronlademann-wf: The check did not do well. I'll revert this.

aaronlademann-wf: https://github.com/Workiva/over_react/pull/893/commits/908bfdf84fd0069e8c3cb591a4e59481482e5a35

greglittlefield-wf (approved):
+10

PR #898, opened 2024-03-25
FED-2443 Merge version 5.0.0 changes to master

aaronlademann-wf:
As a first step to releasing version 5.0.0 of this library, we need to: 

1. Update the changelog to include information about the 5.0.0 release.
1. Temporarily point to the unreleased [3.0.0 changes](https://github.com/Workiva/over_react_test/pull/160) in master of over_react_test.
1. Merge the changes from the `v5_wip` branch into master without actually tagging it as a release.


aviary2-wf:
## Security Insights

## Action Items
---

CHANGELOG.md
```diff
@@ -1,5 +1,50 @@
...
+- The following props are now required:
+    - `FluxUiPropsMixin.actions`, `FluxUiPropsMixin.store`
+    - `ProviderProps.value` (`ProviderProps` is the return type of `Context.Provider`)
+    - `ReduxMultiProviderProps.storesByContext`
+    - `ReduxProviderProps.store`
+- UiPropsMapView (deprecated) 
+    - is now abstract and requires subclasses to override `selfFactory`
+    - now extends directly from UiProps, and longer implements `MapView` (it still implements `Map`)
+    - `props` now returns the backing map instead of `this`
+- Other changes that we don't expect to affect consumers:
```
greglittlefield-wf, line 40: Also noticed this diff that may be worth mentioning:
```suggestion
- `react_dom.render` arg loosened from `ReactElement` to `dynamic`
- Other changes that we don't expect to affect consumers:
```

greglittlefield-wf (commented):
Couple comments on the changelog, and made it partially through the full diff. Will continue reviewing tomorrow!

lib/src/over_react_redux/over_react_flux.dart
```diff
@@ -43,7 +43,7 @@ abstract class _$ConnectFluxPropsMixin<TActions> implements UiProps {
   @override
   Map get props;
 
-  TActions actions;
+  late TActions actions;
```
greglittlefield-wf, line 46: I think this should have disabled validation, just like we do in ConnectPropsMixin:
```suggestion
  // Disable validation since this prop is set by the `connectFlux` HOC, and does not need to be set by consumers.
  @disableRequiredPropValidation
  late TActions actions;
```

I don't see any consumers of it, which is probably why didn't notice this being a potential issue.

pubspec.yaml
```diff
@@ -3,21 +3,20 @@ version: 4.11.1
...
-  collection: ^1.14.11
+  collection: ^1.15.0
   analyzer: ^5.1.0
   build: '>=1.0.0 <3.0.0'
-  built_redux: ^8.0.6
-  built_value: ^8.0.0
   dart_style: ^2.0.0
   js: ^0.6.1+1
-  logging: ^1.0.0
+  logging: '>=0.11.3+2 <2.0.0'
```
greglittlefield-wf, line 14: Looks like this was probably a bad merge from before this dep got updated in master; we need `^1.0.0` since it's the first null-safe release.
```suggestion
  logging: ^1.0.0
```

greglittlefield-wf (changes requested):
Finished doing an additional pass on all the changes in lib!

Just spotted a couple things, and also noted some things missing from the changelog that I can push up or PR

greglittlefield-wf (approved):
- Changelog looks good
- Re-reviewed all changes in lib and pubspec.yaml and they look good
- Almost all commits in this branch have been reviewed and tested in PRs into v5_wip, and other commits look good
- CI passes
- This branch has been recently tested in consumers

+10

greglittlefield-wf:
semver +1

sydneyjodon-wk (approved):
+1 on Greg's most recent commits

dancleereman-wf:
RM +1

PR #920, opened 2024-05-23
FED-1945 Support prop conversion in OverReact to simplify RMUI converted props

sydneyjodon-wk:
## Motivation
  
We have some props that perform conversion in getters and setters. 

Currently, we implement these using a private prop with the "raw" value, and then perform conversion in getters and setters.
```dart
  Map get sx => unjsifyMapProp(_sx$rawJs);
  set sx(Map value) => _sx$rawJs = jsifyMapProp(value);

  /// The raw JS value of [sx] stored in the props map.
  @Accessor(key: 'sx')
  JsMap _sx$rawJs;
```

Apart from being verbose, an issue with this approach is that analyzer tooling in over_react only sees the private props, and not the getter/setter wrappers. See FED-1864.

To work around these tooling challengges, and to clean up these props, we want to update the over_react builder to allow us to declare these converted/unconverted props more closely to how we declare other props.

As a solution, we could add an annotation that takes in a top-level converter, unconverter, and type used for the "JS" value of the prop (necessary to avoid implcit casts in generated code)

```dart
// What you write
@ConvertProp<JsMap, Map>(unjsifyMapProp, jsifyMapProp)
Map foo;

// What gets generated in the generated mixin
@override
Map get foo => unjsifyMapProp(props['foo'] as JsMap);
@override
set foo(Map value) => props['foo'] = jsifyMapProp(value);
```

We could also support aliases for common cases, like Maps and refs.
```dart
@convertJsMapProp
Map foo;

@convertJsRefProp
dynamic barRef;
```

## Changes
  

- Implement prop conversion within over_react via new `ConvertProp` annotation (or maybe the existing `@Accessors` annotation?) that takes in:
    - a converter function
    - an unconverter function
    - the type of the raw JS value
        Ideally we'd make this a generic parameter, which will help ensure the converter and unconverter are typed correctly. The only caveat is that generic annotations are only supported in language versions 2.14 and higher. Are we okay with only supporting this after consumers have opted into null safety?
- Make it a builder error if 
    - consumers don't specify generic parameters, since we need that type to be able to cast the value from the getter and not get implcit cast errors
    - (if it's a generic parameter) the non-raw type doesn't match the type annotation on the prop
- Add shorthand annotations `convertJsMapProp` and `convertJsRefProp`
- Add tests

## Review

  

Please review: 

### QA Checklist
    - [ ] Steps from PR author: 
        
      - [ ] CI passes
      - [ ] RMUI PR incoming - https://github.com/Workiva/react_material_ui/pull/1072

## Merge Checklist


aviary3-wk:
## Security Insights

## Action Items
---

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart
```diff
@@ -0,0 +1,83 @@
...
+// limitations under the License.
+
+import 'package:over_react/over_react.dart';
+import 'package:react_testing_library/react_testing_library.dart' as rtl;
+import 'package:test/test.dart';
+
+part 'prop_conversion_annotations_test.over_react.g.dart';
+
+void main() {
+  group('(New boilerplate) prop conversion annotations:', () {
```
sydneyjodon-wk, line 23: Not sure if this is the best way to test this functionality 🤷‍♀️

test/vm_tests/builder/codegen_test.dart
```diff
@@ -1088,6 +1184,48 @@ main() {
...
           verify(logger.severe(contains(expectedLateAndAnnotationErrorMessage)));
         });
+
+        group('an unsupported usage of @ConvertProp', () {
+          test('the prop type does not match the Converted type', () {
+            var body = '''
+              @ConvertProp<int, int>(addOne, addOne)
+              late String foo;''';
+
+            expect( () => setUpAndGenerate(OverReactSrc.abstractProps(backwardsCompatible: false, body: body).source), throwsUnsupportedError);
```
sydneyjodon-wk, line 1194: Is there a way to test the error message too? This was the only place I could figure out to test errors since testing it in actual implementation throws before the test runs

lib/over_react.dart
sydneyjodon-wk, line 1: I had to add these imports because the generated part files need them now and the consumer doesn't necessarily import them anymore since they aren't directly used outside the part file

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -308,9 +308,78 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           docComment = '';
         }
 
+        // Parse @ConvertProp annotation.
+        var shouldConvertProp = false;
```
greglittlefield-wf, line 312: I think using an object for the conversion configuration could make the code a little cleaner, so that we can statically ensure that the various variables (shouldConvertProp , rawType, convertedType, setter, getter) are assigned all together, and so we can easily promote their nullabilities.

Also, this is a decently-sized, mostly-self-contained chunk of logic in an already long method, and I think it'd be nice from an organizational perspective to split out.

I was in the refactoring mood 😅  so I took a stab at that:
```diff
diff --git a/lib/src/builder/codegen/accessors_generator.dart b/lib/src/builder/codegen/accessors_generator.dart
index 22c91052..5d9b078e 100644
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -153,10 +153,6 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
     final requiredPropChecks = <String>[];
 
     node.members.whereType<FieldDeclaration>().where((field) => !field.isStatic).forEach((field) {
-      T? getConstantAnnotation<T>(AnnotatedNode member, String name, T value) {
-        return member.metadata.any((annotation) => annotation.name.name == name) ? value : null;
-      }
-
       final accessorMeta = instantiateAnnotationTyped<annotations.Accessor>(field);
       final requiredProp = getConstantAnnotation(field, 'requiredProp', annotations.requiredProp);
       final nullableRequiredProp =
@@ -308,75 +304,20 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
           docComment = '';
         }
 
-        // Parse @ConvertProp annotation.
-        var shouldConvertProp = false;
-        final convertProp =
-            field.metadata.firstWhereOrNull((annotation) => annotation.name.name == 'ConvertProp');
-        var rawType = convertProp?.typeArguments?.arguments.firstOrNull?.toSource();
-        var convertedType = convertProp?.typeArguments?.arguments.lastOrNull?.toSource();
-        var setter = convertProp?.arguments?.arguments.firstOrNull?.toSource();
-        var getter = convertProp?.arguments?.arguments.lastOrNull?.toSource();
-
-        UnsupportedError buildPropAnnotationError(String message) {
-          return UnsupportedError(
-              'Unsupported prop annotation combination for prop $accessorName: $message');
-        }
-
-        if (convertProp != null) {
-          if (rawType == null || convertedType == null) {
-            throw buildPropAnnotationError(
-                'The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`');
-          }
-          if (setter == null || getter == null) {
-            // Analysis errors with `ConvertProp` should prevent this from happening.
-            throw buildPropAnnotationError(
-                'The @ConvertProp annotation must have two arguments: `@Convert<Raw, Converted>(setter, getter)`');
-          }
-          if (convertedType != typeSource) {
-            throw buildPropAnnotationError(
-                'A prop annotated with `@Convert<Raw, Converted>(setter, getter)` should have the same type as the `Converted` generic parameter.');
-          }
-          shouldConvertProp = true;
-        }
-
-        // Look for special-case conversion annotations (@convertJsMapProp and
-        // @convertJsRefProp) only if @ConvertProp(...) annotation isn't already used.
-        if (!shouldConvertProp) {
-          final convertJsMapProp =
-              getConstantAnnotation(field, 'convertJsMapProp', annotations.convertJsMapProp);
-          final convertJsRefProp =
-              getConstantAnnotation(field, 'convertJsRefProp', annotations.convertJsRefProp);
-          if (convertJsMapProp != null) {
-            rawType = 'JsMap?';
-            convertedType = 'Map?';
-            setter = 'jsifyMapProp';
-            getter = 'unjsifyMapProp';
-            shouldConvertProp = true;
-            if (convertedType != typeSource) {
-              throw buildPropAnnotationError(
-                  'A prop annotated with `@convertJsMapProp` should be typed as `Map?`.');
-            }
-          } else if (convertJsRefProp != null) {
-            rawType = 'dynamic';
-            convertedType = 'dynamic';
-            setter = 'jsifyRefProp';
-            getter = 'unjsifyRefProp';
-            shouldConvertProp = true;
-            if (convertedType != typeSource) {
-              throw buildPropAnnotationError(
-                  'A prop annotated with `@convertJsRefProp` should be typed as `dynamic`.');
-            }
-          }
-        }
+        final conversionConfig = parseConversionConfig(
+          field: field,
+          accessorName: accessorName,
+          typeSource: typeSource,
+        );
 
         String castGetterMapValueIfNecessary(String expression) {
-           if (typeSource == null) return expression;
-          return '($expression) as ${shouldConvertProp ? rawType : typeSource}';
+          final typeToCastTo = conversionConfig?.rawType ?? typeSource;
+          if (typeToCastTo == null) return expression;
+          return '($expression) as $typeToCastTo';
         }
 
         String convertIfNecessary(String expression, [bool isGetter = true]) {
-          final convertFunc = isGetter ? getter : setter;
-          if (shouldConvertProp) {
+          final convertFunc = isGetter ? conversionConfig?.getter : conversionConfig?.setter;
+          if (conversionConfig != null) {
             return '$convertFunc($expression)';
           }
           return expression;
@@ -461,6 +402,97 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
   }
 }
 
+T? getConstantAnnotation<T>(AnnotatedNode member, String name, T value) {
+  return member.metadata.any((annotation) => annotation.name.name == name) ? value : null;
+}
+
+// Parse @ConvertProp annotation.
+PropConversionConfig? parseConversionConfig({
+  required FieldDeclaration field,
+  required String accessorName,
+  required String? typeSource,
+}) {
+  PropConversionConfig? conversionConfig;
+
+  final convertProp = field.metadata.firstWhereOrNull((annotation) => annotation.name.name == 'ConvertProp');
+
+  UnsupportedError buildPropAnnotationError(String message) {
+    return UnsupportedError('Unsupported prop annotation combination for prop $accessorName: $message');
+  }
+
+  if (convertProp != null) {
+    var rawType = convertProp.typeArguments?.arguments.firstOrNull?.toSource();
+    var convertedType = convertProp.typeArguments?.arguments.lastOrNull?.toSource();
+    var setter = convertProp.arguments?.arguments.firstOrNull?.toSource();
+    var getter = convertProp.arguments?.arguments.lastOrNull?.toSource();
+    if (rawType == null || convertedType == null) {
+      throw buildPropAnnotationError(
+          'The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`');
+    }
+    if (setter == null || getter == null) {
+      // Analysis errors with `ConvertProp` should prevent this from happening.
+      throw buildPropAnnotationError(
+          'The @ConvertProp annotation must have two arguments: `@Convert<Raw, Converted>(setter, getter)`');
+    }
+    if (convertedType != typeSource) {
+      throw buildPropAnnotationError(
+          'A prop annotated with `@Convert<Raw, Converted>(setter, getter)` should have the same type as the `Converted` generic parameter.');
+    }
+    conversionConfig = PropConversionConfig(
+      rawType: rawType,
+      convertedType: convertedType,
+      setter: setter,
+      getter: getter,
+    );
+  }
+
+  // Look for special-case conversion annotations (@convertJsMapProp and
+  // @convertJsRefProp) only if @ConvertProp(...) annotation isn't already used.
+
+  if (conversionConfig == null) {
+    final convertJsMapProp = getConstantAnnotation(field, 'convertJsMapProp', annotations.convertJsMapProp);
+    final convertJsRefProp = getConstantAnnotation(field, 'convertJsRefProp', annotations.convertJsRefProp);
+    if (convertJsMapProp != null) {
+      conversionConfig = PropConversionConfig(
+         rawType: 'JsMap?',
+         convertedType: 'Map?',
+         setter: 'jsifyMapProp',
+         getter: 'unjsifyMapProp',
+      );
+      if (conversionConfig.convertedType != typeSource) {
+        throw buildPropAnnotationError('A prop annotated with `@convertJsMapProp` should be typed as `Map?`.');
+      }
+    } else if (convertJsRefProp != null) {
+      conversionConfig = PropConversionConfig(
+        rawType: 'dynamic',
+        convertedType: 'dynamic',
+        setter: 'jsifyRefProp',
+        getter: 'unjsifyRefProp',
+      );
+      if (conversionConfig.convertedType != typeSource) {
+        throw buildPropAnnotationError('A prop annotated with `@convertJsRefProp` should be typed as `dynamic`.');
+      }
+    }
+  }
+
+  return conversionConfig;
+}
+
+class PropConversionConfig {
+  final String rawType;
+  final String convertedType;
+  final String setter;
+  final String getter;
+
+  PropConversionConfig({
+    required this.rawType,
+    required this.convertedType,
+    required this.setter,
+    required this.getter
+  });
+
+}
+
 /// Generates mixin implementations (e.g., `$FooPropsMixin`) containing accessors for
 /// all boilerplate versions of props/state mixins.
 class _TypedMapMixinAccessorsGenerator extends TypedMapAccessorsGenerator {
```

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -308,9 +308,78 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
...
+              throw buildPropAnnotationError(
+                  'A prop annotated with `@convertJsRefProp` should be typed as `dynamic`.');
+            }
+          }
+        }
+
         String castGetterMapValueIfNecessary(String expression) {
           if (typeSource == null) return expression;
-          return '($expression) as $typeSource';
+          return '($expression) as ${shouldConvertProp ? rawType : typeSource}';
```
greglittlefield-wf, line 374: Not sure if this case is possible with the current code, but I think we should update this to not short-circuit if `typeSource` is null but `rawType` isn't:
```dart
        String castGetterMapValueIfNecessary(String expression) {
          final typeToCastTo = rawType ?? typeSource;
          if (typeToCastTo == null) return expression;
          return '($expression) as $typeToCastTo';
        }

File: lib/src/builder/codegen/accessors_generator.dart [hunk 5]
```diff
@@ -308,9 +308,78 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
...
+        }
+
+        String convertIfNecessary(String expression, [bool isGetter = true]) {
```
greglittlefield-wf, line 377: Could we make this a named arg, so it's clear what the boolean represents (see [this Dart style guide note](https://dart.dev/effective-dart/design#avoid-positional-boolean-parameters) about positional booleans), and also required to ensure callers are explicit about which they want?

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart
greglittlefield-wf: Another way could be to verify behavior using raw values within the underlying map:

```dart
test('@ConvertProp works as expected', () {
  final fooPropKey = Test().getPropKey((p) => p.foo);

  expect((Test()..foo = 100), equals({fooPropKey: r'$100'}), 
      reason: 'setter should convert value');

  expect(Test({fooPropKey: r'$200'}).foo, 200, 
      reason: 'getter should convert value');
});
```

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart
```diff
@@ -0,0 +1,83 @@
...
+
+String? makeMoney(int? i) {
+  setterReturnValue = i == null ? null : '\$$i';
+  return setterReturnValue!;
+}
+int? getNumber(String? i) => i == null ? null : int.tryParse(i.replaceFirst('\$', ''));
+
+mixin TestProps on UiProps {
+  @ConvertProp<String?, int?>(makeMoney, getNumber)
+  int? foo;
```
greglittlefield-wf, line 76: Could we also add a test case for a non-nullable `late` prop, just to make sure that works?

test/vm_tests/builder/codegen_test.dart
greglittlefield-wf: Oh, I just noticed that the current implementation throws. Could we switch to using `log.severe` instead? (There's a couple other examples in accessors_generator.dart you can reference)

That way:
- you can test the error messages (see other tests in this file that do `verify(logger.severe(contains(yourErrorMessage)));`
- Consumers get slightly better behavior; `logger.severe` will still fail the build, but keep processing the rest of the file and log any other errors/warnings as well. I can't remember if throwing halts the build for other files, but I know `logger.severe` doesn't.

test/vm_tests/builder/codegen_test.dart
```diff
@@ -1088,6 +1184,48 @@ main() {
           setUpAndGenerate(OverReactSrc.abstractProps(backwardsCompatible: false, body: body).source);
           verify(logger.severe(contains(expectedLateAndAnnotationErrorMessage)));
         });
+
+        group('an unsupported usage of @ConvertProp', () {
```
greglittlefield-wf, line 1188: Can we add a test for missing generic parameters?

lib/over_react.dart
```diff
@@ -48,7 +48,7 @@ export 'package:react/react.dart'
         SyntheticEventTypeHelpers,
         DataTransferHelper;
 
-export 'package:react/react_client/js_backed_map.dart' show JsBackedMap;
+export 'package:react/react_client/js_backed_map.dart' show JsBackedMap, JsMap;
```
greglittlefield-wf, line 51: Hmm so we have to export JsMap since it's used for `@convertJsMapProp`... I'm not super in love with adding these to the main entrypoint.

Ooh wait, won't most consumers using those also be using `package:over_react/js_component.dart`, which already exports JsMap, unjsifyMapProp, jsifyMapProp, unjsifyRefProp, and jsifyRefProp?

If so, maybe we don't need to export them here?

greglittlefield-wf (changes requested):
Added some comments, this is looking good though!

test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart
sydneyjodon-wk: ooo I love this! Thank you!!

lib/over_react.dart
```diff
@@ -103,6 +103,7 @@ export 'src/util/key_constants.dart';
 export 'src/util/map_util.dart';
 export 'src/util/memo.dart';
 export 'src/util/pretty_print.dart';
+export 'src/util/prop_conversion.dart' show unjsifyMapProp, jsifyMapProp, unjsifyRefProp, jsifyRefProp;
```
greglittlefield-wf, line 106: Since these are exported from js_util.dart could we also remove them, similar to the comment here: https://github.com/Workiva/over_react/pull/920#discussion_r1664891838

greglittlefield-wf (changes requested):
One more small comment and looks like there's a test failure in CI, otherwise LGTM

lib/over_react.dart
sydneyjodon-wk: ope oops yeah for sure!

greglittlefield-wf (approved):
+1, let me know when this is ready to QA!

lib/src/builder/codegen/accessors_generator.dart
```diff
@@ -392,6 +405,106 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
...
+  }
+
+  if (convertProp != null) {
+    var rawType = convertProp.typeArguments?.arguments.firstOrNull?.toSource();
+    var convertedType = convertProp.typeArguments?.arguments.lastOrNull?.toSource();
+    var setter = convertProp.arguments?.arguments.firstOrNull?.toSource();
+    var getter = convertProp.arguments?.arguments.lastOrNull?.toSource();
+    if (rawType == null || convertedType == null) {
+      logPropAnnotationError(
+          'The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`');
```
greglittlefield-wf, line 435: For these, I think it'd be nice to make the error message a little more helpful by including a source-span-based message that point to the offending code in question, like we do for other errors in this file.

For example, this is what the error output looks like now:
```
[SEVERE] over_react:overReactBuilder on test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart:
Unsupported prop annotation combination for prop foo: The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`
[SEVERE] over_react:overReactBuilder on test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart:
Unsupported prop annotation combination for prop mapProp: A prop annotated with `@convertJsMapProp` should be typed as `Map?`.
```

Vs using using span messages:
```
[SEVERE] over_react:overReactBuilder on test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart:
Unsupported prop annotation combination for prop 'foo': The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`
Source:    ╷
83 │   @ConvertProp(nullableMakeMoney, nullableGetNumber)
   │   ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   ╵
[SEVERE] over_react:overReactBuilder on test/over_react/component_declaration/builder_integration_tests/new_boilerplate/prop_conversion_annotations_test.dart:
Unsupported prop annotation combination for prop 'mapProp': A prop annotated with `@convertJsMapProp` should be typed as `Map?`.
Source:    ╷
90 │   late Map mapProp;
   │   ^^^^^^^^^^^^^^^^^
   ╵
```


This is what that might look like

```diff
diff --git a/lib/src/builder/codegen/accessors_generator.dart b/lib/src/builder/codegen/accessors_generator.dart
index 01927b8d..d6fa73ee 100644
--- a/lib/src/builder/codegen/accessors_generator.dart
+++ b/lib/src/builder/codegen/accessors_generator.dart
@@ -310,7 +310,13 @@ abstract class TypedMapAccessorsGenerator extends BoilerplateDeclarationGenerato
         }
 
         final conversionConfig = parseConversionConfig(
-            field: field, accessorName: accessorName, typeSource: typeSource, logger: logger);
+          field: field,
+          accessorName: accessorName,
+          typeSource: typeSource,
+          onError: (message, [spanNode]) {
+            logger.severe(messageWithSpan(message, span: getSpan(sourceFile, spanNode ?? field)));
+          },
+        );
 
         String castGetterMapValueIfNecessary(String expression) {
           final typeToCastTo = conversionConfig?.rawType ?? typeSource;
@@ -414,15 +420,15 @@ PropConversionConfig? parseConversionConfig({
   required FieldDeclaration field,
   required String accessorName,
   required String? typeSource,
-  required Logger logger,
+  required void Function(String errorMessage, [AstNode? spanNode]) onError,
 }) {
   PropConversionConfig? conversionConfig;
 
   final convertProp =
       field.metadata.firstWhereOrNull((annotation) => annotation.name.name == 'ConvertProp');
 
-  void logPropAnnotationError(String message) {
-    logger.severe('Unsupported prop annotation combination for prop $accessorName: $message');
+  void handleAnnotationError(String message, [AstNode? spanNode]) {
+    onError("Unsupported prop annotation combination for prop '$accessorName': $message", spanNode);
   }
 
   if (convertProp != null) {
@@ -431,19 +437,22 @@ PropConversionConfig? parseConversionConfig({
     var setter = convertProp.arguments?.arguments.firstOrNull?.toSource();
     var getter = convertProp.arguments?.arguments.lastOrNull?.toSource();
     if (rawType == null || convertedType == null) {
-      logPropAnnotationError(
-          'The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`');
+      handleAnnotationError(
+          'The @ConvertProp annotation must be used with generic parameters: `@Convert<Raw, Converted>(setter, getter)`',
+          convertProp);
       return null;
     }
     if (setter == null || getter == null) {
       // Analysis errors with `ConvertProp` should prevent this from happening.
-      logPropAnnotationError(
-          'The @ConvertProp annotation must have two arguments: `@Convert<Raw, Converted>(setter, getter)`');
+      handleAnnotationError(
+          'The @ConvertProp annotation must have two arguments: `@Convert<Raw, Converted>(setter, getter)`',
+          convertProp);
       return null;
     }
     if (convertedType != typeSource) {
-      logPropAnnotationError(
-          'A prop annotated with `@Convert<Raw, Converted>(setter, getter)` should have the same type as the `Converted` generic parameter.');
+      handleAnnotationError(
+          'A prop annotated with `@Convert<Raw, Converted>(setter, getter)` should have the same type as the `Converted` generic parameter.',
+          convertProp);
       return null;
     }
     conversionConfig = PropConversionConfig(
@@ -470,8 +479,8 @@ PropConversionConfig? parseConversionConfig({
         getter: 'unjsifyMapProp',
       );
       if (conversionConfig.convertedType != typeSource) {
-        logPropAnnotationError(
-            'A prop annotated with `@convertJsMapProp` should be typed as `Map?`.');
+        handleAnnotationError(
+            'A prop annotated with `@convertJsMapProp` should be typed as `Map?`.', field);
         return null;
       }
     } else if (convertJsRefProp != null) {
@@ -482,8 +491,8 @@ PropConversionConfig? parseConversionConfig({
         getter: 'unjsifyRefProp',
       );
       if (conversionConfig.convertedType != typeSource) {
-        logPropAnnotationError(
-            'A prop annotated with `@convertJsRefProp` should be typed as `dynamic`.');
+        handleAnnotationError(
+            'A prop annotated with `@convertJsRefProp` should be typed as `dynamic`.', field);
         return null;
       }
     }
```

greglittlefield-wf (changes requested):
Just one more comment on error message improvements found during QA, otherwise LGTM!

greglittlefield-wf (approved):
+10

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #934, opened 2024-07-23
FED-2979 Clean up dependency ranges, clean out the lint trap

aaronlademann-wf:
## Motivation
1. During the null safety migration, we opened up some rather massive dependency ranges to support both null safe, and non-null safe dependencies.
1. After upgrading these dependencies, and the SDK minimum - we have some new lints we should address.

Now that this lib is null safe, those ranges are inaccurate and unnecessary - and can cause bloated CI times due to slower pub get times.

## Changes
1. Bump the Dart SDK Minimum to `2.19.0`
1. Bump all package dependency minimums to their minimum null safe version
1. Clean the lint trap


aviary-wf:
## Security Insights

## Action Items
---

greglittlefield-wf (approved):
+10

@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #958, opened 2024-10-10
FED-3254 Officially support Dart 3

greglittlefield-wf:
Related react-dart PR: https://github.com/Workiva/react-dart/pull/409

## Motivation

We should support Dart 3.

over_react can currently be consumed in Dart 3, despite not supporting it in its SDK constraints, due to the way [Dart 3 treats null-safe packages as compatible](https://dart.dev/resources/dart-3-migration#dart-3-backwards-compatibility). 

However, we don't officially support Dart 3, and our automated CI checks don't validate that it works properly in Dart 3.

Also, we still need to support Dart 2 for the time being, so ideally we'd be able to support both Dart 2 and 3 in the same over_react version, and run CI against each Dart version.

## Changes
- Raise SDK constraints to explicitly support Dart 3
- Fix miscellaneous analysis warnings and test failures in Dart 3
    - See review comments for specifics on these
- Update CI to run against both Dart 2 and Dart 3
    - In Dart 3 runs, delete non-null-safe browser test cases, since they don't apply in Dart 3 and won't compile
    - Add `dart-2-only` tag for VM tests that only apply in Dart 2
    - Work around test hanging issue in Dart 3, add test timeouts so CI doesn't hang
    - Break SBOM job out of matrix, since it can only get run once
- Update dev_dependencies:
    - Widen build_web_compilers to allow newer versions needed by Dart 3
    - Remove unused pedantic and built_value_generator dependencies

> [!CAUTION]
> There's a bug in Dart >=3.3.0 <3.5.0 that breaks react-dart behavior in DDC of certain functions passed to components, such as refs and potentially JS callback props: https://github.com/dart-lang/sdk/issues/56897
>
> We can't prevent users on those versions from pulling in react-dart, since they can already resolve to [null-safe versions](https://dart.dev/resources/dart-3-migration#dart-3-backwards-compatibility) that have already been published.
>
> So, consumers will just have to avoid those versions of Dart, or use dart2js during development or tests. 
>
> If needed, we could potentially follow up with a warning that detects this bug and provides a helpful error.

## Review

  

Please review: 

### QA Checklist
- [x] Tests were updated and provide good coverage of the changeset and other affected code
    - [ ] Steps from PR author: 
        - CI runs in both Dart 2 and 3, and all steps pass
        - Verify the analyzer plugin runs in `tools/analyzer_plugin/playground/`  in an IDE using Dart 3 (just need to `pub get` in that directory and you're good)

## Merge Checklist


test/vm_tests/builder/parsing/members_test.dart
```diff
@@ -195,34 +195,31 @@ main() {
           };
 
           // Loop over the deprecated lifecycle methods
-          legacyLifecycleMethodsMap.keys.forEach((lifecycle) => {
```
greglittlefield-wf, line 198: There were some syntax shenanigans going on here, it looks like from a TS/Dart arrow function syntax mixup 🙂.

This set of curly braces and the ones in the below for-loop are actually being interpreted as set literals, not blocks,  were causing warnings.

So, I fixed that and switched from forEach to a for-loop while I was at it.

lib/src/util/css_value_util.dart
```diff
@@ -132,7 +132,7 @@ class CssValue implements Comparable<CssValue> {
 
   /// Returns whether this value's [number] and [unit] are equal to that of [other].
   @override
-  bool operator ==(dynamic other) {
+  bool operator ==(Object other) {
```
greglittlefield-wf, line 135: This fixes a new built-in Dart warning; apparently `==` will never be passed a null value (`== null` bypasses `operator==`).

lib/src/component_declaration/disposable_manager_proxy.dart
```diff
@@ -63,21 +63,18 @@ mixin DisposableManagerProxy on react.Component implements DisposableManagerV7 {
       _getDisposableProxy().listenToStream(stream, onData,
           onError: onError, onDone: onDone, cancelOnError: cancelOnError);
 
-  @override
```
greglittlefield-wf, line 66: These weren't actually overrides and were infos in Dart 2, but got upgraded to warnings in Dart 3.

test/over_react/component_declaration/builder_integration_tests/backwards_compatible/null_safe_accessor_integration_test.dart
```diff
@@ -96,13 +96,6 @@ void main() {
             testValue: 'test value',
           );
         });
-        test('nullable dynamic with extraneous ? syntax', () {
```
greglittlefield-wf, line 99: These test cases tested a `dynamic?` type as an edge case, which now warns instead of infos when used (because the `?` is unnecessary).

I tried suppressing those warnings, but `dynamic?` also appears in the generated part file, and there's no easy way to suppress that warning for just that file.

Since this was an edge-case that warns when the user authors it, preserving the exact type is covered by other test cases, and we originally added this case "just in case", I opted to just remove this case here and in other files.

tools/analyzer_plugin/analysis_options.yaml
```diff
@@ -15,8 +15,9 @@ analyzer:
     missing_return: warning
     # Override workiva_analysis_options upgrade to warning
     unused_import: info
-    import_of_legacy_library_into_null_safe: warning
```
greglittlefield-wf, line 18: This no longer exists in Dart 3

lib/src/util/css_value_util.dart
robbecker-wf: interesting

robbecker-wf (approved):

robbecker-wf:
QA+1 CI passes on all combinations of Dart 2 and analyzer. There are validations in place to check that pub resolves the correct analyzer version.
Verified the plugin worked in IDE using Dart 3
Ran some of the examples using Dart 3

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #985, opened 2025-08-05
no_entrypoint_imports

sourcegraph-wk:
## Problem

An entrypoint import can be defined as an import within `lib/src` that imports a file within `lib/*.dart` (the entrypoints of a dart package). These imports are always unnecessary, and can contribute to slower dart build performance.

## Solution

This PR naively removes every entrypoint import within the repo, and updates the `gha-dart/.../checks.yaml` to the latest version. In order to merge this PR a few manual changes must be performed:

- Navigate to every file which has analysis errors caused from the missing import. Rely on intellisense to import the necessary symbol from the specific file, not the entrypoint import
- Implement the `no_entrypoint_imports` additional-checks option in `gha-dart@v2.10.13` that was added in [this PR](https://github.com/Workiva/gha-dart/pull/716). This will prevent new entrypoint imports from getting added once this PR merges

Our request to teams is to perform the above tasks, so we can default enable the `no_entrypoint_imports` check for all gha-dart consumers. Please reach out to [#support-frontend-dx](https://workiva.enterprise.slack.com/archives/C03ESR91BNU) for any questions or issues

## QA

- [ ] This pr is heavily dependant on the analysis server, and as such CI passes should be sufficient

[_Created by Sourcegraph batch change `Workiva/no_entrypoint_imports`._](https://workiva.sourcegraphcloud.com/organizations/Workiva/batch-changes/no_entrypoint_imports)

### DIFF
--- a/lib/src/component/aria_mixin.dart
+++ b/lib/src/component/aria_mixin.dart
@@ -18,7 +18,7 @@ import 'dart:collection';
 
 // Must import these consts because they are used in the transformed code.
 // ignore: unused_shown_name
-
+import 'package:over_react/over_react.dart' show PropDescriptor, PropsMeta;
 import 'package:over_react/src/component_declaration/annotations.dart';
 
 part 'aria_mixin.over_react.g.dart';
--- a/lib/src/component/aria_mixin.dart
+++ b/lib/src/component/aria_mixin.dart
@@ -18,8 +18,7 @@ import 'dart:collection';
 
 // Must import these consts because they are used in the transformed code.
 // ignore: unused_shown_name
-import 'package:over_react/over_react.dart'
-    show PropDescriptor, PropsMeta;
+
 import 'package:over_react/src/component_declaration/annotations.dart';
 
 part 'aria_mixin.over_react.g.dart';
--- a/lib/src/component/aria_mixin.dart
+++ b/lib/src/component/aria_mixin.dart
@@ -16,10 +16,10 @@ library over_react.aria_mixin;
 
 import 'dart:collection';
 
+import 'package:over_react/src/component_declaration/annotations.dart';
 // Must import these consts because they are used in the transformed code.
 // ignore: unused_shown_name
-import 'package:over_react/over_react.dart' show PropDescriptor, PropsMeta;
-import 'package:over_react/src/component_declaration/annotations.dart';
+import 'package:over_react/src/component_declaration/component_base.dart' show PropDescriptor, PropsMeta;
 
 part 'aria_mixin.over_react.g.dart';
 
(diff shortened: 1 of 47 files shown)


annawatson-wk (approved):
+10, CI passes

lib/src/component/aria_mixin.dart
```diff
@@ -18,8 +18,7 @@ import 'dart:collection';
 
 // Must import these consts because they are used in the transformed code.
 // ignore: unused_shown_name
-import 'package:over_react/over_react.dart'
-    show PropDescriptor, PropsMeta;
+import 'package:over_react/over_react.dart' show PropDescriptor, PropsMeta;
```
greglittlefield-wf, line 21: Should this be updated to a `src` import?

lib/src/util/rem_util.dart
```diff
@@ -20,12 +20,14 @@ import 'dart:async';
...
-import 'package:over_react/over_react.dart';
-import 'package:over_react/components.dart' as v2;
+import 'package:over_react/src/component/dom_components.dart';
+import 'package:over_react/src/component/resize_sensor.dart' as v2;
 import 'package:over_react/src/component_declaration/component_base.dart' as component_base;
-import 'package:over_react/react_dom.dart' as react_dom;
+import 'package:over_react/src/util/css_value_util.dart';
 import 'package:platform_detect/platform_detect.dart';
 
+import '../../react_dom.dart' as react_dom;
```
greglittlefield-wf, line 29: Could we make this a `package:` import instead of a relative path?

lib/src/util/safe_render_manager/safe_render_manager.dart
```diff
@@ -15,10 +15,13 @@
...
 
-import 'package:over_react/over_react.dart';
-import 'package:over_react/react_dom.dart' as react_dom;
+import 'package:over_react/src/component/ref_util.dart';
+import 'package:react/react_client.dart'
+    show
+    ReactElement;
 import 'package:w_common/disposable.dart';
 
+import '../../../react_dom.dart' as react_dom;
```
greglittlefield-wf, line 24: Same comment as above

greglittlefield-wf (changes requested):
Just a couple small comments, otherwise LGTM!

lib/src/component/aria_mixin.dart
sydneyjodon-wk: ohh good catch :loading:

lib/src/util/rem_util.dart
sydneyjodon-wk: because that file is lib/react_dom.dart so I couldn't get a src import

PR #1012, opened 2026-08-06
FED-4894 Support analyzer >=10 <15

greglittlefield-wf:
## Motivation
The latest analyzer package version is 14.x, but the latest over_react supports is 10.x.

We want to support the latest, to unblock upgrading of other packages, and to allow consumers to use newer language features without risk of parsing errors.

## Solution
- Bump analyzer dependency from `>=5.13.0 <11.0.0` to `>=10.0.0 <15.0.0`
    - Supporting a wide range is difficult due to cascading API removals, so the lower bound was raised to 10
- Update analyzer-consuming code in builder to work with breakages throughout this new range
    - `NamedCompilationUnitMember` removals - changed to `CompilationUnitMember`, added `ClassishDeclaration.name` to help with this and used it since we already had an instance of it in most places
    - Add `BlockClassBody` casts when needed
    - Add `namedArgumentParts` to deal with `ArgumentList` element type changes
- Update tool/set_analyzer_constraint.dart to delete dependencies that unnecessarily constrain `analyzer`
     - This was an issue when other packages not being up to date were blocking over_react from resolving without overrides, and probably isn't necessary today, but I'm gonna leave it in since it'll likely be useful in the next analyzer upgrade

## QA Instrucions
- CI passes
- Verify that each supported analyzer major has a corresponding `validate_analyzer` CI run
- (optional) Pull into a big app and verify it builds with the latest analyzer


lib/src/builder/parsing/ast_util.dart
```diff
@@ -68,31 +68,59 @@ extension TypeAnnotationNameHelper on TypeAnnotation {
 /// Extension built on [NameHelper] to allow for easy access to the `name`
 /// field of [Identifier]s.
 extension TypeNameHelper on NamedType {
-  // Backwards compatibility for various analyzer versions that remove name/name2.
```
greglittlefield-wf, line 71: Removed these since they were only needed for older analyzer versions

test/test_util/test_util.dart
```diff
@@ -23,12 +23,12 @@ export 'package:over_react_test/over_react_test.dart' hide testJsComponentFactor
 
 dynamic getJsChildren(instance) => getProperty(instance.props as Object, 'children');
 
-dynamic getRawDartChildren(var renderedInstance) {
+dynamic getRawDartChildren(dynamic renderedInstance) {
```
greglittlefield-wf, line 26: This was a random compiler warning I fixed while I was debugging build_runner/test issues

danaoredson-wf (approved):

.github/workflows/ci.yml
```diff
@@ -157,7 +156,9 @@ jobs:
 
       - id: link
         name: Override over_react dependency with local path
-        run: cd ../.. && dart pub get && dart tool/travis_link_plugin_deps.dart
+        # TODO reinstate this once we update analyzer_plugin's analyzer range to overlap with what over_react supports
```
greglittlefield-wf, line 159: Note that analyzer_plugin isn't up to date on analyzer yet; that'll be a separate PR

test/test_util/test_util.dart
robbecker-wf: yep, this is now an error to use var as a type in a declaration

pubspec.yaml
```diff
@@ -1,18 +1,18 @@
 name: over_react
-version: 5.7.0
+version: 5.8.0
```
robbecker-wf, line 2: should this be bumped now or as part of the release?

robbecker-wf (approved):

robbecker-wf:
QA +1 CI passes, builds and serves on analyzer 14 in internal projects

pubspec.yaml
greglittlefield-wf: Oh good catch; usually it gets bumped in the release, but I don't think having it in this PR will cause any issues. If it does, I'll fix it up in the release PR

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

PR #1013, opened 2026-08-06
FED-5022 Drop Dart 2 support

greglittlefield-wf:
## Motivation
Maintaining a large range of analyzer versions comes with a bit of overhead, and getting to the latest analyzer versions (https://github.com/Workiva/over_react/pull/1012) while also supporting Dart 2 just isn't feasible.

We've supported Dart 2 + 3 in the same version of over_react for a while, and now it's time to drop Dart 2 support.

## Changes
- Update CI to no longer run on Dart 2, remove conditional Dart-2-only/Dart-3-only logic
- Raise Dart SDK lower bound from `>=2.19` to `>=3.0.0`
- Fix errors/warnings resulting from upping the language version to 3.0 via the upped SDK constraint
    - class-used-as-mixin errors. From CHANGELOG.md:

        > - Fix Dart language level 3.0 errors and downstream compile errors by switching internal classes to `mixin` declarations
        >   - `GeneratedClass` - switched to `mixin`
        >   - `PropsMapViewMixin`, `StateMapViewMixin`, `MapViewMixin` - switched to `mixin`, also moved `implements` to `on`
        >   - Hand-authored mixin "generated" classes: 
        >     - Core: `$AriaPropsMixin`, `$ConnectFluxPropsMixin`, `$CssClassPropsMixin`, `$DomPropsMixin`, `$ReactPropsMixin`, `$SvgPropsMixin`, `$UbiquitousDomPropsMixin`
        >     - Components: `$ErrorBoundaryPropsMixin`, `$ErrorBoundaryStateMixin`, `$ResizeSensorPropsMixin`, `$TransitionPropsMixin`
        >   - While some of these are technically a breaking changes, there's very little chance of them impacting consumers, so we're releasing it as a minor.
        >   
        >     These classes have no known usages outside of over_react or its generated code, and are for internal use only. 
        >   
        >     Additionally, the `*MapViewMixin` ones never should have been exported in the first place, and have been deprecated accordingly.

    - Remove unreachable switch statement `default` cases
    - Use modern equivalent to implicit-casts flag in analysis_options.yaml
have ever been exported
- Add regression tests for props mixin compile errors I didn't discover until pulling this branch into an internal Dart 3 consumer: test/over_react/component/props_mixin_compile_test.dart
- Delete all test cases for non-null-safe and unsound null safety consumer code
     - Everything must be null-safe in Dart 3, so these are no longer applicable and wouldn't compile even if we tried to run them
- Add changelog entry

## QA steps
- CI passes
- Verify that new test/over_react/component/props_mixin_compile_test.dart regression tests fail on the commit before the fix ([35d57cb](https://github.com/Workiva/over_react/pull/1013/commits/35d57cb5fbd65da5718917e6c06e86b02e011a21)), and pass on latest commit
- Override a big Dart 3 app with this branch and verify it compiles cleanly


.github/workflows/ci.yml
```diff
@@ -33,7 +38,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        sdk: [ 2.19.6, stable ]
+        sdk: [ 3.9.0, stable ]
```
greglittlefield-wf, line 41: This is kind of arbitrary; I just picked an older-but-not-too-old version of Dart 3

robbecker-wf (approved):
Looks great!

danaoredson-wf (approved):

robbecker-wf:
QA +1 CI passes. Builds and runs in our internal app.

greglittlefield-wf:
@Workiva/release-management-p

rmconsole-wf (approved):
+1 from RM

